PR #59, opened 2016-04-12
Allow child iframe to go fullscreen

lmeyerov:
cc @thibaudh can you check this? the corresponding button is already graph-viz/streamgl master, triggered when in an iframe

### DIFF
--- a/graphistry/util.py
+++ b/graphistry/util.py
@@ -25,6 +25,8 @@ def make_iframe(raw_url, height, protocol):
 
     iframe = '''
             <iframe id="%s" src="%s"
+                    allowfullscreen="true" webkitallowfullscreen="true" mozallowfullscreen="true"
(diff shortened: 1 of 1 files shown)


lmeyerov:
Note: without this, the new button does nothing. Force a major version bump as part of the next release?

thibaudh:
Why a major bump? There are no user facing api changes.

lmeyerov:
Well, not sure how to handle. If they use a more recent StreamGL and an old PyGraphistry, they'll see a button that doesn't do anything. (The parent frame will squash the request w/out the explicit permission from this PR.) So, w/out a bump, there's a broken-feeling button.

thibaudh:
We have a warning when the pygraphistry version is not up-to-date. That should help quite a bit. I would rather keep major bump for API overhauls.

PR #80, opened 2017-08-19
Python 3 syntax error lambda (r): --> lambda r:

cclauss:
Parens around lambda parameters is a syntax error in Python 3


lmeyerov:
Thanks, LGTM. We don't have CI setup for python 3, which would help here in the future.

PR #87, opened 2018-01-12
include etl-1's edgelist and unsorted labels if `&echo=true`

trxcllnt:


lmeyerov:
@trxcllnt Just saw. It looked innocuous, but when I tried with a simple edges-only triangle, got errors.

lmeyerov:
Let me know if I should dig deeper..

PR #92, opened 2018-03-27
add Neo4j example notebook

johnymontana:
Added demo Jupyter notebook showing how to create PyGraphistry visualizations from data in Neo4j using the Python driver for Neo4j and a common publicly available Neo4j Twitter dataset.


lmeyerov:
Hi @johnymontana love this notebook! 

I'm guessing users will want to copy-paste, would it make sense to abstract out the boilerplate? In particular, a way to get a neo2pandas function:

```
def neo2pandas(driver, query):
     with driver.session() as session:
              return DataFrame(session.run(query).data())
```

Also, what may be fun, is to use the data in the tool. For example, doing a live filter in-tool on pagerank or k-cores using the histogram component. 

Either way, pretty exciting!

PR #100, opened 2018-10-30
fix(default to empty arguments for cypher query)

cwharris:


lmeyerov:
## The lane now bootstraps and scores — 3 runs

Final blocker was the **same hardcoding defect a third time**: `--memory` for the safe_run container passed `DEFAULT_NEO4J_DOCKER_MEMORY` instead of the suite's `neo4j: docker_memory`, so the container was capped however large a heap the config asked for. The transaction pool was never the cause — it only became visible once the container was real. Failures then unwound in order, each genuine: boots at 48g → tx pool trips at 5.6 GiB → `tx_total_max: 24G` → Java heap OOM at 8G → heap 24G / pagecache 8G → **bootstrap ok**.

### LDBC SNB SF1, un-starved Neo4j 5 (48g / 24G heap / 8G pagecache / 24G tx), rows validated, median of 3

| query | GFQL polars (pooled 8-slot A/B) | Neo4j (median of 3) | verdict |
|---|---|---|---|
| IS1 seed-lookup | **77.41** | 173.67 `[150.1, 202.6, 173.7]` | **GFQL WIN 2.24×** |
| IS4 message-content | **3.44** | 23.29 `[24.5, 23.3, 22.7]` | **GFQL WIN 6.77×** |
| IS7 recent-replies | **13.15** | 111.95 `[112.0, 111.0, 126.1]` | **GFQL WIN 8.51×** |
| IS2 expand-order-limit | 92.89 | **66.06** `[66.1, 76.2, 63.5]` | LOSE 1.41× |
| IS5 message-creator | 72.38 | **31.00** `[32.0, 27.8, 31.0]` | LOSE 2.35× |
| IS3 one-hop-expand | unanswerable | 128.14 | GFQL capability gap |

Wins hold across all three runs with margins far outside the host's ~6–10% spread.

### What this is, precisely

- **Not flips of known losses.** Before this, only IC8 had ever been scored against Neo4j — IS1–IS7 were *unscored*, not lost. Six cells now exist where one did, and three are wins.
- **IS1 is the notable one:** a 77× loss against Kuzu is a 2.24× win against Neo4j. The competitors aren't interchangeable, and a single "we lose seeded point lookups" story doesn't hold.
- **The Kuzu board is unchanged** (3W/4L). Nothing here touches it.
- **IC8's existing "WIN 1.3× vs Neo4j" is still suspect** — measured pre-fix, i.e. against a starved competitor. Re-measure on this branch before reuse.
- IS2 carries `adapter_workaround` on the Neo4j side; inspect before publishing that cell. IS6 is absent from the adapter.

PR #107, opened 2019-01-25
Graphistry APIv3 / Arrow support - v1.0.0-alpha

cwharris:


lmeyerov:
Wooo!

If we can publish to pypi as `v1.0.0-alpha` (so default pip install ignores), happy to start testing live!

Looks like we lost `hyper`: instead of `rm`, we should `mv`, and hopefully that way get the recent bugfixes + `direct=True` mode (+ tests).

lmeyerov:
Also, Travis seems moody: https://travis-ci.org/graphistry/pygraphistry .

cwharris:
hypergraph is back, and `hyper.py` is `mv`'d. I think we can merge to `v1.0.0-alpha` branch, tag it there, and the tag will be maintained when we merge to master later. That way we have time to fix bugs before altering master.

cwharris:
Also, I'm wondering if I should make a `travis-cicd-pypi` branch we can use to make sure pypi publishing works correctly...

trxcllnt:
@cwharris both of those sound like good ideas to me!

PR #110, opened 2019-04-05
feature(client path control): specify scheme/authority/path

lsb:
So that, for example, one might be able to specify "", for the iframe to direct to "/graph/graph.html", or one might be able to specify "/graphistry-subpath", for the iframe to direct to "/graphistry-subpath/graph/graph.html"


lmeyerov:
cool, afaict is backwards compatible with extent client code.

I'd like to get travisci passing. for interim, may just remove networkx from test deps or force to an old version.

PR #113, opened 2019-04-08
feature(client path control): specify client url scheme/authority/path

lsb:
Everything before /graph/graph.html basically


lmeyerov:
lgtm

PR #115, opened 2019-04-17
Avoid setting variables in %%time blocks, and fix link at end

lsb:


lmeyerov:
@lsb  Can we keep the labs graph.html link? otherwise default viz won't load

lmeyerov:
also, setting variables in %%time works afaict: https://colab.research.google.com/drive/1fPR4QIoDw8uOBtDdtKCMxI8znjstI4po#scrollTo=6rFukyamC8X1  

your issue may have been with %%timeit..

lsb:
when-i-do-a-straight-run

PR #118, opened 2019-04-23
fix(nb): remove time

lmeyerov:
remove %%time because scoping bugs and also trim long result scroll


lmeyerov:
We can close when review passes

PR #260, opened 2021-08-16
Fixes exception thrown for networkx recent version stamps

cerebis:
Parsing of the version string for Networkx relies on the assumption that there are only 2 parts (eg. 1.2 or 2.3). Releases which include subversions (2.5.1) fail to be parsed and raise an exception.

Since only the major part of the version number is needed, I have modified the method to split only this component. Additionally, I have removed the integer conversion, since there is no need for making the subsequent decesion on API usage.

### DIFF
--- a/graphistry/PlotterBase.py
+++ b/graphistry/PlotterBase.py
@@ -1208,10 +1208,10 @@ class PlotterBase(Plottable):
     def networkx_checkoverlap(self, g):
 
         import networkx as nx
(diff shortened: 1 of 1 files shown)


lmeyerov:
I'm on the road, will work on getting out as 0.19.6 shortly. Thank you!

lmeyerov:
... and it's out @ https://pypi.org/project/graphistry/0.19.6/ . Thanks!

cerebis:
No worries.

PR #280, opened 2021-12-02
fix(api): move all examples to api=3 

pradkrish:


lmeyerov:
I went through the diffs, that was heroic, thank you!

Issue #193, opened 2021-01-07
[FEA] Node reductions


lmeyerov:
Node reductions are great!

# What 
This is an important case of graph reductions. There are requests in both the UI and APIs to do reductions at the level of node, multiedge, and multiple nodes / subgraphs. Instead of simply filtering out the selected entities, the idea is they'd be partitioned into subgraphs, and different kinds of operations would replace them new topologies, and attributes representing them. They seem like some sort of continuous derivative operators, and may even be invertable.

# Example 
One of the most common cases is dropping a node and propagating its attributes and edges/attributes through its neighborhood, so 

In hypergraph:

```python
g2 = graphistry.hypergraph(
    pd.DataFrame([
       {'x': 'a', 'y': 'm'},
       {'x': 'b', 'y': 'm'}      
   ],
  direct=True)
```

We'd get 3-node graph `(a)-[e1]->(m)<-[e2]-(b)`

However, we really just want `(a)-[m]-(b)` ... so need a way to drop node `(m)` and synthesize ede `-[m]->`. This gets weirder in the cases of multiedges, cliques, and different kinds of nodes.

# Sample reducers

This gets at patterns of graph reducers, where we want to take nodes/edges, remove them, and replace with other nodes edges. For example, we can replace all `type='m'` nodes with `(a: x)-[e]->(b: x)` via a node reduction driven by a selector predicate:

## 1. Node reducers: Drop a node by converting to edges

`g2.replace_node_with_edges(selected=g._nodes['type'] == m, reducer=lambda edges: strong_component(edges_to_nodes(edges)))`

The idea is pull out nodes matching some predicate

```python
def replace_node_with_edges(g):
  nodes = g._nodes[ selected ]
  new_edges = []
  for node in g._nodes[g._node]:
    in_edges = g._edges[ g._edges[g._src] == node ]
    out_edges = g._edges[ g._edges[g._src] == node ]
    new_edges.append( reducer(pd.concat([in_edges, out_edges]))
  new_nodes = g._nodes[ ~selected ]
  g2 = g.nodes(new_nodes).edges(new_edges)
  return g2
```

where 

```
def strong_component(edges, s='src', d='dst')
    node_ids = pd.concat([edges[s].unique() + edges[d].unique()]).unique()
   edges = []
    for x in range(len(node_ides)):
       for y in range(x+1, len(node_ids)):
           edges.append([x,y])
   return edges
```

## 2. Sugar for common cases

[...]

PR #283, opened 2021-12-07
[WIP] Node reductions

pradkrish:
For issue https://github.com/graphistry/pygraphistry/issues/193

### DIFF
--- a/graphistry/compute.py
+++ b/graphistry/compute.py
@@ -1,5 +1,6 @@
 from typing import Any, Callable, Iterable, List, Optional, Set, Union, TYPE_CHECKING
 import logging
+import itertools, pandas
 from .Plottable import Plottable
 logger = logging.getLogger('compute')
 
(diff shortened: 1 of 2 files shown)


pradkrish:
@lmeyerov Despite this being pretty much a work in progress, I already opened a PR to get some initial feedback. I hope that's okay :)

This is a partial implementation of table reductions. The function `replace_nodes_with_edges(node_selector)` can currently accept both a string and a list of strings. For example the following calls `g.replace_nodes_with_edges(['m']) `and `g.replace_nodes_with_edges('m')` both yield the same results. Here is an example:

```python
g = graphistry.edges(pd.DataFrame([{'x': 'a', 'y': 'm'},
                                   {'x': 'b', 'y': 'm'},
                                   {'x': 'c', 'y': 'n'},
                                   {'x': 'd', 'y': 'm'}]), 'x', 'y')
g.replace_nodes_with_edges(['m',])
print(g._edges)

#result
   x  y reduced_from
0  a  b            m
1  a  d            m
2  b  d            m
3  c  n          NaN

```
What are your thoughts?

lmeyerov:
Cool, some thoughts:

- what happens if `m` is a src, not dst?
- how about specify nodes as a `Series`, say 1-column or if multi-column, index on `g._node`? ex: user might pick `g._nodes[ g._nodes['type'] == 'account' ]`
- `reduced_from` right now (afaict) would report `"m,n,o,p"` instead of `"m"` when used on a list input

I'll need to look at `itertools`, but the rest seems straightforward to port to `cudf` / `dask_cudf` :)

pradkrish:
@lmeyerov  Thank you very much for your feedback. I have pushed another commit now. Kindly have a look at the `replace_nodes_with_edges(..)` function comments, it should address some of the issues your raised. I have also added some tests to show how to use this function. As far itertools, I have replaced the logic which now uses `numpy.tril`. Please have a look and let me know what you think. :)

lmeyerov:
Great!

In case it helped, I sketched out a few the top use cases we see in the original ticket: https://github.com/graphistry/pygraphistry/issues/193  

re:`tril`, I'm struggling to understand that logic, which suggests someone doing a future bugfix here would need to rewrite it.. can it be something like `pd.concat`? I'm guessing once the aggregation operators come in and more tests, we'll have a better sense of what needs to be fast/slow, so simple vector pd calls should get us far..

lmeyerov:
Just checking in on this. Maybe the thing to do is:

- make sure each of the motivating examples is in a test
- test a few different tricky graph shapes

I'm thinking we can land this as-is, and update the underlying feature request to more carefully explore (a) a custom reductions mode for this and (b) multi-node collapsing/grouping

PR #286, opened 2021-12-15
typecheck api values

### LINKED ISSUE #285: [FEA] typecheck invalid api=... value
**Is your feature request related to a problem? Please describe.**
When mistakingly writing `register(api='3')` vs `api=3`, error happened at `.plot()` and in a surprising way. 

**Describe the solution you'd like**
Would have been better for `register()` to throw a type error, and `plot()` errors that nothing set.

pradkrish:
Fixes #285

### DIFF
--- a/graphistry/pygraphistry.py
+++ b/graphistry/pygraphistry.py
@@ -272,6 +272,8 @@ class PyGraphistry(object):
         Also set via environment variable GRAPHISTRY_API_VERSION."""
         if value is None:
             return PyGraphistry._config['api_version']
+        if value not in [1,2,3]:
+            raise  TypeError("Invalid api, wrong type")
         # setter
         PyGraphistry._config['api_version'] = value
 
--- a/graphistry/pygraphistry.py
+++ b/graphistry/pygraphistry.py
@@ -182,6 +182,7 @@ class PyGraphistry(object):
         if value is None:
             return PyGraphistry._config['hostname']
 
+
         # setter
         shortcuts = {}
         if value in shortcuts:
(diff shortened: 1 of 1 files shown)


pradkrish:
@lmeyerov  isn't it enough with a typecheck in register?

lmeyerov:
The `.plot()` case is more if their `.register()` calls didn't finalize enough info, such as forgetting to specify `api=x` in at least one of them, or not calling it at all

lmeyerov:
Thank you -- we try to make exceptions prescriptive, e.g., `"Received value x for parameter y, instead use one of z"`

pradkrish:
@lmeyerov  Now it should throw an error for the following scenarios

- no calls to register but try to use plot
- api='' 
- forgetting to specify `api=x` in register
- api='3'

graphistry/pygraphistry.py
```diff
@@ -270,8 +271,10 @@ def protocol(value=None):
     def api_version(value=None):
         """Set or get the API version: 1 or 2 for 1.0 (deprecated), 3 for 2.0
         Also set via environment variable GRAPHISTRY_API_VERSION."""
-        if value is None:
```
lmeyerov, line 273: I think we still need `:273` for the getter mode...

pradkrish: Currently when the value is `None`, that is, not set by the user, `api` is set to `1` (default config). Do you expect that we keep this behaviour?

lmeyerov: Yes we should preserve -- that will work in some environments that set params via env vars, so rejecting would be a breaking change for them . When we start defaulting to api=3, that'd be a friendlier time to reject.

The current need is just to catch people accidentally passing in say `4` or `"3"`

PR #308, opened 2022-02-17
Layered layout implementation.

lmeyerov:
Creating PR to aid w/ landing discussion

EDIT: Switching to release candidate at https://github.com/graphistry/pygraphistry/pull/314


lmeyerov:
- [ ] `demos/layout.ipynb`: Especially if we expect some useful knobs, maybe belongs in `demos/more_examples/graphistry_features/layout_hierarchical.ipynb`

graphistry/layout/networkx.py
```diff
@@ -0,0 +1,35 @@
+def to_networkx(g):
```
lmeyerov, line 1: Is this for reusing networkx methods, or a dev convenience for stuff like rendering?

Maybe it's just an internal method. From a public interface perspective, better if we do something like `sugiyuma` -> `graphistry plottable via nodes/edges dataframes` -> `networkx`

graphistry/layout/poset.py
```diff
@@ -0,0 +1,151 @@
+from collections import OrderedDict
+
+
+class Poset(object):
```
lmeyerov, line 4: A little surprising to see here -- sounds like it's primarily here for better speed on just __cmp__ / __eq__ ?

Orbifold:
Chart documentation, see the [demo notebook](https://github.com/graphistry/pygraphistry/blob/sugiyama/demos/layout.ipynb).

Orbifold:
Folder organization: depends whether you wish to add other layout algorithms in the future. Creating mindmaps, Sankey and whatnot is a lifetime however. I mean, clear folder organization requires clear plans for the future.

graphistry/layout/networkx.py
Orbifold: Convenience to render something. The layout mechanics does not (intentional) depend on NetworkX. It would have been eaiser to rely on NetworkX.

graphistry/layout/poset.py
Orbifold: A poset, seen as so-called Hasse diagram, is a layered layout without the multi-layer edges. In partially order language, if you remove the edges representing transitive closures you end up with a poset representing the layout you are after. Said differently, the Sugiyama layout is an algorithm which tries to find a partial order fitting a given pre-order. So, the poset structure embodies the layout in an algebraic thing. The cmp and all that is the partial order. You can implement the layout without a poset but the  operations (e.g. comparing where a vertex sits in the stack of layers wrt to another vertex) become less elegant. Layout algorithms are in general quite abstract and it helps to define things like this to support the conceptual flow.

lmeyerov: Yes, so I was wondering if we can reuse someone else's implementation, or even need it. It is a generic class and adds maintenance burden - a big file in this commit

lmeyerov:
Yes, this is the 2nd or 3rd already, and we expect more. I do not have a sense whether any future will share code with this impl due to not being in a vector friendly style. So putting in a layout subfolder now helps :)

graphistry/layout/poset.py
Orbifold: The Poset class is ~140 lines of code. You would like to use an external poset package to replace this file?

graphistry/layout/networkx.py
Orbifold: Removed the NetworkX references but it means that the demo notebook had to go as well.

Orbifold: The NetworkX reference is gone in the layout module but used it in the demo notebook in order to visualize things. Having in the notebook only a bunch of coordinates is not really pleasing and I suppose it's alright to have NetworkX in the notebook.

graphistry/layout/poset.py
lmeyerov: If there is something reasonable; custom collections is a bad idea if can be avoided

graphistry/layout/__init__.py
```diff
@@ -1,12 +1,4 @@
...
-from .layer import Layer
-from .edge import Edge
-from .edgeBase import EdgeBase
-from .graph import Graph
-from .graphBase import GraphBase
-from .sugiyamaLayout import SugiyamaLayout
-from .routing import route_with_lines, EdgeViewer, route_with_splines, route_with_rounded_corners
-from .vertex import Vertex
-from .vertexBase import VertexBase
+from .utils import *
```
lmeyerov, line 1: Can we avoid `*` imports?

lmeyerov:
@Orbifold  Can you hook this up to `layouts.py` so we can call it directly, e.g., `graphistry.edges(df, 'src', 'dst').this_lib().plot()`?

`tree_layout` shows a lot of how this works: https://github.com/graphistry/pygraphistry/blob/53bd7a779b9efc216301fcee94b493de9184cbc2/graphistry/layouts.py#L16

The mixin gets:

* `self._nodes, self._node` : nodes df and node id col. Might not be bound if only an edge table, so may need to synthesize nodes (see helper call in tree_layout)
* `self._edges, self._source, self._destination`: edges df and src/dst id cols

It is in a functional style, so you'd return a new g that's a clone of the base, but with a new nodes df with new x/y cols, just like `tree_layout`

lmeyerov:
@Orbifold  I'm having difficulty trying this --

```python
! pip install --user https://github.com/graphistry/pygraphistry/archive/sugiyama.zip
# ... restart kernel ...
import graphistry
```

=>

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/tmp/ipykernel_4513/1067964223.py in <module>
----> 1 import graphistry

~/.local/lib/python3.8/site-packages/graphistry/__init__.py in <module>
----> 1 from graphistry.pygraphistry import (  # noqa: E402, F401
      2     client_protocol_hostname, protocol, server,
      3     register, privacy, login, refresh, api_token, verify_token,
      4     store_token_creds_in_memory,
      5     name, description,

~/.local/lib/python3.8/site-packages/graphistry/pygraphistry.py in <module>
     13 from . import util
     14 from . import bolt_util
---> 15 from .plotter import Plotter
     16 
     17 import logging

~/.local/lib/python3.8/site-packages/graphistry/plotter.py in <module>
      2 from .compute import ComputeMixin
      3 from .gremlin import GremlinMixin, CosmosMixin, NeptuneMixin
----> 4 from .layouts import LayoutsMixin
      5 
      6 class Plotter(  # type: ignore

~/.local/lib/python3.8/site-packages/graphistry/layouts.py in <module>
      2 import logging
      3 from .Plottable import Plottable
----> 4 from .layout import SugiyamaLayout
      5 
      6 logger = logging.getLogger('layouts')

ModuleNotFoundError: No module named 'graphistry.layout'
```

lmeyerov:
@Orbifold The other part is this does not seem to work the graphistry library, as it should work similar to tree_layout in flow, while current is more for networkx etc users instead


Compare:

```python
g1 = graphistry.edges(pd.read_csv('some_logs.csv'), 'src_ip', 'dst_ip')
g2 = g2.tree_layout(ascending=False, level_align='center')
g2.plot()
```

vs:

```python
from graphistry.layout.sugiyama import SugiyamaLayout
from graphistry.layout.graph import Graph, Vertex, Edge
...
gg = SugiyamaLayout.graph_from_pandas(g, source_column = source_column, target_column = target_column)
positions = SugiyamaLayout.arrange(gg, layout_direction = layout_direction)
nx.draw(nxg, pos = positions, with_labels = True, verticalalignment = 'bottom', arrowsize = 3, horizontalalignment = "left", font_size = 20)
```

PR #316, opened 2022-03-13
Sugialex

lmeyerov:
Adds:
* Sugiyama layout for `.tree_layout(...)`: See  https://github.com/graphistry/pygraphistry/pull/308
* `.rotate(degree)`
* `.hops(nodes, hops, to_fixed_point, direction)`

Fixes:
* local tester propagates `$@`

Infra:
* Passes lint + mypy + tests
* Switches to explicit imports
* Marks non-vectorized files as deprecated (should also do in class+method defs to further dissuade use..)


lmeyerov:
@silkspace @Orbifold  Landing as `0.21.0` 😎

PR #322, opened 2022-03-20
Cleanup

lmeyerov:
A lot!

- `.umap()`
- `.featurize()`
- `.build_dgl_graph()`
- heavy refactors
- py 3.10


lmeyerov:
@silkspace  making progress!

I got most of the way there for being able to do dependency-light version `pip install graphistry[umap-learn]; ... g.nodes(...).umap(featurize=False)` , but getting some umap sparsity errors, probably because too small of a test:

https://github.com/graphistry/pygraphistry/runs/5621738490?check_suite_focus=true

I want to move on to the with-heavier-dependencies versions (`feature=True` + `DGLMixin`) -- can you take a look at making the tests pass for `umap(kind='nodes', featurize=False)`  + `umap(kind='edges', featurize=False)` ?

This branch makes it a bit easier:

```bash
docker $ ./test-cpu-umap.sh
```

will call docker build + run the umap test on umap-learn with only pandas/numpy/pytest/umap-learn in the container . Likewise, the testsuite will auto-skip umap tests it lacks dependency for, e.g., it'll run new `TestUMAPMethods` and auto-skip `TestUMAPAIMethods`

graphistry/umap_utils.py
```diff
@@ -165,7 +165,7 @@ def umap(
         self,
         kind: str = "nodes",
         use_columns: Union[List, None] = None,
-        featurize: bool = True,
+        featurize: bool = False,
```
lmeyerov, line 168: @silkspace  is  disabling by default intentional? 

my thinking is default should be default on, and if libs not installed, we nudge them to install or disable. I can imagine an `auto` mode that tries to do the best for available installed packages, but seems not worth it yet..

graphistry/feature_utils.py
```diff
@@ -32,21 +35,22 @@
...
     has_dependancy = False
     scipy = (Any,)
     torch = Any
     SuperVectorizer = (Any,)
-    #SimilarityEncoder = (Any,)
-    #TargetEncoder = (Any,)
-    #MinHashEncoder = (Any,)
+    # SimilarityEncoder = (Any,)
+    # TargetEncoder = (Any,)
+    # MinHashEncoder = (Any,)
```
lmeyerov, line 53: Should these look more like `TargetEncoder : Any = None` ? Why is `Any` on the right-hand side, instead of just being a Type?

lmeyerov: If too annoying, do some sort of ` # type: ignore`

PR #333, opened 2022-04-13
Topology Aware Collapsable Graphs by Attributes

silkspace:
g = graphistry.nodes(...).edges(...)

g2 = g.collapse(start_node, attribute='some_attribute', column='some_column_in_nodes_dataframe')

g2.plot() -> produces a graph where all directed nodes with said attribute are collapsed while preserving topology of graph. Nodes without property have edges merged to super node, while all internal nodes collapse. Disconnected super clusters are not merged together, as is the trick with this algorithm. 

Traversal is inefficient for large graphs (scaling is between O(2*(n_edges)*(log(n_edges)) and 2*(n_edges)**(3/2))


graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    """
+    ndf = g._nodes
+    edf = g._edges
+    src = g._source
+    dst = g._destination
+    node = g._node
+    return ndf, edf, src, dst, node
+
+
+def get_children(g, node_id, hops=1):
```
lmeyerov, line 31: unncessary as just calling hop?

silkspace: Yes and then I have to set it each time with a dataframe? Less writing?

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    ----------------------------------------------------------------------------------
+
+    :param g: graphistry instance
+    :param node_id: `node` to find edges from
+    :param directed: bool, if true, finds all outgoing edges of `node`, default True
+    :param hops: the number of hops from `node` to take, default = 1
+    :returns DataFrame of edges
+    """
+    _, _, src, dst, _ = unpack(g)
+    g2 = get_children(g, node_id, hops=hops)
```
lmeyerov, line 81: need to pass direction to get_children/hops?

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    else:
+        if (n2 in edf[edf[src] == n1][dst].values) or (
+            n1 in edf[edf[src] == n2][dst].values
+        ):
+            return True
+    return False
+
+
+def get_edges_of_node(g, node_id, directed=True, hops=1):
+    """
```
lmeyerov, line 69: parameter `directed` currently seems to either pick all dst nodes, or all src+dst nodes

while the name `directed` makes it sound like more about the traversal direction (so an arg to `.hop()`) -- clarify the name/comment?

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    :param g: graphistry instance
+    :param node_id: `node` to find edges from
+    :param directed: bool, if true, finds all outgoing edges of `node`, default True
+    :param hops: the number of hops from `node` to take, default = 1
+    :returns DataFrame of edges
+    """
+    _, _, src, dst, _ = unpack(g)
+    g2 = get_children(g, node_id, hops=hops)
+    if directed:
+        edges = g2._edges[dst]
```
lmeyerov, line 83: need a `.drop_duplicates()` here too... move to be part of the return?

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+#
+#
+# def remove_edges(g, edgelist):
+#     # src_edges, dst_edges = np.array(edgelist).T
+#     ndf, edf, src, dst, node = unpack(g)
+#     tedf = edf[~edf[src].isin(edgelist) & ~edf[dst].isin(edgelist)]
+#     return tedf
+
+
+def get_edges_in_out_cluster(g, node_id, attribute, column, directed=True):
```
lmeyerov, line 105: mypy types

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    -----------------------------------
+
+    :param name: node name
+    :returns wrapped node name
+    """
+    name = str(name)
+    if WRAP in name:  # idempotency
+        return name
+    return f"{WRAP}{name}{WRAP}"
+
```
lmeyerov, line 209: so if `"~"` is in a node id, what happens?

graphistry/collapse.py
silkspace: Yes, agreed

graphistry/collapse.py
silkspace: 4 eyes are better than 2, thanks

graphistry/collapse.py
lmeyerov: hmm maybe we can make `hop` signature more convenient then:

ex:  this could be replaced with an even short `g.hop([node_id], hops)`, where we teach `hop()` that if instead of nodes=pd.DataFrame, it gets `List[str]`, it auto-coerces into a nodes df for them

lmeyerov: so `hop(self, nodes=Union[ pd.DataFrame, List[str] ], ...`

graphistry/collapse.py
silkspace: The age old problem of escaping/split/replace, open to suggestions -- might even have WRAP be parameter

`g.collapse(..., escaping_token='~')` 

?

lmeyerov: (that's ok, but maybe we can detect-and-throw at the beginning if that's the case?)

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+
+def melt(ndf, node) -> str:
+    """
+        Reduces node if in cluster store, otherwise passes it through.
+    ex:
+        node = "4" will take any sequence from get_cluster_store_keys, "1 2 3", "4 3" and returns "1 2 3 4"
```
lmeyerov, line 215: `node = "4"`... or `node = "3"` ? (trying to follow)

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    # THIS IS IMPORTANT FUNCTION -- it is where we wrap the parent/child in WRAP
+    # if child in cluster group, we melt it
+    ckey = in_cluster_store_keys(ndf, child)
+
+    if ckey:
+        new_parent_name = melt(ndf, child)
+    else:  # if not, then append child to parent as the start of a new cluster group
+        # might have to escape parent and child if node names are dumb eg, 'this value key'
+        new_parent_name = f"{wrap_key(parent)} {wrap_key(child)}"
+    if VERBOSE:
```
lmeyerov, line 260: `logger.debug(...)`

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    :param g: graphistry instance
+    :param parent: `node` with `attribute` in `column`
+    :param child: `node` with `attribute` in `column`
+    :returns: graphistry instance, new_parent_name of super cluster
+    """
+
+    ndf, edf, src, dst, node = unpack(g)
+
+    new_parent_name = get_new_node_name(ndf, parent, child)
+
```
lmeyerov, line 285: hmmm this is a mutation... probably better not to return anything

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    for node in incluster:
+        edf.loc[edf[src] == parent, COLLAPSE_SRC] = new_parent_name
+        edf.loc[edf[dst] == parent, COLLAPSE_DST] = new_parent_name
+
+        edf.loc[edf[dst] == node, COLLAPSE_DST] = new_parent_name
+        edf.loc[edf[src] == node, COLLAPSE_SRC] = new_parent_name
+
+    g._edges = edf
+    return g
```
lmeyerov, line 320: again, this is really a mutation of existing g, not returning a new g, so better to not return anything

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    -------------------------------------------------------------------------
+
+    :param g: graphistry instance
+    :returns graphistry instance
+    """
+    ndf, edf, src, dst, node = unpack(g)
+    if COLLAPSE_NODE not in ndf.columns:
+        ndf[COLLAPSE_NODE] = DEFAULT_VAL
+        ndf[node] = ndf[node].astype(str)
+        print(f'Converted ndf to type({type(ndf[node].values[0])})')
```
lmeyerov, line 350: logger.debug

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+        g._nodes = ndf
+    if COLLAPSE_SRC not in edf.columns:
+        edf[COLLAPSE_SRC] = DEFAULT_VAL
+        edf[COLLAPSE_DST] = DEFAULT_VAL
+        edf[src] = edf[src].astype(str)
+        edf[dst] = edf[dst].astype(str)
+        print(f'Converted edf to type({type(edf[src].values[0])})')
```
lmeyerov, line 357: logger.debug

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+def check_default_columns_present_and_coerce_to_string(g):
+    """
+        Helper to set COLLAPSE columns to nodes and edges dataframe, while converting src, dst, node to dtype(str)
+
+    -------------------------------------------------------------------------
+
+    :param g: graphistry instance
+    :returns graphistry instance
+    """
+    ndf, edf, src, dst, node = unpack(g)
```
lmeyerov, line 346: this is a mutation on input ndf, probably should be:

```python
ndf2 = ndf.copy(deep=False) # or True if later you mutate the cells.. which you seem to...
...
g2 = g.nodes(ndf2)
...
return g2 
```

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+        if has_property(g, parent, attribute, column):  # if (T, *)
+            # add start node to super node index
+            tkey = f"{parent} {parent}"  # it will reduce this to `parent` but can add to `seen`
+            if tkey not in compute_key:  # its love!
+                seen[tkey] = 1
+                g, _ = collapse_nodes(g, parent, parent)
+            if has_property(g, child, attribute, column):  # if (T, T)
+                # add to seen
+                seen[compute_key] = 1
+                if VERBOSE:
```
lmeyerov, line 409: logger.debug

lmeyerov: and `if logger.isEnabledFor(logging.DEBUG):  ...`

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+
+    :param g: graphistry instance
+    :param child: child node to start traversal, for first traversal, set child=parent or vice versa.
+    :param parent: parent node to start traversal, in main call, this is set to child.
+    :param attribute: attribute to collapse by
+    :param column: column in nodes dataframe to collapse over.
+    :returns graphistry instance with collapsed nodes.
+
+    """
+    g = check_default_columns_present_and_coerce_to_string(g)
```
lmeyerov, line 390: this gets called on every recursion, is that right? or just should be at entrypoint?

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    --------------------------------------------------------------------------------------------------------------------
+
+    :param g: graphistry instance
+    :param self_edges: bool, whether to keep duplicates from ndf, edf, default False
+    :param unwrap: bool, whether to unwrap node text with `~`, default True
+    :param remove_collapse: bool, whether to drop COLLAPSE columns from dataframe, default True
+    :returns final graphistry instance
+    """
+    # at the end of collapse, move anything untouched to new graph
+    ndf, edf, src, dst, node = unpack(g)
```
lmeyerov, line 455: these are mutations on ndf/edge, should probably be a `ndf.copy(deep=True)`

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+    from time import time
+
+    __doc__ = collapse.__doc__
+    
+    
+    
+    n_edges = len(g._edges)
+    complexity_min = int(2 * n_edges * np.log(n_edges))
+    complexity_max = int(2 * n_edges ** (3 / 2))
+    if VERBOSE:
```
lmeyerov, line 509: logger.debug

graphistry/collapse.py
```diff
@@ -0,0 +1,532 @@
...
+        ndf = ndf.drop(columns=[COLLAPSE_NODE])
+        edf = edf.drop(columns=[COLLAPSE_SRC, COLLAPSE_DST])
+
+    # set the dataframes
+    g._nodes = ndf
+    g._edges = edf
+    return g
+
+
+def collapse_by(g, parent, start_node, attribute, column, seen):
```
lmeyerov, line 488: do you have a sense of how many ~seconds this is running on say:

- 100 node / 1K edges
- 1K nodes / 10K edges
- 10K nodes / 100K edges

lmeyerov (changes requested):
Main changes:
- add mypy types
- print -> logger.debug for logger of name __name__
- for mutating funcs parents, `ndf2 = ndf.copy(deep=True); return g.nodes(ndf2)`
- for mutating funcs helpers, do not return `g`


Ideally, improve `hop()` sig instead of adding a new alias

Am curious about the perf on diff size graphs. OK if not 100K node grpahs, but would hope we can do 10K node (100K edge) ones. Not sure though, `hop()` etc may be slow as was meant for bulk calls, not individual.

graphistry/collapse.py
silkspace: Fixed

graphistry/collapse.py
silkspace: Fixed

graphistry/collapse.py
silkspace: I handle this in outer scope, in compute.py

`        from .collapse import collapse_by

        #__doc__ += collapse_by.__doc__
        g = self.bind()
        # TODO FIXME CHECK SELF LOOPS?
        return collapse_by(
            g,
            start_node=node,
            parent=node,
            attribute=attribute,
            column=column,
            seen={},
            self_edges=self_edges,
            unwrap=unwrap,
            remove_collapse=remove_collapse,
        )
`

graphistry/collapse.py
silkspace: Every recursion, but lightweight bools overall. I can move it out of scope, and change `collapse` -> `_collapse`

graphistry/collapse.py
silkspace: Get to this tomorrow in tests

graphistry/collapse.py
lmeyerov: @silkspace  `g = self.bind()` is a shallow copy: it'll do `res._nodes = self._nodes` instead of `res._nodes = self._nodes.copy()` . Likewise, I'm fuzzy on pandas semantics around mutation, so you may even need `nodes.copy(deep=True)`.

lmeyerov: a test can be:

```python
g = graphistry.edges(...)
g2 = g.bind()
g2._edges[...] = something # mutate...
print(g._edges)
```

graphistry/Plottable.py
```diff
@@ -87,6 +87,18 @@ def get_topological_levels(
         raise RuntimeError('should not happen')
         return self
 
+    def collapse(
```
lmeyerov, line 90: Is there some sort of outer loop that is supposed to be used in a typical case? I'd think we'd want 2 methods:

1. clarify via renaming the current entrypoint as `collapse_adjacent_by_col_val_match(g, root, col, val)`
  - later, we cab generalize to the still-vectorizable form `collapse_adjacent_by_dict_match(g, root, {col1: val1, col2: val2, ...})`
2. add typical outer loop `collapse_adjacent_by_col_match(g, root, col)`
  - basically: `for v in g._nodes[col].unique(): g = collapse(g, root, col, v)`
  - later do the same generalization

We can land without for now..

lmeyerov:
Merging!

I'll file follow-up tickets:
- generalize interface: pass in a kv-dict instead of kv-pair, and pass in a col list instead of kv-pair
- reductions, including provenance tracking
- optimization

PR #350, opened 2022-05-11
Dev/igraph layouts

lmeyerov:
Add:
* `g.compute_igraph()`, `g.layout_igraph()` to generically call igraph bits, either enriching or swapping the graph
* `g.scene_settings()` like `layout_settings()`, adds to current url object

Fix:
* `from_igraph()` was often adding a `'name'` nodes col instead of using `g._node` for IDs

---

Sample use:

```python
g2 = g.compute_igraph('betweenness').encode_point_size('betweenness')
g3 = g2.layout_igraph('sugiyama', directed=True).rotate(90).scene_settings(edge_curvature=0)
g3.plot()
```
Screenshot from 2022-05-11 16-29-59


graphistry/plugins/igraph.py
```diff
@@ -209,3 +214,164 @@ def to_igraph(self,
...
+
+def compute_igraph(
+    self: Plottable, alg: str, alg_as: Optional[str] = None, directed: Optional[bool] = None, params: dict = {}
+) -> Plottable:
+    """Enrich or replace graph using igraph methods
+
+    :param alg: Name of an igraph.Graph method like `pagerank`
+    :type alg: str
+
+    :param alg_as: For algorithms that compute values for nodes, which attribute to write to. If None, use the algorithm's name. (default None)
```
silkspace, line 257: Docstring is not clear in my first reading of it.

graphistry/plugins/igraph.py
```diff
@@ -209,3 +214,164 @@ def to_igraph(self,
...
+    if alg_as is None:
+        alg_as = alg
+
+    try:
+        ig = self.to_igraph(directed=True if directed is None else directed)        
+        out = getattr(ig, alg)(**params)
+    except NotImplementedError as e:
+        if directed is None:
+            ig = self.to_igraph(directed=False)        
+            out = out = getattr(ig, alg)(**params)
```
silkspace, line 281: Extra `out`?

silkspace (commented):
Good stuff

graphistry/plugins/igraph.py
lmeyerov: great, changing output cols to `out_col`, `x_out_col`, and `y_out_col`

lmeyerov: and adding examples

PR #354, opened 2022-05-21
Team phase1

lmeyerov:
Adds:

```python
graphistry.register(..., org_name='myorg') # optional
g.privacy(mode='organization') # optional
```


lmeyerov:
@vaimdev Can we make `.register(org_name='bad org or unauthorized org')` throw an exn?

lmeyerov:
@vaimdev  when a pygraphistry upload is to an org that a user does not have access to, instead of a `403`/`404`, it succeeds as part of their personal org

lmeyerov:
Failing due to api=2 protobuf stuff already fixed in ai branch, can land without

PR #359, opened 2022-06-06
Temp

lmeyerov:


lmeyerov:
@silkspace i'm seeing the same error in py3.6 + py3.9, any ideas? types etc seems fine, so sounds like logic: https://github.com/graphistry/pygraphistry/runs/6839481923?check_suite_focus=true 

would like to merge..

graphistry/util.py
```diff
@@ -21,7 +21,10 @@ def setup_logger(name, verbose=VERBOSE, fullpath=TRACE):
         FORMAT = " %(message)s\n"
     logging.basicConfig(format=FORMAT)
     logger = logging.getLogger(f'graphistry.{name}')
-    logger.setLevel(logging.INFO if verbose else logging.DEBUG)
+    if verbose is None:
```
lmeyerov, line 24: we need to kill `verbose` anti-patterns here, i'll file an issue to get this stuff normal

PR #365, opened 2022-06-18
Scraper

silkspace:


graphistry/scrape_utils.py
```diff
@@ -0,0 +1,357 @@
...
+import json
+from urllib.parse import urlparse
+from collections import Counter
+import pandas as pd
+
+import tempfile
+from datetime import datetime
+
+# pip3 install newspaper3k
+import newspaper
```
lmeyerov, line 14: can you `git mv` to `graphistry/plugins/newspaper.py` ?

graphistry/scrape_utils.py
```diff
@@ -0,0 +1,357 @@
...
+    import os
+    pathfile, ext = os.path.splitext(path_to_file)
+    pathfile += "_metadata"+ext
+    return pathfile
+    
+
+
+class Scraper:
+	"""
+	Takes in a url (and url to pdf) or document, if url, scrapes it, and returns a DataFrame of the scraped information.
```
lmeyerov, line 176: super cool!

maybe a usage example in here?

graphistry/scrape_utils.py
```diff
@@ -0,0 +1,357 @@
...
+        return None
+    try:
+        article.parse()
+    except newspaper.ArticleException:
+        logger.info(newspaper.ArticleException)
+        return None
+    return article
+
+
+def get_article_information(url):
```
lmeyerov, line 89: maybe we make a named return type for these?

graphistry/scrape_utils.py
```diff
@@ -0,0 +1,357 @@
...
+}
+
+DATAFRAME_SCHEMA = {
+    "article_title": "",
+    "document": "",
+    "summary": "",
+    "meta_title": "",
+    "hashes": "",
+    "url": "",
+}
```
lmeyerov, line 46: maybe we make a named return type for these?

graphistry/scrape_utils.py
```diff
@@ -0,0 +1,357 @@
...
+	
+	def _get_session_metadata(self):
+		if os.path.exists(self.tracker_file_metadata):
+			with open(self.tracker_file_metadata, "r") as f:
+				metadata = json.load(f)
+			return metadata
+		return {'hashes':[], 'url_counts': Counter()}
+
+
+def ParallelScraper(urls, tracker_file, n_jobs=-1, *args, **kwargs):
```
lmeyerov, line 342: ok for now, though for server support, we'd want to switch to dask

graphistry/scrape_utils.py
```diff
@@ -0,0 +1,357 @@
...
+
+import tempfile
+from datetime import datetime
+
+# pip3 install newspaper3k
+import newspaper
+
+# pip install textract
+import textract
+
```
lmeyerov, line 18: maybe a new `setup.py` section for `pip install graphistry[scraping]` ?

graphistry/scrape_utils.py
lmeyerov: Also, may be good to separate processes vs threads here, as i'm guessing mix of compute threads vs async network io here, like `n` jobs with `k` threads. Can defer till dask impl. Relevant: async io interface: https://stackoverflow.com/questions/68360767/newspaper3k-get-articles-from-html-instead-of-url

graphistry/scrape_utils.py
lmeyerov: looking further, we convert to pandas, at which point an apache arrow schema may make sense for dynamic typing

(trick: convert pd df to arrow, get arrow schema, then bake that generated arrow schema into code)

lmeyerov (changes requested):
main things:
- move from main dir into plugins
- setup.py

Anything recommended for testing?

lmeyerov:
We can revisit this later esp wrt Louie

PR #369, opened 2022-07-06
Publishing to conda

Blanderbuss:
To install all:
`conda install -c pygraphistry_dev graphistry-dev`
To install one package:
`conda install -c pygraphistry_dev graphistry_dev.graphistry_core`
Right now action is only run manually. 
Currently configured for the dev anaconda account


.github/workflows/publish-conda.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: Build and publish Python 🐍 distributions 📦 to Conda
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: publish-to-conda
+      uses: blanderbuss/conda-publish-action@v1.4
+      with:
+        subDir: 'conda'
+        AnacondaToken: ${{ secrets.ANACONDA_TOKEN }}
+        platforms: 'linux'
```
lmeyerov, line 21: Is there a prepublish test we can do? Ex: In pip, we do a test publish to `test.pypi.org` first

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
+package:
+  name: graphistry-dev
+  version: "0.25.2"
```
lmeyerov, line 3: The version seems hard-coded here, while it should be taken from the git tag -- am I misreading somehow?

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - typing-extensions
+      - numpy
+      - pandas>=0.17.0
+      - protobuf>=2.6.0
+      - pyarrow>=0.15.0
+      - requests
+  - name: graphistry_dev.graphistry-umap
+    requirements:
+      - umap-learn
+      - dirty_cat
```
lmeyerov, line 39: See `setup.py` for min req

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - name: graphistry_dev.graphistry-ai
+    requirements:
+      - scikit-learn
```
lmeyerov, line 42: see `setup.py` for min reqs

conda/meta.yaml
lmeyerov: Maybe: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#templating-with-jinja

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - flake8
+    - mock
+    - mypy
+    - pytest
+    - pandas-stubs
+    - types-requests
+
+outputs:
+  - name: graphistry_dev.graphistry_core
+    requirements:
```
lmeyerov, line 28: is there anything like `outputs.*.description` where they can point to the main package?

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - umap-learn
+      - dirty_cat
+      - dgl
+      - pytorch
+      - sentence-transformers
+      
+      
+about:
+  home: https://www.graphistry.com/
+  license: BSD
```
lmeyerov, line 53: BSD3 + license file

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  summary: Visual Graph Analytics library for Python
```
lmeyerov, line 54: Graphistry Visual Graph ...

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  description: |
+    Graphistry is a visual graph analytics library for extracting, transforming, 
+    displaying, and sharing big graphs with end-to-end GPU acceleration
+  doc_url: https://www.graphistry.com/docs
```
lmeyerov, line 58: [https](http://hub.graphistry.com/docs)

conda/meta.yaml
```diff
@@ -0,0 +1,65 @@
+package:
+  name: graphistry-dev
```
lmeyerov, line 2: Guessing as part of landing, change to `graphistry` ?

lmeyerov (changes requested):
Getting there! See per-line reviews:

- [x] Parameterize version by git tag (we use `X.Y.Z`, in case conda needs a prefix)
- [x] Update dependencies, esp. min versions
- [ ] Various individual settings changes

And some questions:
- [ ] How should pre-publishing testing work?
- [ ] Can we put instructions somewhere to guide users on different ways to install?

.github/workflows/publish-conda.yml
Blanderbuss: After researching, I don't belive that there is a way to perform a test publish for conda

conda/meta.yaml
Blanderbuss: Working on it, for some reason this env variable is not present

conda/meta.yaml
Blanderbuss: Done

conda/meta.yaml
Blanderbuss: We can put instructions in the main description and it will be used for all outputs

conda/meta.yaml
Blanderbuss: Done

conda/meta.yaml
Blanderbuss: Done

conda/meta.yaml
Blanderbuss: Done

conda/meta.yaml
Blanderbuss: Resolved by switching to different github action

conda/meta.yaml
Blanderbuss: You can check [here ](https://anaconda.org/pygraphistry_dev/graphistry.graphistry-core) to see how the description looks like right now

Blanderbuss:
@lmeyerov Answering you question about testing:
- Conda [runs tests](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#meta-test) after the build if there is a `run_test.[py,pl,sh,bat]` file in the recipe. If you want to test it without publishing locally, you can use `--no-anaconda-upload` option.

Other issues have been resolved as well as the possible answer to instruction, you can check them.

conda/meta.yaml
```diff
@@ -0,0 +1,83 @@
+package:
+  name: graphistry
+  version: {{ GIT_DESCRIBE_TAG }}
+
+source:
+  git_rev: 0.25.2
```
lmeyerov, line 6: Should this also be a variable?

Blanderbuss: Changed it to a variable

lmeyerov (commented):
lgtm - will try on prod!

PR #391, opened 2022-09-06
cuml automagic or engine flag

dcolinmorgan:
automatic detection if cuml is present to use for umap engine, otherwise defaults to umap


lmeyerov:
@dcolinmorgan  the heavier ai dependencies switched to lazy loading last week (pay-as-you-go) vs top-level : can you update your base branch and make.any rapids imports deferred until necessary?

lmeyerov (changes requested):
See comment - we switched to lazy imports vs top-level for our 6 heaviest bc was.taking 10s-2min to import in real scenarios , so update base branch and defer cuml import as late as possible?

PR #397, opened 2022-09-09
main cuml umap

dcolinmorgan:
- a few stray prints (couldnt get logger to work) but good for testing
- now main branch has cuml umap with `g.umap(engine='cuml')`, otherwise defaults to umap-learn (also takes that as engine arg)

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,13 @@ It is easy to turn arbitrary data into insightful graphs. PyGraphistry comes wit
     graphistry.edges(edges, 'src', 'dst').plot()
     ```
 
+* GPU [RAPIDS.ai](https://www.rapids.ai) cuml
+
+    ```python
+    g = graphistry.nodes(G)
+    g.umap(engine='cuml',metric='euclidean')
+    ```
+
 * GPU [RAPIDS.ai](https://www.rapids.ai) cugraph ([notebook demo](demos/demos_databases_apis/gpu_rapids/cugraph.ipynb))
 
     ```python
@@ -359,7 +366,7 @@ Automatically and intelligently transform text, numbers, booleans, and other for
 
 See `help(g.featurize)` for more options
 
-### [UMAP](https://umap-learn.readthedocs.io/en/latest/)
+### [sklearn-based UMAP](https://umap-learn.readthedocs.io/en/latest/), [cuML-based UMAP](https://docs.rapids.ai/api/cuml/stable/api.html?highlight=umap#cuml.UMAP)
 
 * Reduce dimensionality and plot a similarity graph from feature vectors:
 
@@ -382,6 +389,12 @@ See `help(g.featurize)` for more options
 
     ```python
       g.umap(kind='nodes', X=['col_1', ..., 'col_n'], y=['label', ..., 'other_targets'], ...)
+    ```
+
+* UMAP supports different backend `engine`, which by default exploit an availabile GPU, resulting in marked speed increases, and falling back to CPU processing:
+
+    ```python
+      g.umap(engine='cuml')
     ```
 
 You can also featurize edges and UMAP them as we did above.
--- a/README.md
+++ b/README.md
@@ -178,11 +178,12 @@ It is easy to turn arbitrary data into insightful graphs. PyGraphistry comes wit
     graphistry.edges(edges, 'src', 'dst').plot()
     ```
 
-* GPU [RAPIDS.ai](https://www.rapids.ai) cuml
+* GPU [RAPIDS.ai](https://www.rapids.ai) cuML
 
     ```python
+    g = graphistry.nodes(cudf.read_csv('rows.csv'))
     g = graphistry.nodes(G)
-    g.umap(engine='cuml',metric='euclidean')
+    g.umap(engine='cuml',metric='euclidean').plot()
     ```
 
 * GPU [RAPIDS.ai](https://www.rapids.ai) cugraph ([notebook demo](demos/demos_databases_apis/gpu_rapids/cugraph.ipynb))
@@ -391,7 +392,7 @@ See `help(g.featurize)` for more options
       g.umap(kind='nodes', X=['col_1', ..., 'col_n'], y=['label', ..., 'other_targets'], ...)
     ```
 
-* UMAP supports different backend `engine`, which by default exploit an availabile GPU, resulting in marked speed increases, and falling back to CPU processing:
+* `umap(engine="...")` supports multiple implementations. It defaults to using the GPU-accelerated `engine="cuml"` when a GPU is available, resulting in orders-of-magnitude speedups, and falls back to CPU processing via `engine="umap"`.:
 
     ```python
       g.umap(engine='cuml')
--- a/demos/demos_databases_apis/gpu_rapids/part_iv_gpu_cuml.ipynb
+++ b/demos/demos_databases_apis/gpu_rapids/part_iv_gpu_cuml.ipynb
@@ -1,0 +1,307 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### clone and install graphistry, print version"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {
+    "vscode": {
+     "languageId": "python"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "import pandas as pd, networkx as nx\n",
+    "# !git clone https://github.com/graphistry/pygraphistry.git\n",
+    "\n",
+    "from time import time\n",
+    "!pip install -U pygraphistry/ --quiet\n",
+    "\n",
+    "import graphistry \n",
+    "graphistry.register(api=3,protocol=\"https\", server=\"hub.graphistry.com\", username='dcolinmorgan', password='f5UwthGEF@F@xnP')\n",
+    "graphistry.__version__"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### test engines on same synthetic dataset"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "vscode": {
+     "languageId": "python"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "['time: 0.011002473036448161 line/min: 226494.5337057124']\n"
+     ]
+    }
+   ],
+   "source": [
+    "N,P = 1000, 0.005\n",
+    "A = nx.generators.random_graphs.gnp_random_graph(N, P)\n",
+    "G=nx.to_pandas_edgelist(A)\n",
+    "G=G.rename(columns={\"target\":'destination'})\n",
+    "G['edge_weight']=np.random.randn(len(G))\n",
+    "\n",
+    "g = graphistry.nodes(G)\n",
+    "t=time()\n",
+    "g2 = g.umap(engine='cuml')\n",
+    "min=(time()-t)/60\n",
+    "lin=G.shape[0]/min\n",
+    "print(['time: '+str(min)+' line/min: '+str(lin)])\n",
+    "# g2.plot()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {
+    "vscode": {
+     "languageId": "python"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "['time: 0.14623186588287354 line/min: 17041.42927367146']\n"
+     ]
+    }
+   ],
+   "source": [
+    "# N,P = 1000, 0.005\n",
+    "# G = nx.generators.random_graphs.gnp_random_graph(N, P)\n",
+    "G=nx.to_pandas_edgelist(A)\n",
+    "G=G.rename(columns={\"target\":'destination'})\n",
+    "G['edge_weight']=np.random.randn(len(G))\n",
+    "\n",
+    "g = graphistry.nodes(G)\n",
+    "t=time()\n",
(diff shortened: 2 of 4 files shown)


lmeyerov (changes requested):
[ ] `engine : UMAPEngine` (public interfaces) + `engine_resolved : UMAPEngineCOncrete` (internal calls) so users can pick which engine, default to `"auto"`. Match how `featurize()` does it.
[ ] print -> logger
[ ] un-delete `feature` threading through params (needed for memoization)

In a follow-on PR, let's add an ipynb demo!

graphistry/umap_utils.py
```diff
@@ -144,8 +160,21 @@ def umap_lazy_init(
...
             self.repulsion_strength = repulsion_strength
             self.negative_sample_rate = negative_sample_rate
-            self._umap = umap.UMAP(**umap_kwargs)
+            # self._umap = umap.UMAP(**umap_kwargs)
             self.umap_initialized = True
+            # self.engine = engine
+
+            # self._umap = umap.UMAP(**umap_kwargs)
+            # if self.engine == 'cuml':
+            if 'cuml' in sys.modules:
```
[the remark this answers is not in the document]
dcolinmorgan: i assume 'conrete_umap_engine' here == 'resolve_umap_engine' you described above? the example feature_util has no 'conrete_feature_engine' analog

graphistry/umap_utils.py
```diff
@@ -31,17 +31,57 @@ def lazy_umap_import_has_dependancy():
...
+        raise import_umap_exn
+
+def assert_imported_cuml():
+    has_cuml_dependancy_, import_cuml_exn, _ = lazy_cuml_import_has_dependancy()
+    if not has_cuml_dependancy_:
+        logger.error("cuML not found, trying running "
+                     "`pip install cuml`")
+        raise import_cuml_exn
+
+UMAPEngineConcrete = Literal["none", "cuml", "umap_learn"]
```
lmeyerov, line 62: `"none"` seems like an illegal value in this case -- how do we do a umap w/ no engine?

(it's valid for feature engineering b/c that means a no-op passthrough of features)

graphistry/umap_utils.py
```diff
@@ -119,12 +167,18 @@ def umap_lazy_init(
...
         n_components: int = 2,
         metric: str = "euclidean",
+        engine: str = "auto", #this seems to override user input
+        # has_umap: str = "True",
     ):
 
         # FIXME remove as set_new_kwargs will always replace?
-        has_umap, _, umap = lazy_umap_import_has_dependancy()
+        engine_resolved = resolve_umap_engine(engine)
+        if engine_resolved == "umap_learn":
```
lmeyerov, line 176: cleaner like

```python
if engine_resolved == "umap_learn":
    _, _, umap_engine = ...
elif engine_resolved == "umap":
   _, _, umap_engine = ..
else:
    raise ValueError('No umap engine, ensure "auto", "umap", or "cuml", and the library is installed`)
```

b/c later guaranteed to have it and have the obj, and can do any conditionals on engine

graphistry/umap_utils.py
```diff
@@ -144,8 +199,15 @@ def umap_lazy_init(
             self.local_connectivity = local_connectivity
             self.repulsion_strength = repulsion_strength
             self.negative_sample_rate = negative_sample_rate
-            self._umap = umap.UMAP(**umap_kwargs)
             self.umap_initialized = True
+            self.engine = engine
+
+            if engine_resolved == "umap_learn":
+                del umap_kwargs['engine']
```
lmeyerov, line 206: confusing for dels to be here, can be part of umap_kwargs construction

graphistry/umap_utils.py
```diff
@@ -144,8 +199,15 @@ def umap_lazy_init(
             self.local_connectivity = local_connectivity
             self.repulsion_strength = repulsion_strength
             self.negative_sample_rate = negative_sample_rate
-            self._umap = umap.UMAP(**umap_kwargs)
             self.umap_initialized = True
+            self.engine = engine
```
lmeyerov, line 203: save resolved engine here

graphistry/umap_utils.py
```diff
@@ -361,7 +434,10 @@ def umap(
         :return: self, with attributes set with new data
         """
         assert_imported()
+        assert_imported_cuml()
         self.umap_lazy_init()
```
lmeyerov, line 438: pass engine

graphistry/umap_utils.py
```diff
@@ -361,7 +434,10 @@ def umap(
...
+        engine_resolved = resolve_umap_engine(engine)
+        
 
         self.suffix = suffix
         umap_kwargs = dict(
```
lmeyerov, line 443: need engine here?

graphistry/umap_utils.py
```diff
@@ -361,7 +434,10 @@ def umap(
         :return: self, with attributes set with new data
         """
         assert_imported()
+        assert_imported_cuml()
         self.umap_lazy_init()
+        engine_resolved = resolve_umap_engine(engine)
+        
 
         self.suffix = suffix
```
lmeyerov, line 442: all assignments should be on `self`, and after `if inplace: res = self else res = self.bind()`...

dcolinmorgan: ok i missed that one point and have changed it in the wimpiest commit ever

graphistry/umap_utils.py
```diff
@@ -181,6 +210,8 @@ def umap_fit(self, X: pd.DataFrame, y: Union[pd.DataFrame, None] = None):
         mins = (time() - t) / 60
         logger.info(f"-UMAP-ing took {mins:.2f} minutes total")
         logger.info(f" - or {X.shape[0]/mins:.2f} rows per minute")
+        print(f"-UMAP-ing took {mins:.2f} minutes total")
+        print(f" - or {X.shape[0]/mins:.2f} rows per minute")
```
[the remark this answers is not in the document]
silkspace: @dcolinmorgan if you want to have logging set `VERBOSE=True` in constants.py  https://github.com/graphistry/pygraphistry/blob/master/graphistry/constants.py

graphistry/umap_utils.py
```diff
@@ -119,21 +160,30 @@ def umap_lazy_init(
...
+        elif engine_resolved =='cuml':
+            _, _, umap_engine = lazy_cuml_import_has_dependancy()
+        else:
+            raise ValueError("No umap engine, ensure 'auto', 'umap', or 'cuml', and the library is installed")
+        
+        if has_umap and not self.umap_initialized: 
             umap_kwargs = dict(
                 n_components=n_components,
-                metric=metric,
+                # metric=metric,
```
lmeyerov, line 179: why remove metric?

if a problem for cuml, we can throw a targeted exn if engine is cuml

graphistry/umap_utils.py
```diff
@@ -229,10 +280,8 @@ def _process_umap(
         """
             Returns res mutated with new _xy
         """
-        _, _, umap = lazy_umap_import_has_dependancy()
-        # need this function to use memoize
-        res._umap = umap.UMAP(**umap_kwargs)
-
+        res._umap = self._umap
```
lmeyerov, line 283: @silkspace  is reusing `self._umap` here right?

lmeyerov: afaict `def umap(...)` calls lazy_init, which calls `self.umap = UMAP(**umap_kwargs)` , so indeed not needed here

lmeyerov (changes requested):
Fixes:
- [ ] ensure `def umap(self, ...)` immediately switches to `res = ..`, so subsequent calls do not mutate original `self` (lazy_init...) in typical case of `in_place=False`. RN, I think it'll `self._umap = ...` . 
- [ ] memoization seems to be dropping impactful params like metric & engine

Recommended:
- [ ] surprisingly typed (and unnecessary?) `has_umap : str = "True"`
- [ ] notes on ipynb
- [ ] add to CHANGELOG and to README

Follow-on PR?
- [ ] if input is a pandas df yet engine is cuml, can we auto-coerce pdf to cudf (and keep cudf after)?

graphistry/umap_utils.py
dcolinmorgan: as far as I can tell this flag is still an open issue, with some hacks to get it working https://github.com/rapidsai/cuml/issues/1653#issuecomment-765652438. From what I can tell there is no such flag listed in the UMAP function https://docs.rapids.ai/api/cuml/stable/api.html?highlight=umap#cuml.UMAP

lmeyerov: yeah that was my guess why you did it. it ends up big for umap users, so to support mult engines, a pattern can be like:

```python
opts = {
  a=b,
  c=d,
  **({'x': 'y'} if engine='umap' else {})
}
```

README.md
```diff
@@ -178,6 +178,13 @@ It is easy to turn arbitrary data into insightful graphs. PyGraphistry comes wit
     graphistry.edges(edges, 'src', 'dst').plot()
     ```
 
+* GPU [RAPIDS.ai](https://www.rapids.ai) cuml
+
+    ```python
+    g = graphistry.nodes(G)
+    g.umap(engine='cuml',metric='euclidean')
```
lmeyerov, line 185: `g.umap(engine='cum').plot()` :)

README.md
```diff
@@ -178,6 +178,13 @@ It is easy to turn arbitrary data into insightful graphs. PyGraphistry comes wit
     graphistry.edges(edges, 'src', 'dst').plot()
     ```
 
+* GPU [RAPIDS.ai](https://www.rapids.ai) cuml
+
+    ```python
+    g = graphistry.nodes(G)
```
lmeyerov, line 184: `g = graphistry.nodes(cudf.read_csv('rows.csv'))`

README.md
```diff
@@ -178,6 +178,13 @@ It is easy to turn arbitrary data into insightful graphs. PyGraphistry comes wit
     graphistry.edges(edges, 'src', 'dst').plot()
     ```
 
+* GPU [RAPIDS.ai](https://www.rapids.ai) cuml
```
lmeyerov, line 181: `cuML`

README.md
```diff
@@ -384,6 +391,12 @@ See `help(g.featurize)` for more options
       g.umap(kind='nodes', X=['col_1', ..., 'col_n'], y=['label', ..., 'other_targets'], ...)
     ```
 
+* UMAP supports different backend `engine`, which by default exploit an availabile GPU, resulting in marked speed increases, and falling back to CPU processing:
```
lmeyerov, line 394: > `umap(engine="...")` supports multiple implementations. It defaults to using the GPU-accelerated `engine="cuml"` when a GPU is available, resulting in orders-of-magnitude speedups, and falls back to CPU processing via `engine="umap"`.

graphistry/umap_utils.py
```diff
@@ -432,6 +430,9 @@ def umap(
         else:
```
lmeyerov, line 430: I think this needs to move up before the lazy initialization too, right, because otherwise all the self.suffix and lazy init calls will mutate `self` (and those should be called on `res` instead of `self`)

@silkspace any reason that's not the case?

silkspace: Working with Daniel now, the engine should be resolved inside the constructor, and bound to the res.engine after the fact.

graphistry/umap_utils.py
```diff
@@ -160,31 +161,28 @@ def umap_lazy_init(
...
-                local_connectivity=local_connectivity,
-                repulsion_strength=repulsion_strength,
-                negative_sample_rate=negative_sample_rate,
-                # engine=engine,
-            )
+
+        if not self.umap_initialized: 
+            umap_kwargs = dict({
+                'n_components':n_components,
+                **({'metric':metric} if engine_resolved=="umap_learn" else {}),
```
silkspace, line 178: cool

graphistry/umap_utils.py
lmeyerov: @silkspace but isn't that unsafe?

```python
g1 = g.umap(engine='cuml')
g2 = g.umap(engine='umap')
assert g1.engine != g2.engine
```

demos/demos_databases_apis/gpu_rapids/part_iv_gpu_cuml.ipynb
```diff
@@ -0,0 +1,705 @@
...
+   ],
+   "source": [
+    "import pandas as pd, networkx as nx\n",
+    "# !git clone https://github.com/graphistry/pygraphistry.git\n",
+    "\n",
+    "from time import time\n",
+    "!pip install -U pygraphistry/ --quiet\n",
+    "\n",
+    "import graphistry \n",
+    "graphistry.register(api=3,protocol=\"https\", server=\"hub.graphistry.com\", username='dcolinmorgan', password='f5UwthGEF@F@xnP')\n",
```
lmeyerov, line 56: @dcolinmorgan  can you scrub your pwd and reset your hub acct?

graphistry/umap_utils.py
lmeyerov: Thinking more: maybe what's needed here is throwing an exn if a non-cuml-allowed val? or would cuml do for us, and we can always put in?

lmeyerov:
FYI, a bunch of fixes on main branch in case your CI has a lot of false positives. Still a false positive from py3.10 around mypi, but that should be it.

graphistry/umap_utils.py
silkspace: I see where I was confused, yes, i was commenting on the previous 

`res.engine = engine`

which was changed to 

`res.engine = resolve_engine(engine)`

But you were referring to the `lazy_umap_init` code

silkspace: put
```
res.umap_lazy_init(engine=engine)  after the code blocks
        if inplace:
            res = self
        else:
            res = self.bind()
```
that will ensure that all settings are placed in res, no matter what.
also then there is no need to have
`res.engine = resolve_umap_engine(engine)` after the fact, as it has been set in self which is res

(confused on what part we were referring), all setting should be set in constructor, which in this setting happens in `umap_lazy_init(..)`

PR #404, opened 2022-09-28
`G.embed()` Base

tanmoyio:
### work in progress

This PR will integrate a simple way to perform Knowledge graph embedding on Graphistry graphs. 
`HeterographEmbedModuleMixin` this class should be inherited by the main Graphistry graph class. 

```python3
G.embed('TransE', d=128).plot()
```

Additional node features will get augmented with the embeddings after we perform `featurize()` and `umap()` on them

#### Protocols
- [x] TransE
- [ ] TransD
- [ ] ComplEx
- [ ] RotatE
- [ ] pRotatE
- [ ] DistMult

[probably need to integrate autoML for best performance]
[GPU/TPU compatibility will be added after the baseline model]
[A working code snippet on FB15k dataset will be published after completion of the PR]

### DIFF
--- a/graphistry/embed_utils.py
+++ b/graphistry/embed_utils.py
@@ -1,0 +1,117 @@
+import torch
+import torch.nn as nn
+from torch.utils.data import Dataset, DataLoader
+
+class HeterographEmbedModuleMixin(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+        self.protocol = {
+                'TransE': self.TransE
+        }
+
+    def embed(self, proto='TransE', d=128, gamma=12):
+
+        # initialize hparams (from arguments)
+        self.EPS = 2.0
+        self.d = d
+        self.gamma = gamma
+        self.emrange = gamma + self.EPS / d
+
+        # initializing the model
+        em_mod = HeteroEmbed(eps, d, gamma, erange, self.protocol[proto])
+
+        # initialize the dataloader
+        # TODO: prepare entity2id <- from graphistry graph
+        # TODO: prepare relation2id <- from graphistry graph
+        dataset = EmbedDataset(entity2id, relation2id)
+        train_generator = DataLoader(dataset, batch_size=32) # need autoscale batch for gpu?
+
+        # training loop
+        # optim (TODO: ADAM)
+        optim = optim.SGD(em_mod.parameters(), lr=0.01)
+
+        for _ in range(10):
+        
+            model.train()
+            for h, r, t in train_generator:
+
+                optim.zero_grad()
+
+                pos_triples = torch.stack((h, r, t), dim=1)
+                neg_triples = _ #TODO: need to generate 
+
+                # negative sample generation
+                h_o_t = torch.randint(high=2, size=h.size())
+                random_nodes = torch.randint(high=len(entity2id), size=h.size())
+                neg_h = torch.where(h_o_t == 0, random_nodes, h)
+                neg_t = torch.where(h_o_t == 1, random_nodes, t)
+                neg_triples = torch.stack((neg_h, r, neg_t), dim=1)
+
+                loss = model(pos_triples, neg_triples)
+                loss.mean().backward()
+
+                optim.step()
+
+        # out vectors
+        # --------------------------------------
+        # exception handling will be added later
+
+
+    def TransE(self, h, r, t):
+        o = h + r - t
+        return self.gamma.item() - torch.norm(o, p=1, dim=2)
+
+    def TransD(self, h, r, t):
+        return
+
+
+class HeteroEmbed(nn.Module):
+    def __init__(self, eps, d, gamma, erange, proto):
+        super().__init__()
+
+        self.erange = erange
+        
+        self.node_em = nn.Embedding(_,d) # _ will be num_nodes
+        self.edge_em = nn.Embedding(_,d) # _ will be num_edges
+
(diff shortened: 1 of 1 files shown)


graphistry/embed_utils.py
```diff
@@ -0,0 +1,117 @@
+import torch
+import torch.nn as nn
+from torch.utils.data import Dataset, DataLoader
```
lmeyerov, line 3: @silkspace  can you walk him through lazy imports?

graphistry/embed_utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+class HeterographEmbedModuleMixin(nn.Module):
```
lmeyerov, line 5: not sure how we want to do class-level lazy imports, I think I had to do that elsewhere already for `nn`, but may have been sloppy just to push initial lazy loader through

graphistry/embed_utils.py
tanmoyio: @lmeyerov Thanks for mentioning, I have seen it, will integrate it. Need to add many more things like, exception handling, asserts, unit tests. I will work on those once I set up the pipeline 😃

graphistry/PlotterBase.py
```diff
@@ -199,7 +199,7 @@ def __init__(self, *args, **kwargs):
 
 
     def __repr__(self):
-        bindings = ['edges', 'nodes', 'source', 'destination', 'node', 
+        bindings = ['edges', 'nodes', 'source', 'destination', 'relation', 'node',
```
lmeyerov, line 202: can we call this edge_type ?

lmeyerov: within GNN side, we might want to create a different data rep than 1 big edge table + 1 big node table, e.g., type-indexed property tables, and the same for triple versions like <src.id, src.type, src.col1, ...., edge.id, edge.type, edge.col1 ... dst....> . Will be curious how that shapes up (scale concerns), for now we can hide from the user and do purpose-specific...

graphistry/embed_utils.py
```diff
@@ -40,18 +43,20 @@ def embed(self, proto='TransE', d=128, gamma=12):
...
 
         # training loop
-        # optim (TODO: ADAM)
-        optim = torch.optim.SGD(em_mod.parameters(), lr=0.01)
+        optim = torch.optim.Adam(em_mod.parameters(), lr=0.01)
 
         # TODO: define epoch [AUTOML]
-        for _ in range(10):
 
+        epoch = trange(1000, leave=True)
```
lmeyerov, line 55: yes, exposing params both for users (smart overridable defaults) + later optuna is A++

graphistry/embed_utils.py
```diff
@@ -68,22 +73,31 @@ def embed(self, proto='TransE', d=128, gamma=12):
...
     def TransE(self, h, r, t):
-        gamma = 12
-        o = h + r - t
-        return gamma - (o/o.max()) # will be replaced by 2 norm later
+        return (h + r - t).norm(p=1, dim=1)
 
     def TransD(self, h, r, t):
         return
 
+    def DistMult(self, h, r, t):
```
lmeyerov, line 94: interest q: how to let users override their own  loss func, dist func, objective, etc

graphistry/embed_utils.py
silkspace: 1) Eventually: add early stopping after loss fans out, etc

silkspace: 1) Eventually: add early stopping after loss fans out, etc

graphistry/embed_utils.py
```diff
@@ -0,0 +1,153 @@
...
+                loss_acc += loss.mean().item()
+                c += 1
+                optim.step()
+
+            epoch.set_description(f"loss: {loss_acc/c}")
+            epoch.refresh()
+
+        self.node_em = em_mod.node_em.weight.detach().numpy()
+        self.edge_em = em_mod.edge_em.weight.detach().numpy()
+        return self.node_em, self.edge_em
```
silkspace, line 85: return a new graphistry instance, rather than vectors, so we can continue with the fluent api.

graphistry/embed_utils.py
silkspace: Great q:  perhaps 

```from torch import nn
g.embed(model: ConcreteLossType = 'str' | nn.LogSigmoid )```

with ConcreteLossType checking that, if not string, it sends in a proper loss function (y_true, y_predicted)

File: graphistry/embed_utils.py (hunk 6 above)
Comment (lmeyerov) [reply graphistry/embed_utils.py:85]: so just like `_node_features`, we have `_node_embedding` with some variable dim, and `.embed()` can memoize on features, model -> embed?

File: graphistry/embed_utils.py (hunk 5 above)
Comment (tanmoyio) [reply graphistry/embed_utils.py:109]: added custom distance function support, now you can use something like this
```python3
def custom_dist(h, r, t):
    return (h + r - t).norm(dim=1)

G.embed(proto=custom_dist)
```

graphistry/embed_utils.py
tanmoyio: now it returns a new graphistry instance with `G_node_features = [G_old_node_features | G_relational_embed_features]`

lmeyerov: @silkspace should we concat or add `G._{node,edge}_embedding` ?

PR #413, opened 2022-10-13
Cuml

lmeyerov:


graphistry/umap_utils.py
```diff
@@ -197,8 +205,11 @@ def umap_lazy_init(
             self._umap = umap_engine.UMAP(**umap_kwargs)
             self.umap_initialized = True
             self.engine = engine_resolved
-            self.suffix = suffix
-
+            if (engine_resolved=='cuml'):# and (float(umap_engine.__version__[:5])<22.06):
+                self.suffix = float(umap_engine.__version__[:5]) ##
+            else:
+                self.suffix = suffix
+
```
lmeyerov, line 212: I think field suffix is reserved for some other use? No need to save the version name, can dynamically look.up...

graphistry/umap_utils.py
```diff
@@ -222,14 +233,37 @@ def umap_fit(self, X: pd.DataFrame, y: Union[pd.DataFrame, None] = None):
         y = self._check_target_is_one_dimensional(y)
         logger.info('-' * 90)
         logger.info(f"Starting UMAP-ing data of shape {X.shape}")
+        # mod_ver=eval(self._umap().__module__.split('.')[0]).__version__
+        if (self.engine=='cuml') and (self.suffix<22.06): #(mod_ver<22.06):
+            from cuml.neighbors import NearestNeighbors
+            import cupy
+            logger.info(f"using cuml<22.06 requires setting knn_graph, default n_neighbors=5. try upgrading `cuml` or using `umap_learn`")
+            k=5
+            knn = NearestNeighbors(k)
```
lmeyerov, line 242: Instead of hardcoding/adding new k, use existing  n_neighbors?

graphistry/umap_utils.py
```diff
@@ -40,24 +41,30 @@ def lazy_cuml_import_has_dependancy():
...
-    has_cuml_dependancy_, import_cuml_exn, _ = lazy_cuml_import_has_dependancy()
+    has_cuml_dependancy_, import_cuml_exn, cuml = lazy_cuml_import_has_dependancy()
     if not has_cuml_dependancy_:
-        logger.error("cuML not found, trying running "
+        logger.warning("cuML not found, trying running "
                      "`pip install cuml`")
         raise import_cuml_exn
+    elif (has_cuml_dependancy_) and (float(cuml.__version__.rsplit('.',1)[0])>=min_cuml_version):
+        logger.warning('cuml engine specified, but only version '+cuml.__version__+' installed; umap requires cuml >= 22.06.00')
+        raise import_cuml_exn
```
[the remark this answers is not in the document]
dcolinmorgan: yeah actually both of these are never called

graphistry/umap_utils.py
```diff
@@ -126,17 +133,18 @@ def reuse_umap(g: Plottable, memoize: bool, metadata: Any):  # noqa: C901
...
 
 
-def umap_graph_to_weighted_edges(umap_graph, engine, cfg=config):
+def umap_graph_to_weighted_edges(umap_graph, engine, k, cfg=config):
     logger.debug("Calculating weighted adjacency (edge) DataFrame")
     coo = umap_graph.tocoo()
     src, dst, weight_col = cfg.SRC, cfg.DST, cfg.WEIGHT
-    if engine == "cuml":
+
+    if (k>1) or (engine == "umap_learn"): ## old cuml or umap_learn
```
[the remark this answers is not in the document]
dcolinmorgan: just a way to not need to test version again

graphistry/umap_utils.py
dcolinmorgan: i tried looking it up again -- but `self._umap().__module__` does not work

graphistry/umap_utils.py
```diff
@@ -412,8 +446,10 @@ def umap(
                 default True.
         :return: self, with attributes set with new data
         """
-
-        assert_imported()
+        # if engine=='umap_learn':
+        #     assert_imported()
+        # elif engine=='cuml':
+        #     assert_imported_cuml()
```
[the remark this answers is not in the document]
dcolinmorgan: sure, are these just to give user feedback?

graphistry/umap_utils.py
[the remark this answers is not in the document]
dcolinmorgan: i switched to if knn_graph exists

graphistry/umap_utils.py
```diff
@@ -222,14 +233,37 @@ def umap_fit(self, X: pd.DataFrame, y: Union[pd.DataFrame, None] = None):
...
+            knn_graph = cupy.sparse.csr_matrix((distances, indices, indptr), shape=(X.shape[0], X.shape[0])).get()
+            df = pd.DataFrame.sparse.from_spmatrix(knn_graph)
+            J=df.sparse.to_dense() 
+            self._umap.fit(X=cupy.asnumpy(X),y=y,knn_graph=J.values)
+            self._weighted_edges_df = (
+                umap_graph_to_weighted_edges(knn_graph,self.engine,k)
+            )
+            df = pd.DataFrame.sparse.from_spmatrix(knn_graph)
+            J=df.sparse.to_dense() 
+            self._weighted_adjacency = J #self._umap.graph_
```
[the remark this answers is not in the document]
dcolinmorgan: yeah i thought it was needed for adjacency, but can use coo so removed these

graphistry/umap_utils.py
dcolinmorgan: since `self._umap = umap_engine.UMAP(**umap_kwargs)`
`self._umap().__module__` returns
TypeError: 'UMAP' object is not callable

dcolinmorgan: agreed trying to hide 22.04 in suffix brings more trouble than its worth with type check errors

lmeyerov: `suffix` is reserved for other stuff, so would just entirely remove your use of `suffix`, and revert back to the behavior before this

graphistry/umap_utils.py
```diff
@@ -222,14 +228,33 @@ def umap_fit(self, X: pd.DataFrame, y: Union[pd.DataFrame, None] = None):
         y = self._check_target_is_one_dimensional(y)
         logger.info('-' * 90)
         logger.info(f"Starting UMAP-ing data of shape {X.shape}")
+        # mod_ver=eval(self._umap().__module__.split('.')[0]).__version__
+        if (self.engine == 'cuml') and (self.suffix < 22.06):
```
lmeyerov, line 232: you can make a helper like

```
def is_old_cuml():
   import cuml
   v = cuml.__version__
   if ...
 ```

then this becomes `if (self.engine == 'cuml') and is_old_cuml(): ...`

File: graphistry/umap_utils.py [hunk 8]
```diff
@@ -197,8 +200,11 @@ def umap_lazy_init(
             self._umap = umap_engine.UMAP(**umap_kwargs)
             self.umap_initialized = True
             self.engine = engine_resolved
-            self.suffix = suffix
-
+            if (engine_resolved == 'cuml'):
+                self.suffix = float(umap_engine.__version__[:5])
```
lmeyerov, line 204: this seems unsafe... maybe:

```python
def is_old_cuml():
    import cuml
    vs = cuml.__version__.split(".")
    if vs[0] in ['0', '21']  or (vs[0] == '22' and float(vs[1]) < ???':
     return True
    return False
````

File: graphistry/umap_utils.py [hunk 9]
```diff
@@ -359,7 +384,7 @@ def umap(
         negative_sample_rate: int = 5,
         n_components: int = 2,
         metric: str = "euclidean",
-        suffix: str = "",
+        suffix = "",
```
lmeyerov, line 387: guessing you can backout the `suffix` type changes and it should work again

lmeyerov (commented):
see inline -- i think back out the use of `suffix` in favor of `is_old_cuml()`, ensure passes tests, and that's it!!!

graphistry/umap_utils.py
```diff
@@ -222,14 +223,41 @@ def umap_fit(self, X: pd.DataFrame, y: Union[pd.DataFrame, None] = None):
...
+                indices = indices.reshape(X.shape[0] * self.n_neighbors)
+                indptr = cupy.arange(0, (self.n_neighbors * X.shape[0]) + 1, self.n_neighbors)
+                knn_graph = cupy.sparse.csr_matrix((distances, indices, indptr), shape=(X.shape[0], X.shape[0])).get()
+                self._umap.fit(X=cupy.asnumpy(X),y=y,knn_graph=knn_graph)
+                self._weighted_edges_df = (
+                    umap_graph_to_weighted_edges(knn_graph,self.engine,knn_graph)
+                )
+                self._weighted_adjacency = knn_graph
+            else:
+                return
```
lmeyerov, line 245: stray early `return`?

graphistry/umap_utils.py
dcolinmorgan: then we import cuml even if its not engine -- I think it has to be nested if -- thats why I used return to break from second if -- so cuml>22.06 runs without knn

graphistry/tests/test_umap_utils.py
```diff
@@ -534,8 +541,8 @@ def test_feature_kwargs_yield_different_values_using_umap_api(self):
         assert g2._node_target.shape[1] == n_topics_target, 'Targets '
 
     @pytest.mark.skipif(
-        not has_dependancy or not has_cuml,
-        reason="requires cuml feature dependencies",
+        not has_dependancy or not has_umap,
+        reason="requires ai+umap feature dependencies",
```
lmeyerov, line 544: Should still be testing for cuml bc cuml class?

lmeyerov:
@dcolinmorgan  cool, is the ci test fail your side or @silkspace (I saw he has a fix pr)?

lmeyerov:
(when ready i can test local for a.old cuml)

graphistry/umap_utils.py
```diff
@@ -126,17 +125,17 @@ def reuse_umap(g: Plottable, memoize: bool, metadata: Any):  # noqa: C901
     )
 
 
-def umap_graph_to_weighted_edges(umap_graph, engine, cfg=config):
+def umap_graph_to_weighted_edges(umap_graph, engine, knn, cfg=config):
     logger.debug("Calculating weighted adjacency (edge) DataFrame")
     coo = umap_graph.tocoo()
     src, dst, weight_col = cfg.SRC, cfg.DST, cfg.WEIGHT
-    if engine == "cuml":
+    if (knn is not None) or (engine == "umap_learn"):
```
lmeyerov, line 132: I'm confused...  If someone passes knn with engine=cuml, this will throw an exn bc coo would need .get(), right? The if condition needs to be just on the engine name+version, right?

graphistry/feature_utils.py
```diff
@@ -2361,8 +2365,8 @@ def featurize(
                 #confidence=confidence,  # deprecated
                 min_words=min_words,
                 model_name=model_name,
-                #similarity=similarity,  # deprecated
-                #categories=categories,  # deprecated
+                similarity=similarity,  # deprecated
+                categories=categories,  # deprecated
```
lmeyerov, line 2369: @silkspace if deprecated why in your fix?

lmeyerov (commented):
One CI Bug likely on `res = self._process_umap(res, X_, y_, kind, memoize` should be `res = res._process...` -- should look at the code at why it uses res vs self

graphistry/feature_utils.py
silkspace: It is not Deprecated, it was my error -- I switched branches on VS code but it doesn't naturally switch branches on terminal where i do the checkin. I was playing with the idea of deprecating (and did a hot fix on Friday night so tests would pass).

graphistry/umap_utils.py
```diff
@@ -214,22 +225,29 @@ def _check_target_is_one_dimensional(self, y: Union[pd.DataFrame, None]):
...
         y = self._check_target_is_one_dimensional(y)
         logger.info('-' * 90)
         logger.info(f"Starting UMAP-ing data of shape {X.shape}")
 
-        self._umap.fit(X, y)
+        is_old = is_old_cuml()
+        if (self.engine == 'cuml' and is_old):
+            import cuml
+            knn = cuml.neighbors.NearestNeighbors(n_neighbors=self.n_neighbors)
+            cc = cuml.UMAP().fit(X, y, knn_graph=knn)
```
lmeyerov, line 239: Use self._umap as its UMAP is already init with proper params, and ensures self._umap is fit when leaving this call

Also, neat on passing in knn_graph here -- maybe worth checking cuml source code on whether they do something special for configuring theirs?

graphistry/umap_utils.py
```diff
@@ -214,22 +225,29 @@ def _check_target_is_one_dimensional(self, y: Union[pd.DataFrame, None]):
...
         if self._umap is None:
             raise ValueError("UMAP is not initialized")
-
         t = time()
         y = self._check_target_is_one_dimensional(y)
         logger.info('-' * 90)
         logger.info(f"Starting UMAP-ing data of shape {X.shape}")
 
-        self._umap.fit(X, y)
+        is_old = is_old_cuml()
```
lmeyerov, line 235: Avoid the check in the typical case by making it part of the if statement ...

graphistry/umap_utils.py
```diff
@@ -238,15 +256,18 @@ def umap_fit(self, X: pd.DataFrame, y: Union[pd.DataFrame, None] = None):
 
     def umap_fit_transform(self, X: pd.DataFrame,
                            y: Union[pd.DataFrame, None] = None):
-        self.umap_lazy_init()
         if self._umap is None:
             raise ValueError("UMAP is not initialized")
         self.umap_fit(X, y)
-        emb = self._umap.transform(X)
+
+        if is_old_cuml():
```
lmeyerov, line 263: Why this special case? That means it is using an unconfigured UMAP...

lmeyerov (commented):
Should use self._umap as configured, and ensures subsequent uses are fit, vs a fresh cuml.UMAP()

lmeyerov:
Awesome i'll see about releasing today! cc @silkspace

PR #422, opened 2022-10-23
Features

silkspace:


README.md
```diff
@@ -356,15 +356,33 @@ Automatically and intelligently transform text, numbers, booleans, and other for
...
     new_df = pandas.read_csv(...)
     X_new, _ = g.transform(new_df, None, kind='nodes')
     preds = model.predict(X_new)
     ```
 
+ * Encode model definitions and compare models against each other
+
+   ```python
+    # graphistry
+    from graphistry.features import search_model, topic_model, ngrams_model, ModelDict, default_parameters
```
lmeyerov, line 372: here and docstr, put a link to where to find models, e.g., hf?

README.md
```diff
@@ -356,15 +356,33 @@ Automatically and intelligently transform text, numbers, booleans, and other for
...
+
+    g = graphistry.nodes(df)
+    g = g.umap(X=[..], y=[..], **search_model)
+
+    # set custom encoding model
```
lmeyerov, line 377: swap in a more compelling & real model, like a different language or for translation?

lmeyerov (changes requested):
Code looks good, just docs can be more useful

Maybe there should be an ipynb on a few representative & most likely feature engineering / model tricks? Changing language model, doing translation, ...

README.md
silkspace: I agree it makes sense, but in this demo the `search_model` (MSMARCO2) is the best model.

lmeyerov: yep and maybe also a url for where they can find them

Issue #416, opened 2022-10-13
[FEA] Visual pattern mining


lmeyerov:
**Is your feature request related to a problem? Please describe.**

Imagine a scenario like an org chart or supply chain hierarchy or kubernetes replicated resource graph, where subgraphs / subtrees may be topologically equivalent. Meaning  degrees and maybe types are the same, but something like the timestamp and ID changes across replicas.

It should be easier to understand these patterns:
- what are the canonical motifs?
- for each motif, what are the instances?
- how are they the same/different?
- maybe we're ok with overlap (multilabel), or need them all to be distinct
- results exposed both data-level & visually

**Describe the solution you'd like**

```python
g2  = g1.motifs(
min_nodes=1, max_nodes=10, # component size heuristic
require_node_fields=['type'],  # attributes to include in the match
optional_node_fields=['subtype'],  # attributes to include.. if possible
fuzzy_match=False,  # whether ok to have slight topological/attribute differences
layout='motifs' | 'partition' | 'graph', # whether to keep cross-instance edges or layout as independent motifs (or their labeled instances)
)

print(g2._nodes[['id', 'motif_id', 'instance_id']])
g2.plot()
```

**Describe alternatives you've considered**

- Eventually move to in-tool

- Different options for implementation, like subgraph mining, learned embeddings...

**Additional context**

Came up in a customer scenario

lmeyerov:
cc @silkspace @tanmoyio  for visibility

tanmoyio:
@silkspace @lmeyerov I can quickly create a prototype on it, something like ETNs

lmeyerov:
That'd be great!

If it can become an enrichment in the original nodes, we can then plugin to something like gib layout (I can help on that step)

Alex has access to some supplychain data, you should have winlogs, not sure if another of interest as well

Will defer to @silkspace  for priority over current link prediction bits, and feedback on model design...

PR #424, opened 2022-11-14
Temporal Motif Mining

tanmoyio:
dummy script
issue: https://github.com/graphistry/pygraphistry/issues/416
```python3
import graphistry
import pandas as pd

edges = pd.DataFrame([
    [0, 1, 2],
    [5, 2, 3],
    [10, 2, 4],
    [15, 5, 6],
    [20, 6, 7],
    [25, 6, 8],
    [30, 7, 4]], 
    columns=['time', 'src', 'dst']
)
g = graphistry.edges(edges, 'src', 'dst')
print(g.motif('time', accumulate=10, step=1))
```

### DIFF
--- a/graphistry/motif_utils.py
+++ b/graphistry/motif_utils.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pandas as pd
 
 class MotifMixin:
     def __init__(self):
@@ -16,6 +17,15 @@ class MotifMixin:
         self._temporal = False
         if timestamp is not None:
             self._temporal = True
+        else:
+            self._temporal = False
+
+        # appending a dummy timestamp col
+        if not self._temporal:
+            timestamp = 'time'
+            accumulate = 1
+            step = 1
+            self._edges[timestamp] = 0
 
         partitions = MotifMixin.split(
                     self._edges, 
@@ -25,6 +35,9 @@ class MotifMixin:
 
         # fetching motifs
         motifs = {}
+        if not self._temporal:
+            motif_edge_table = [] 
+
         for i in range(len(partitions)-step+1):
             _nodes = partitions[i][src].tolist() + partitions[i][dst].tolist()
             _nodes = set(_nodes)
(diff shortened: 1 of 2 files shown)


lmeyerov:
Super cool!

I'm wondering:

- What's a good interface? 
     - Typical use: I'm think should probably have something like `motifs_temporal() -> Plottable` so users can `.plot()` right after or continue chaining things
     - .... and have an underlying `motifs_temporal_core() -> DataFrame` closer to current
      - ... where it returns something like a `| node id | motif id` , and i'm unclear on nodes vs edges (`kind="nodes" | "edges" | "both"`, default `"both"` ?) Returning something that matches `g._nodes[[g._node]]` is common to igraph/cugraph/etc: preserves index order, ...
 -  ... what does this actually do (ex: any papers?)  😊 --- cc @silkspace 

Some things to keep in mind:
- Should use mypy on headers and pass lint
- We can add tests as gets more ready
- Stay away from `.iterrows()` and other non-vector ops: for us to run on GPUs, we need written in a vectorized style that we can quickly port to cudf

graphistry/motif_utils.py
```diff
@@ -62,22 +83,22 @@ def motif(
         if self._temporal:
             return motifs
         else:
-            return self.edges(pd.DataFrame(motif_edge_table, columns=[src, dst, "motifID"]))
+            return self.edges(pd.DataFrame(motif_edge_table, columns=[src, dst, motifID]))
 
     @staticmethod
-    def neighbors(graph, v, src, dst):
```
lmeyerov: if this is mixed into the global, should be something like

```
def neighbors(self, v : Union[int, series], mode="ID" | "graph", include_self=False, direction='incoming'|'outgoing'|'both') ->  union[series, plottable]
```

graphistry/motif_utils.py
```diff
@@ -153,14 +174,19 @@ def get_motifs(ETN):
         return '0e'+''.join(e[2:] for e in binary_node_encodings)
 
     @staticmethod
-    def split(x, accumulate, timestamp):
-        times = x[timestamp]
-        pivot = times[0]
+    def _split_edges(edf: pd.DataFrame, accumulate, timestamp):
```
lmeyerov, line 177: what is this?

lmeyerov:
@tanmoyio can we close?

tanmoyio:
yes

PR #425, opened 2022-11-22
rapid demos update 

DataBoyTX:
add limit to read_csv and fix aggregate in notebook ii


lmeyerov (approved):
lgtm

when more time, probably:
* replace blazing demo with dask_cudf & dask-sql-cudf demos
* before hypergraph()]'graph'].plot(), do a regular .plot() -- way simpler

PR #429, opened 2022-12-13
Databricks notebook update 

DataBoyTX:
- remove call to DisplayHTML - no longer required


lmeyerov (approved):
lgtm

fwiw, going forward --

* [x] repo's like these change a CHANGELOG.md so should include change as part of those (ok if put in Development section when not doing as a release, such as in this case)
* [ ] try to use commit messages in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format, like `fix(databricks demo): remove deprecated dispalyHTML` <--- `<commit type>(<object impacted>): transform to the object`

PR #432, opened 2023-01-06
Cleanup edits

dess890:
most are suggestions only caught a few typos!


lmeyerov:
@silkspace  thoughts on landing?

* which branch? this seems to be both code + docs, so not sure what it should merge into wrt code
* text edits ok or drifted?

lmeyerov:
I think these are already in the merge branch

PR #445, opened 2023-02-21
Cudf

dcolinmorgan:
not sure how to check if X_ is cudf.DataFrame without loading cudf at the top


graphistry/feature_utils.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 import os
 import pandas as pd
+import cudf
```
lmeyerov, line 5: no top-level cudf import

graphistry/feature_utils.py
```diff
@@ -167,7 +168,7 @@ def resolve_feature_engine(
 
 def resolve_y(df: Optional[pd.DataFrame], y: YSymbolic) -> pd.DataFrame:
 
-    if isinstance(y, pd.DataFrame):
+    if isinstance(y, pd.DataFrame) or isinstance(y, cudf.DataFrame):
```
lmeyerov, line 171: a trick to avoid an import here may be to use an introspection routine like `'cudf.core.dataframe' in str(getmodule(o))`

graphistry/umap_utils.py
```diff
@@ -3,6 +3,7 @@
 from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
 
 import pandas as pd
+import cudf
```
lmeyerov, line 6: same as above, defer

graphistry/umap_utils.py
```diff
@@ -492,6 +492,12 @@ def umap(
             logger.debug("umap X_: %s", X_)
             logger.debug("umap y_: %s", y_)
 
+            if isinstance(X_,pd.DataFrame):
+                index_to_nodes_dict = dict(zip(range(len(nodes)), nodes))
+            elif isinstance(X_,cudf.DataFrame):
+                index_to_nodes_dict=cudf.DataFrame(nodes).reset_index()
+                X_=pd.DataFrame(X_.to_numpy())
```
lmeyerov, line 499: shouldn't we keep X_ as a GPU df, and enable process_umap etc handle it? when in gpu mode, generally easier to keep everything as gpu obj and only temporarily back out to cpu for calls that can't... a lot less to keep track of

dcolinmorgan: handling X_ properly is last task

graphistry/feature_utils.py
```diff
@@ -1891,7 +1891,7 @@ def _featurize_nodes(
         ndf = res._nodes
         node = res._node
 
-        if remove_node_column:
+        if remove_node_column and 'cudf.core.dataframe' not in str(getmodule(ndf)):
```
lmeyerov, line 1894: I'm not sure this makes sense? Shouldn't the code path still run?

graphistry/umap_utils.py
```diff
@@ -593,11 +595,11 @@ def _bind_xy_from_umap(
             emb = res._node_embedding
         else:
             emb = res._edge_embedding
+        if 'cudf.core.dataframe' not in str(getmodule(emb)): ## cuda cannot support nulls https://github.com/cupy/cupy/issues/5918#issuecomment-946327237
+            df[x_name] = emb.values.T[0]  # if embedding is greater
+            # than two dimensions will only take first two coordinates
+            df[y_name] = emb.values.T[1]
```
lmeyerov, line 601: maybe we need a fillna or something?

also, how are nulls even getting here?

graphistry/umap_utils.py
```diff
@@ -286,8 +286,7 @@ def _bundle_embedding(self, emb, index):
         if emb.shape[1] == 2 and 'cudf.core.dataframe' not in str(getmodule(emb)):
             emb = pd.DataFrame(emb, columns=[config.X, config.Y], index=index)
         elif emb.shape[1] == 2 and 'cudf.core.dataframe' in str(getmodule(emb)):
-            import cudf
-            emb = cudf.DataFrame(emb, columns=[config.X, config.Y], index=index)
+            emb = pd.DataFrame(emb.to_numpy(), columns=[config.X, config.Y], index=index.to_numpy())
```
lmeyerov, line 289: We should stick w GPU vs CPU if we can

I thought we have an embedding in the form of a cudf df with multiple x y z etc cols to begin with, is it just we are being sloppy in how.we reference the cols ?

dcolinmorgan (commented):
umap works with cudf by setting remove_node_column=False, but then get issues with g.plot() from g._nodes.to_arrow(). feeling closer

graphistry/umap_utils.py
```diff
@@ -287,7 +287,7 @@ def _bundle_embedding(self, emb, index):
             emb = pd.DataFrame(emb, columns=[config.X, config.Y], index=index)
         elif emb.shape[1] == 2 and 'cudf.core.dataframe' in str(getmodule(emb)):
             import cudf
-            emb = cudf.DataFrame(emb.to_cupy(), columns=[config.X, config.Y], index=index.to_cupy())
+            emb = cudf.DataFrame(emb.values, columns=[config.X, config.Y], index=index.values)
```
lmeyerov, line 290: what is `emb` -- isn't it a `cudf.DataFrame` already? this might just be a `emb.rename(columns={..})`? I'm not sure of what/why here

dcolinmorgan: emb is float values so it needs to be made into df

dcolinmorgan: very strange... getmodule does say it is a cudf df, but rename does not work on emb since its a cupy
following works but not pretty enough im guessing:
emb.columns=[config.X, config.Y]
emb.index=index

lmeyerov: what do `isinstance(emb, cudf.DataFrame)` and `type(emb)` say?

graphistry/umap_utils.py
```diff
@@ -286,17 +286,15 @@ def _bundle_embedding(self, emb, index):
...
         else:
             columns = [config.X, config.Y] + [
                 f"umap_{k}" for k in range(2, emb.shape[1] - 2)
             ]
             if 'cudf.core.dataframe' not in str(getmodule(emb)):
                 emb = pd.DataFrame(emb, columns=columns, index=index)
             elif 'cudf.core.dataframe' in str(getmodule(emb)):
-                import cudf
-                emb = cudf.DataFrame(emb.values, columns=columns, index=index.values)
+                emb.columns=columns
```
lmeyerov, line 297: why does shape matter? maybe 286-300 can all be `emb.columns=columns` ?

i'm pretty lost on cases here, and feels like there's happy-path coding here that may merit unit tests for 1/2/3/4-dim to ensure we're agreed on intended output col names

silkspace: It takes care of building an arbitrary umap embedding. I can set n_components=10 and still plot, send EMB to torch, etc

graphistry/umap_utils.py
```diff
@@ -491,6 +495,11 @@ def umap(
 
             logger.debug("umap X_: %s", X_)
             logger.debug("umap y_: %s", y_)
+            logger.debug("data is type :: %s", (type(X_)))
+            if isinstance(X_, pd.DataFrame):
+                index_to_nodes_dict = dict(zip(range(len(nodes)), nodes))
+            elif 'cudf.core.dataframe' in str(getmodule(X_)):
+                index_to_nodes_dict = nodes
```
silkspace, line 502: when cudf, is `index_to_nodes_dict = nodes` still a dict? @dcolinmorgan

lmeyerov: how about... always make it a df, and just varies whether cpu vs gpu?

graphistry/umap_utils.py
lmeyerov: yes, if `emb` is a df, I don't know why any of this munging is here, is all this just a df.rename(columns={..})?

lmeyerov: (earlier there was even weirder numpy & .values stuff)

lmeyerov:
@dcolinmorgan can you add unit tests around the various conversions and dimensionality? Should only run on UMAP tests, and for GPU ones, can do only when GPU available

lmeyerov:
Cc @silkspace I know you have bits here too, should he merge into cleanup instead?

silkspace:
Lets merge cleanup into his branch and see if it works. Once okay, we can merge back in to cleanup.

lmeyerov:
> Lets merge cleanup into his branch and see if it works. Once okay, we can merge back in to cleanup.

^^^ @dcolinmorgan can you try the merge?

(Also, looks like you're adding tests, thank you!!)

dcolinmorgan:
> 

oh wow i didnt realize cleanup was a branch. 
and yes im working on tests, but they are naive for now just based on cuml

graphistry/umap_utils.py
dcolinmorgan: yeah its already a dict from cudf

lmeyerov:
@tanmoyio what base should daniel be targeting?

tanmoyio:
> what base should daniel be targeting?

cudf-final
cc @dcolinmorgan

lmeyerov:
@tanmoyio can we close?

dcolinmorgan:
I believe we can close, have merged this into #486 already

PR #446, opened 2023-02-24
cu_cat

dcolinmorgan:
g. flag for cucat, opt-in

### DIFF
--- a/graphistry/feature_utils.py
+++ b/graphistry/feature_utils.py
@@ -925,9 +925,9 @@ def process_dirty_dataframes(
     :return: Encoded data matrix and target (if not None),
             the data encoder, and the label encoder.
     """
-    if feature_engine=='dirty_cat':
+    if feature_engine =='dirty_cat':
         from dirty_cat import SuperVectorizer, GapEncoder, SimilarityEncoder
-    elif feature_engine=='cuCat':
+    elif feature_engine =='cuCat':
         from cuCat import SuperVectorizer, GapEncoder, SimilarityEncoder
     
     from sklearn.preprocessing import FunctionTransformer
--- a/graphistry/feature_utils.py
+++ b/graphistry/feature_utils.py
@@ -143,6 +143,8 @@ def assert_cuml_imported():
                      "`pip install rapids`"  # noqa
         )
         raise import_cuml_exn
+    
+    
 # ############################################################################
 #
 #     Rough calltree
--- a/graphistry/feature_utils.py
+++ b/graphistry/feature_utils.py
@@ -135,7 +135,14 @@ def assert_imported():
         )
         raise import_min_exn
 
-
+def assert_cuml_imported():
+    has_cuml_dependancy_, import_cuml_exn = lazy_import_has_cuml_dependancy()
+    if not has_cuml_dependancy_:
+        logger.error(  # noqa
+                     "cunl not found, trying running"  # noqa
+                     "`pip install rapids`"  # noqa
+        )
+        raise import_cuml_exn
 # ############################################################################
 #
 #     Rough calltree
@@ -891,6 +898,7 @@ def process_dirty_dataframes(
     similarity: Optional[str] = None,  # "ngram",
     categories: Optional[str] = "auto",
     multilabel: bool = False,
+    feature_engine: Optional[str] = "dirty_cat",
 ) -> Tuple[
     pd.DataFrame,
     Optional[pd.DataFrame],
@@ -922,7 +930,6 @@ def process_dirty_dataframes(
     elif feature_engine=='cuCat':
         from cuCat import SuperVectorizer, GapEncoder, SimilarityEncoder
     
-    from dirty_cat import SuperVectorizer, GapEncoder, SimilarityEncoder
     from sklearn.preprocessing import FunctionTransformer
     t = time()
 
@@ -1164,7 +1171,8 @@ def process_nodes_dataframes(
         n_topics_target=n_topics_target,
         similarity=similarity,
         categories=categories,
-        multilabel=multilabel
+        multilabel=multilabel,
+        feature_engine=feature_engine,
     )
 
     if embedding:
--- a/graphistry/feature_utils.py
+++ b/graphistry/feature_utils.py
@@ -140,7 +140,7 @@ def assert_imported():
         )
         raise import_min_exn
 
-def assert_cuml_imported():
+def assert_cuml_cucat():
     has_cuml_dependancy_, import_cuml_exn = lazy_import_has_cu_cat_dependancy()
     if not has_cuml_dependancy_:
         logger.error(  # noqa
@@ -2381,7 +2381,7 @@ class FeatureMixin(MIXIN_BASE):
         if feature_engine == 'dirty_cat':
             assert_imported()
         elif feature_engine == 'cu_cat':
-            assert_cuml_imported()
+            assert_cuml_cucat()
         if inplace:
             res = self
         else:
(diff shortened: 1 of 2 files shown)


graphistry/feature_utils.py
```diff
@@ -47,6 +47,16 @@
         SuperVectorizer = Any
         GapEncoder = Any
         SimilarityEncoder = Any
+    try:
```
lmeyerov, line 50: sanity check: does cuCat do lazy cudf etc. imports?

graphistry/feature_utils.py
```diff
@@ -47,6 +47,16 @@
...
+        from cuCat import (
+            SuperVectorizer,
+            GapEncoder,
+            SimilarityEncoder,
+        )
+    except:
+        SuperVectorizer = Any
+        GapEncoder = Any
+        SimilarityEncoder = Any
```
lmeyerov, line 59: this is a bit wrong, right? depending on if someone wants to use dirty_cat vs cuCat, which is a runtime engine decision, this would statically override at load-time, and not allow on-the-fly switching, right?

graphistry/feature_utils.py
```diff
@@ -111,7 +135,16 @@ def assert_imported():
         )
         raise import_min_exn
 
-
+def assert_cuml_imported():
+    has_cuml_dependancy_, import_cuml_exn = lazy_import_has_cuml_dependancy()
+    if not has_cuml_dependancy_:
+        logger.error(  # noqa
+                     "cunl not found, trying running"  # noqa
```
lmeyerov, line 142: typo

graphistry/feature_utils.py
```diff
@@ -135,29 +168,32 @@ def assert_imported():
 #
 #      _featurize_or_get_edges_dataframe_if_X_is_None
 
-FeatureEngineConcrete = Literal["none", "pandas", "dirty_cat", "torch"]
+FeatureEngineConcrete = Literal["none", "pandas", "dirty_cat", "torch", "cuCat"]
```
lmeyerov, line 171: lowercase? and wondering if we should do `"cu_cat"` for consistency as well?

graphistry/feature_utils.py
```diff
@@ -91,6 +101,20 @@ def lazy_import_has_min_dependancy():
     except ModuleNotFoundError as e:
         return False, e
 
+def lazy_import_has_cuml_dependancy():
```
lmeyerov, line 104: rename `lazy_import_has_cucat_dependency()` ?

graphistry/feature_utils.py
```diff
@@ -91,6 +101,20 @@ def lazy_import_has_min_dependancy():
...
+    import warnings
+    warnings.filterwarnings("ignore")
+    try:
+        import scipy.sparse  # noqa
+        from scipy import __version__ as scipy_version
+        from cuCat import __version__ as cuCat_version
```
lmeyerov, line 110: should it also check for cuml/cudf?

lmeyerov, line 110: should it also check for cuml/cudf?

setup.py
```diff
@@ -38,7 +38,7 @@ def unique_flatten_dict(d):
 }
 
 base_extras_heavy = {
-  'umap-learn': ['umap-learn', 'dirty-cat==0.2.0', 'scikit-learn>=1.0'],
+  'umap-learn': ['umap-learn', 'dirty-cat==0.2.0', 'scikit-learn>=1.0'] + ['cuCat @ git+https://github.com/graphistry/cuCat/@0.01.0'] ,
```
lmeyerov, line 41: remove from here

setup.py
```diff
@@ -38,7 +38,7 @@ def unique_flatten_dict(d):
...
 }
 base_extras_heavy['ai'] = base_extras_heavy['umap-learn'] + ['scipy', 'dgl', 'torch', 'sentence-transformers', 'annoy', 'joblib']
```
lmeyerov, line 43: move to here

setup.py
lmeyerov: maybe we add a new `base_extras_heavy['cu-cat']` ?

graphistry/feature_utils.py
dcolinmorgan: sure, had just thought for future cudf or cuml all the way thru

graphistry/feature_utils.py
dcolinmorgan: cu-cat is all cupy, no cudf import needed. 
i see now that is how we cant to do gpu all the way thru, but it shouldnt be hard to adapt since the speed/scale issue is solved

setup.py
```diff
@@ -42,6 +42,9 @@ def unique_flatten_dict(d):
...
 
+base_extras_heavy['cu-cat'] = + ['cu-cat @ git+https://github.com/graphistry/cuCat/@0.01.0']
```
lmeyerov, line 45: why the `+` ? (is that legal?)

also, let's add `base_extras_heavy['ai'] = ... + base_extras_heavy['cu-cat']` ?

graphistry/feature_utils.py
lmeyerov: `cupy` import is also slow i think, my point is to ensure `cu_cat` import is lazy here as i'm guessing the `cu_cat` package is slow to load

lmeyerov: oh sorry never mind... this is during type checking, that's fine :)

graphistry/feature_utils.py
```diff
@@ -91,6 +101,25 @@ def lazy_import_has_min_dependancy():
     except ModuleNotFoundError as e:
         return False, e
 
+def lazy_import_has_cu_cat_dependancy():
```
lmeyerov, line 104: i think the callers are still calling `lazy_import_has_cuml_dependency()` instead of the new proper name here

graphistry/feature_utils.py
dcolinmorgan: changed here and graphistry/cu-cat branch

graphistry/feature_utils.py
```diff
@@ -111,7 +140,16 @@ def assert_imported():
         )
         raise import_min_exn
 
-
+def assert_cuml_imported():
```
lmeyerov, line 143: `assert_cuml_cucat` ?

The reason Im being a stickler is I think we have code clones across files, so we need to avoid divergence bugs if folks start editing them and get confused bc they are called the same

graphistry/feature_utils.py
```diff
@@ -135,29 +173,32 @@ def assert_imported():
 #
 #      _featurize_or_get_edges_dataframe_if_X_is_None
 
-FeatureEngineConcrete = Literal["none", "pandas", "dirty_cat", "torch"]
+FeatureEngineConcrete = Literal["none", "pandas", "dirty_cat", "torch", "cu_cat"]
```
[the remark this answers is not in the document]
dcolinmorgan: cucat doesnt use torch anywhere, just cupy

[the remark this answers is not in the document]
lmeyerov: See #ai conv - there is an interface design question here, not obvious what to do

graphistry/feature_utils.py
```diff
@@ -135,29 +173,32 @@ def assert_imported():
 #
 #      _featurize_or_get_edges_dataframe_if_X_is_None
 
-FeatureEngineConcrete = Literal["none", "pandas", "dirty_cat", "torch"]
+FeatureEngineConcrete = Literal["none", "pandas", "dirty_cat", "torch", "cu_cat", "cu_cat|torch"]
```
lmeyerov, line 176: it sounds like we're stuck on a design decision here, or do you have a strong view you can explain already?

I see `"cu_cat|torch"` here, but the rest of the code isn't using

PR #451, opened 2023-03-11
Symai React Interface / Streamlit Component

exrhizo:
image

https://www.loom.com/share/6d783a2d2ff64dd1b923a8785e4b2612


lmeyerov:
@exrhizo  there might be something here for main pygraphistry like a g st component, but not sure

PR #471, opened 2023-04-24
Dev/tweak gpu tests

lmeyerov:
i'm surprised this many commits here, maybe there's another branch target?


graphistry/umap_utils.py
```diff
@@ -565,7 +565,10 @@ def umap(
                     res._nodes = res._nodes.to_pandas()
                 if flag_edges_cudf:
                     res._edges = res._edges.to_pandas()
-                res = res.umap(X=self._nodes, y=self._edges, **umap_kwargs)  # type: ignore
+                if X != None or y != None:
```
lmeyerov: `is not None`, right?

tanmoyio: this is outdated

PR #486, opened 2023-05-15
Cucat Featurization base

tanmoyio:


graphistry/feature_utils.py
```diff
@@ -230,18 +292,19 @@ def features_without_target(
...
     :return: DataFrames of model and target
     """
+    _, _, cudf = lazy_import_has_dependancy_cu_cat()
     if y is None:
         return df
     remove_cols = []
     if y is None:
         pass
-    elif isinstance(y, pd.DataFrame):
+    elif isinstance(y, pd.DataFrame) or isinstance(y, cudf.DataFrame):
```
silkspace, line 301: great catch

lmeyerov:
@tanmoyio can we close, or this is still live?

dcolinmorgan:
- for cu_cat itself (DT3 branch), I have worked out the dynamic memory handling for T4 v A100 flexibility.
- Also worked out datetime passthru. However this needs to bypass cudf dataframing in cu_cat AND pygraphistry so that g.plotter infers datetime correctly to provide time series box
-- currently i accomplish this in a hacky way by binding it to embeddings after transforming but before plotting, thus avoiding cudf requirement
- Now I have refactored code to only require gapencoder and tablevectorizer files/functions DT4 branch forked from DT3

lmeyerov:
Awesome - is the plan to start landing, or more first?

And would it make sense to start reviewing any PRs? If a sequence, can you stack them & point out so clear?

dcolinmorgan:
landing would be wonderful -- before end of july is my dream

DT4 is latest cu_cat PR branch which passes many pytests + works as expected in every demo ive done in last few months

lmeyerov:
ok @tanmoyio can you help double check tests, take for a testdrive, and land first in cu_cat and then here?

After, can you help add to main graphistry (https://github.com/graphistry/graphistry/blob/master/compose/dockerfiles/base/05-nvidia.Dockerfile)  ? I think we should keep default-off for now, and should test that it's truly default off -- that existence doesn't (yet) trigger it to be used, only explicit use.

dcolinmorgan:
[test-full-ai test L395](https://github.com/graphistry/pygraphistry/actions/runs/5618276959/job/15223636983?pr=486#step:8:395) seems to be getting hung up by 1 of 3 features being exactly reproduced

- when first discussing with alex -- this is exactly what we realized approximate estimation would liekly return and user must make sure features make sense, just like with dirty_cat
- likely need to test a few so that 2/3 are always reproduced in several estimations rather that 1 case of 3/3 reproduction

graphistry/embed_utils.py
```diff
@@ -21,12 +21,12 @@ def lazy_embed_import_dep():
...
 
-# def check_cudf():
-#     try:
-#         import cudf
-#         return True, cudf
-#     except:
-#         return False, object
+def check_cudf():
+    try:
+        import cudf
```
lmeyerov: 1. Who calls this on import?

2. And can this be a a) cached call that b) checks module path vs an import?

dcolinmorgan: its only `test_embed_utils` that calls `check_cudf`, swapped out for `lazy_cudf_import` from `umap_utils`

lmeyerov: oh, so that shouldn't be the issue, right? `test/*` shouldn't get imported by `import graphistry`..

lmeyerov:
+ @silkspace  wrt https://github.com/graphistry/pygraphistry/pull/486#issuecomment-1651436514 , may understand better?

graphistry/embed_utils.py
dcolinmorgan: sorry, no i wasnt clear, `test_embed_utils` is only **OTHER** place `lazy_cudf_import` was present. It was used in `embed_utils` and imported cudf to check df dtype, but I have swapped it out in place of just checking via getmodule _e.g._ `if 'cudf' in str(getmodule(self._nodes)):` , so I believe the problem is solved -- tuna looks much better
Screenshot 2023-08-01 at 10 44 04

graphistry/feature_utils.py
```diff
@@ -848,6 +919,7 @@ def process_dirty_dataframes(
     similarity: Optional[str] = None,  # "ngram",
     categories: Optional[str] = "auto",
     multilabel: bool = False,
+    feature_engine: Optional[str] = "dirty_cat",
```
dcolinmorgan, line 922: easy to swap in cu_cat here eventually as opt-in

graphistry/feature_utils.py
lmeyerov: @dcolinmorgan `or (cudf is not None and isinstance(y, cudf.DataFrame)`

lmeyerov: maybe same problem elsewhere?

graphistry/embed_utils.py
```diff
@@ -301,12 +301,14 @@ def embed(
         """
         # this is temporary, will be fixed in future releases
         try:
-            if isinstance(self._nodes, cudf.DataFrame):
+            # if isinstance(self._nodes, cudf.DataFrame):
```
lmeyerov, line 304: remove commented lines

lmeyerov: probably better as lazy import and then the whoe `cudf is not None and isinstance(self._nodes, cudf)` (make a helper, `lazy_isinstance` ?)

lmeyerov: same for all the below too

graphistry/feature_utils.py
```diff
@@ -619,11 +682,19 @@ def fit_pipeline(
...
-
-    return pd.DataFrame(X, columns=columns, index=index)
+    X_type = str(getmodule(X))
+    if 'cudf' not in X_type:
+        X = transformer.fit_transform(X)
+        if keep_n_decimals:
+            X = np.round(X, decimals=keep_n_decimals)  #  type: ignore  # noqa
+        X = pd.DataFrame(X, columns=columns, index=index)
+    else:
+        X = transformer.fit_transform(X.to_numpy())
```
lmeyerov, line 692: how do we know if the transformer is cpu vs gpu? it seems to always assume cpu here, but if X is cudf and transformer is gpu, can't we keep X on gpu?

lmeyerov: a sometimes-ok soln would be checking transformer for being from cuml or maybe cu_cat, but that seems pretty iffy

graphistry/feature_utils.py
```diff
@@ -114,6 +146,33 @@ def assert_imported():
...
+    if not has_cuml_dependancy_:
+        logger.error(  # noqa
+                     "cuml not found, trying running"  # noqa
+                     "`pip install rapids`"  # noqa
+        )
+        raise import_cuml_exn
+
+
+def make_safe_gpu_dataframes(X, y, engine):
+    has_cudf_dependancy_, _, cudf = lazy_import_has_dependancy_cu_cat()
```
lmeyerov, line 160: Add `assert cudf is not None` ?

graphistry/feature_utils.py
```diff
@@ -230,18 +292,19 @@ def features_without_target(
...
         pass
-    elif isinstance(y, pd.DataFrame):
+    elif isinstance(y, pd.DataFrame) or isinstance(y, cudf.DataFrame):
         yc = y.columns
         xc = df.columns
         for c in yc:
             if c in xc:
                 remove_cols.append(c)
-    elif isinstance(y, pd.Series):
+    elif isinstance(y, pd.Series) or isinstance(y, cudf.Series):
```
lmeyerov, line 307: handle non-cu_cat import returning `None` for `cudf`

graphistry/feature_utils.py
```diff
@@ -619,11 +682,19 @@ def fit_pipeline(
...
+    if 'cudf' not in X_type:
+        X = transformer.fit_transform(X)
+        if keep_n_decimals:
+            X = np.round(X, decimals=keep_n_decimals)  #  type: ignore  # noqa
+        X = pd.DataFrame(X, columns=columns, index=index)
+    else:
+        X = transformer.fit_transform(X.to_numpy())
+        if keep_n_decimals:
+            X = np.round(X, decimals=keep_n_decimals)  #  type: ignore  # noqa
+        _, _, cudf = lazy_import_has_dependancy_cu_cat()
```
lmeyerov, line 695: `assert cudf is not None`

graphistry/feature_utils.py
```diff
@@ -911,12 +991,19 @@ def process_dirty_dataframes(
...
-        )
-        X_enc = X_enc.fillna(0.0)
+        data_encoder.get_feature_names_out = callThrough(features_transformed) 
+        if 'cudf' not in str(getmodule(ndf)):
+            X_enc = pd.DataFrame(
+                X_enc, columns=features_transformed, index=ndf.index
+            )
+            X_enc = X_enc.fillna(0.0)
+        else:
+            _, _, cudf = lazy_import_has_dependancy_cu_cat()
```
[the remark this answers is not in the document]
lmeyerov: (and are there tests for this?)

graphistry/umap_utils.py
```diff
@@ -411,8 +411,15 @@ def _process_umap(
...
+            include=["datetime", "datetimetz"]
+        ).columns.to_list()
+        
+        self.R_ = X_[self.datetime_columns]
+        X_ = X_.drop(columns=self.datetime_columns)
+        
         emb = res._umap_fit_transform(X_, y_, verbose=verbose)
-        res._xy = emb
+        res._xy = emb.join(self.R_)
         return res
```
lmeyerov, line 423: @silkspace  what should happen here? it sounds like umap does not support df cols, so is this the right workaround?

@dcolinmorgan is this the dt issue you're talking about, or there's another?

lmeyerov (changes requested):
see comments, looked like a few risky cases, and i'm wondering why testing didn't reveal (or maybe i'm wrong?)

graphistry/feature_utils.py
dcolinmorgan: wouldnt that be a redundant check? this needs to be flexible for pd or cudf

lmeyerov: If cudf is None, I think the current would throw an exn

graphistry/feature_utils.py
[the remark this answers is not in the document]
dcolinmorgan: in general you mean? this gpu/feat branch was @tanmoyio attempt to handle this dilema i believe, but i can take it over

i think in future if cudf is passed as general engine it should initiate cucat and we remove feature_engine flag; also e.g. if cucat + umap-learn passed -- just to bypass user selection error / bad combinations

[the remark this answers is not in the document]
lmeyerov: Yes Tanmoy is out for a bit unexpectedly, so that would help

And yes we need to handle multiple envs & multiple user prefs in them, so we can do smart defaults + allow overrides . Makes automated testing here important bc so tricky ...

graphistry/feature_utils.py
dcolinmorgan: good practice to assert even after if statement check? good to know, yay learning

lmeyerov: It's useful when you don't want the 'if' but can imagine future changes or misuses accidentally getting the assumption wrong

graphistry/feature_utils.py
dcolinmorgan: wow this nearly gave me a heart attack -- good thoughts i will work with... but this is an artifact, no `.to_numpy` needed

graphistry/embed_utils.py
dcolinmorgan: lazy import of cudf (`check_cudf`) was slowing import of graphistry. in embed_utils we dont make cudf df , just check for them... and for some reason send them all to_pandas (need to ask tanmoy about this
)

graphistry/feature_utils.py
```diff
@@ -228,7 +228,7 @@ def resolve_feature_engine(
 YSymbolic = Optional[Union[List[str], str, pd.DataFrame]]
 
 
-def resolve_y(df: Optional[pd.DataFrame], y: YSymbolic) -> pd.DataFrame:
+def resolve_y(df: Optional[pd.DataFrame], y: YSymbolic, cudf: None) -> pd.DataFrame:
```
lmeyerov, line 231: can we keep `cudf` out of the type sig here?

graphistry/feature_utils.py
```diff
@@ -249,7 +249,7 @@ def resolve_y(df: Optional[pd.DataFrame], y: YSymbolic) -> pd.DataFrame:
 XSymbolic = Optional[Union[List[str], str, pd.DataFrame]]
 
 
-def resolve_X(df: Optional[pd.DataFrame], X: XSymbolic) -> pd.DataFrame:
+def resolve_X(df: Optional[pd.DataFrame], X: XSymbolic, cudf: None) -> pd.DataFrame:
```
lmeyerov, line 252: same

graphistry/feature_utils.py
```diff
@@ -1002,7 +1002,7 @@ def process_dirty_dataframes(
             X_enc = cudf.DataFrame(
                 X_enc, columns=features_transformed, index=ndf.index
             )
-            X_enc = X_enc.fillna(0.0).to_pandas()  # will be removed for future cu_cat release
+            X_enc = X_enc.fillna(0.0)#.to_pandas()  # will be removed for future cu_cat release
```
lmeyerov, line 1005: awesome, @tanmoyio  for visibility

graphistry/feature_utils.py
dcolinmorgan: sure that makes sense, i was trying to minimize calling `lazy_cuda` import

graphistry/feature_utils.py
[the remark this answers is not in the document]
dcolinmorgan: there are currently two engine flags -- so full cpu and full gpu work as expected, and independently `feat_eng=dirty_cat` + `umap_eng=cudf` (cpu then gpu)  and `feat_eng=cu_cat` + `umap_eng=umap_learn` (gpu then cpu) work. 

next:
1. add these engine combination tests
2. engine=cudf to encompass cu_cat+cuml or stop worrying and learn to love user prefs

graphistry/feature_utils.py
dcolinmorgan: ok -- after the if statement seems best here again [like other assert you mentioned](https://github.com/graphistry/pygraphistry/pull/486#discussion_r1283943343)

graphistry/umap_utils.py
dcolinmorgan: i thought it was a cudf/cupy issue for some time, but actually i think tested and it seems dirty_cat datetime 'passthrough' just does not work, so need to fix that-- should resolve issues cody and I been seeing using cucat w/dt

dcolinmorgan: [they have wanted to encode such data](https://skrub-data.org/stable/auto_examples/03_datetime_encoder.html#one-liner-with-the-tablevectorizer), but doesnt make sense for us -- they provide a passthrough [but it only works on numeric ](https://github.com/skrub-data/skrub/blob/main/skrub/_table_vectorizer.py#L847C1-L848C1) -- top priority to sort this out

dcolinmorgan: ok -- **working** on [g@dtpass](https://github.com/graphistry/pygraphistry/tree/dtpass) and [cu_cat@DT5](https://github.com/graphistry/cu-cat/tree/DT5) -- will merge

lmeyerov: Fwiw, do we need to update to track latest dirty car, as it has been awhile?

(Maybe do as a follow-on PR after this lands?)

graphistry/feature_utils.py
```diff
@@ -341,7 +410,20 @@ def set_to_numeric(df: pd.DataFrame, cols: List, fill_value: float = 0.0):
...
+        df[new_col] = pd.to_datetime(df[cols], errors="coerce").fillna(0)
+    else:
+        _, _, cudf = lazy_import_has_dependancy_cuda()
+        assert cudf is not None
+        for col in df.columns:
+            try:
+                df[col] = cudf.to_datetime(
+                    df[col], errors="raise", infer_datetime_format=True
+                )
+                print(df[col])
```
lmeyerov, line 424: remove printf

graphistry/tests/test_feature_utils.py
```diff
@@ -344,7 +351,7 @@ def cases_test_graph(self, g, name, value, kind="nodes", df=ndf_reddit):
         print(f'<{name} test graph: {value}>')
         if kind == "nodes":
             ndf = g._nodes
-            self.cases_check_node_attributes(g)
+            # self.cases_check_node_attributes(g) ## causing some issues with types
```
lmeyerov, line 354: should disable the specific tests, not the general helper?

graphistry/embed_utils.py
```diff
@@ -21,12 +21,14 @@ def lazy_embed_import_dep():
...
-    except:
-        return False, object
+# def lazy_isinstance(self._nodes, cudf):
+
+# def check_cudf():
+#     try:
+#         import cudf
+#         return True, cudf
+#     except:
+#         return False, object
```
lmeyerov, line 31: remove dead code

graphistry/embed_utils.py
```diff
@@ -38,7 +40,7 @@ def check_cudf():
     MIXIN_BASE = object
     torch = Any
 
-has_cudf, cudf = check_cudf()
+# has_cudf, cudf = check_cudf()
```
lmeyerov, line 43: remove

graphistry/embed_utils.py
```diff
@@ -301,12 +303,14 @@ def embed(
         """
         # this is temporary, will be fixed in future releases
         try:
-            if isinstance(self._nodes, cudf.DataFrame):
+            # if isinstance(self._nodes, cudf.DataFrame):
```
lmeyerov, line 306: remove dead code

graphistry/embed_utils.py
```diff
@@ -301,12 +303,14 @@ def embed(
...
+            if 'cudf' in str(getmodule(self._nodes)):
                 self._nodes = self._nodes.to_pandas()
         except:
             pass
         try:
-            if isinstance(self._edges, cudf.DataFrame):
+            # if isinstance(self._edges, cudf.DataFrame):
```
lmeyerov, line 312: remove dead code

graphistry/embed_utils.py
```diff
@@ -448,7 +453,8 @@ def predict_links(
         else:
             # this is temporary, will be removed after gpu feature utils
             try:
-                if isinstance(relation, cudf.DataFrame):
+                # if isinstance(relation, cudf.DataFrame):
```
lmeyerov, line 456: remove dead code

graphistry/feature_utils.py
```diff
@@ -93,6 +100,28 @@ def lazy_import_has_min_dependancy():
     except ModuleNotFoundError as e:
         return False, e
 
+def lazy_import_has_dependancy_cuda():
```
lmeyerov, line 103: `_cuda` is confusingly named and confusingly return typed, evidenced by how its callers inconsistently name the results (has_cudf, has_cuml, ..)

for it to be `lazy_import_has_dependecy_cuda`, i'd expect it to only check for something thin like `cupy`

i think you really mean `lazy_import_has_dependency_cu_cat()` and either it returns just `import cu_cat` or say `cudf, cuml, cu_cat`

graphistry/feature_utils.py
```diff
@@ -114,6 +143,35 @@ def assert_imported():
         raise import_min_exn
 
 
+def assert_cuml_cucat():
+    has_cuml_dependancy_, import_cuml_exn, cudf = lazy_import_has_dependancy_cuda()
```
lmeyerov, line 147: i think this is just `has_dependency, import_exn, cudf = ..`

but see note above, the method signature is confusing, which is making it tricky to review (& maintain) this code

graphistry/feature_utils.py
```diff
@@ -114,6 +143,35 @@ def assert_imported():
...
+    if not has_cuml_dependancy_:
+        logger.error(  # noqa
+                     "cuml not found, trying running"  # noqa
+                     "`pip install rapids`"  # noqa
+        )
+        raise import_cuml_exn
+
+
+def make_safe_gpu_dataframes(X, y, engine):
+    has_cudf_dependancy_, _, cudf = lazy_import_has_dependancy_cuda()
```
lmeyerov, line 157: see above comment

graphistry/feature_utils.py
```diff
@@ -137,29 +195,34 @@ def assert_imported():
...
 
 
 def resolve_feature_engine(
     feature_engine: FeatureEngine,
 ) -> FeatureEngineConcrete:  # noqa
 
-    if feature_engine in ["none", "pandas", "dirty_cat", "torch"]:
+    if feature_engine in ["none", "pandas", DIRTY_CAT, "torch", CUDA_CAT]:
         return feature_engine  # type: ignore
+    elif feature_engine in ["cuda", "gpu"]:
```
lmeyerov, line 208: why both `"cuda"` and `"gpu"`, is one an alias for another? can we have just one?

lmeyerov: also, `FeatureEngine` type defintion on :199 does not allow `"cuda", "gpu"`, so we'd need to add there

for consistency, i'd expect this to be also added to umap: https://github.com/graphistry/pygraphistry/blob/db31c0ae0e3af655c6dd33155fca793e6c3ef167/graphistry/umap_utils.py#L93

graphistry/feature_utils.py
```diff
@@ -880,20 +971,36 @@ def process_dirty_dataframes(
     :return: Encoded data matrix and target (if not None),
             the data encoder, and the label encoder.
     """
-    from dirty_cat import SuperVectorizer, GapEncoder, SimilarityEncoder
-    from sklearn.preprocessing import FunctionTransformer
+
+    if feature_engine == CUDA_CAT:
+        lazy_import_has_dependancy_cuda()
```
lmeyerov, line 976: unnecessary, or meant to be an assert?

graphistry/feature_utils.py
```diff
@@ -1959,7 +2108,8 @@ def _featurize_nodes(
         X_resolved = resolve_X(ndf, X)
         y_resolved = resolve_y(ndf, y)
 
-        feature_engine = resolve_feature_engine(feature_engine)
+        res.feature_engine = feature_engine
```
lmeyerov, line 2111: keep as a local variable, otherwise need to add to the class interface: https://github.com/graphistry/pygraphistry/blob/master/graphistry/Plottable.py

graphistry/feature_utils.py
```diff
@@ -2084,6 +2234,9 @@ def _featurize_edges(
                 **{res._destination: res._edges[res._destination]}
             )
 
+        res.feature_engine = feature_engine
```
lmeyerov, line 2237: same

graphistry/feature_utils.py
```diff
@@ -2495,13 +2649,21 @@ def featurize(
                 default True.
         :return: graphistry instance with new attributes set by the featurization process.
         """
-        assert_imported()
+        try:
+            feature_engine = resolve_feature_engine(feature_engine)
+        except:
+            feature_engine = resolve_feature_engine(engine)
```
lmeyerov, line 2655: i'm confused why this is necessary?

also, to keep mypy smoother, may help to do `resolved_feature_engine : FeatureEngineConcrete = ...` so that the `featurize(feature_engine: T1)` and `resolved_feature_engine : T2` typings do not conflict

graphistry/feature_utils.py
```diff
@@ -2495,13 +2649,21 @@ def featurize(
...
+
+        if feature_engine == 'dirty_cat':
+            assert_imported()
```
lmeyerov, line 2658: this is a confusing method name

setup.py
```diff
@@ -47,6 +47,8 @@ def unique_flatten_dict(d):
 # https://github.com/facebookresearch/faiss/issues/1589 for faiss-cpu 1.6.1, #'setuptools==67.4.0' removed
 base_extras_heavy['ai'] = base_extras_heavy['umap-learn'] + ['scipy', 'dgl', 'torch<2', 'sentence-transformers', 'faiss-cpu', 'joblib']
 
+base_extras_heavy['cu_cat'] = base_extras_heavy['ai'] + ['cu_cat @ git+http://github.com/graphistry/cu-cat.git@v0.04.0']
```
lmeyerov, line 50: `graphistry/cu-cat` is 159 commits behind `dcolinmorgan:master`, does that need to land before we can install cu_cat in the dockerfiles?

.github/workflows/ci.yml
```diff
@@ -157,6 +157,54 @@ jobs:
         source pygraphistry/bin/activate
         ./bin/test-umap-learn-core.sh
 
+
+  test-gpu-umap:  # well cpu until get a github actions gpu node
```
lmeyerov, line 161: this might work now? https://medium.com/@tajinder.singh1985/exploring-github-nvidia-powered-gpu-hosted-runner-32b172a92c7e

lmeyerov: maybe we should add some sort of test to confirm the gpu is enabled + gpu is used, e.g.,

```bash
nvidia-smi || exit 1

python3 -c "import cudf; cudf.DataFrame({'x': [0,1,2]})['x'].sum()"
```

lmeyerov: if that is overreach for this pr, we should comment this out

lmeyerov:
this seems to have drifted a bit from main, see merge conflict

i'm unsure about the cu_cat bits here, but if this pr replaces a bunch of `import cudf` with dynamic `"cudf.dataframe" in str(module(df))"` checks to avoid slow imports, sounds useful & overdue...

Issue #488, opened 2023-05-25
[BUG] missing user/pass for register() results in KeyError: 'org_name' in plot()


DataBoyTX:
**Describe the bug**
missing user/pass is not failing on `register()`, but results in KeyError: 'org_name' on `plot()` call 

**To Reproduce**

```import graphistry,os
import pandas as pd

graphistry.register(api=3,  
                    protocol  = "http",
                    server    = "nginx",
                    username  = os.getenv("NON_EXISTENT_ENVAR"),
                    password  = os.getenv("NON_EXISTENT_ENVAR")
                   )

edf = pd.read_csv('https://raw.githubusercontent.com/graphistry/pygraphistry/master/demos/data/honeypot.csv')

graphistry.edges(edf, 'attackerIP', 'victimIP').plot()
```

**Expected behavior**
expect register to fail or warn 

**Actual behavior**

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[8], line 4
      1 n_nodes=50 * 1000
      2 n_edges=1000
----> 4 create_and_plot_df(n_nodes, n_edges)

Cell In[7], line 31, in create_and_plot_df(n_nodes, n_edges, n_df, e_df)
     28 if e_df is None: 
     29     e_df = create_edge_df(n_nodes, n_edges)
---> 31 plot_url(n_df, e_df)

Cell In[7], line 9, in plot_url(n_df, e_df)
      6 g = graphistry.nodes(n_df, 'ID').edges(e_df,'src', 'dest').get_degrees()
      7 print(f'avg_degree = {round(g._nodes["degree"].sum()/len(g._nodes),2)}')
----> 9 url = SERVER + g.plot(render=False)
     10 time_taken = timeit.default_timer() - starttime
     11 #print(f'Time taken generate viz url: {str(round(time_taken/60,2))} mins')

File /opt/conda/envs/rapids/lib/python3.8/site-packages/graphistry/PlotterBase.py:1401, in PlotterBase.plot(self, graph, nodes, name, description, render, skip_upload, as_files, memoize, extra_html, override_html_style)
   1399 elif api_version == 3:
   1400     logger.debug("3. @PloatterBase plot: PyGraphistry.org_name(): {}".format(PyGraphistry.org_name()))
-> 1401     PyGraphistry.refresh()
   1402     logger.debug("4. @PloatterBase plot: PyGraphistry.org_name(): {}".format(PyGraphistry.org_name()))
[...]

PR #491, opened 2023-06-16
Bugfix/register token without org

vaimdev:
The fix expected to fix issue https://github.com/graphistry/pygraphistry/issues/488
and also other issue when register() with JWT token.

### DIFF
--- a/graphistry/pygraphistry.py
+++ b/graphistry/pygraphistry.py
@@ -355,7 +355,7 @@ class PyGraphistry(object):
 
     @staticmethod
     def refresh(token=None, fail_silent=False):
-        """Use self or provided JWT token to get a fresher one. If self token, internalize upon refresh."""
+        """Use self or provided JWT token to get a fresher one. Always save new token."""
         using_self_token = token is None
         logger.debug("1. @PyGraphistry refresh, org_name: {}".format(PyGraphistry.org_name()))
         try:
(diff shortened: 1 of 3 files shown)


graphistry/pygraphistry.py
```diff
@@ -382,6 +377,11 @@ def refresh(token=None, fail_silent=False):
                 PyGraphistry._is_authenticated = True
             return PyGraphistry.api_token()
         except Exception as e:
+            if PyGraphistry.store_token_creds_in_memory():
```
lmeyerov, line 380: This will not actually refresh?

lmeyerov (changes requested):
See comment, fix breaks refresh

graphistry/pygraphistry.py
lmeyerov: Store token creds makes a login() method that can be called, while api-token() returns the stale eve if stale

lmeyerov, line 380: on `if` of 375 if statement, should we add an `else: return token` ?

graphistry/pygraphistry.py
```diff
@@ -382,6 +377,11 @@ def refresh(token=None, fail_silent=False):
...
+                logger.debug("JWT refresh via creds")
+                logger.debug("2. @PyGraphistry refresh :relogin")
+                return PyGraphistry.relogin()
+
```
lmeyerov, line 384: shouldn't it be `if not fail_silent: ... raise e` ?

lmeyerov:
- [ ] @lmeyerov  TODO: For review acceptance, should test against removing https://github.com/graphistry/graphistrygpt/blob/1e745e52c4ebc4eff6b58b22547805d1efff8d53/graphistrygpt/utils.py#L43 in multi-user scenarios

graphistry/pygraphistry.py
vaimdev: fixed

graphistry/pygraphistry.py
lmeyerov: @vaimdev how about the comment ^^^ on missing `else: return token` ? a bug to fix or correct as is?

The implicit `else` case currently ignores `token`, returning instead stale, right?

PR #494, opened 2023-07-19
fix: change python-igraph package name to igraph in setup.py

ntamas:
This is needed because python-igraph was renamed to igraph and the old name is just a legacy redirect to igraph itself that will be deactivated soon.


ntamas:
Related issue in https://github.com/igraph/python-igraph: https://github.com/igraph/python-igraph/issues/699

lmeyerov:
Landed, thanks!

PR #497, opened 2023-08-17
fix(lint)

lmeyerov:


lmeyerov:
Still some strange umap/ai ci fails on 3.9+ where they can't install numba

PR #501, opened 2023-09-02
Feature/gnn

silkspace:
Adds joint training GNN toy model and fits it on Cora dataset. 
For @tanmoyio and @webcoderz as templates for proj work.


lmeyerov:
i'm unclear what problem this solves?

silkspace:
This gives them a base class that doesn't use DGL, is lightweight, and stubs the net while I am away.

PR #508, opened 2023-09-25
fix(saving without an index present)

silkspace:
Originally saving was specific to having an in memory index. 
Change allows saving when and index has not been built.

### DIFF
--- a/graphistry/text_utils.py
+++ b/graphistry/text_utils.py
@@ -272,10 +272,16 @@ class SearchToGraphMixin(MIXIN_BASE):
 
     def save_search_instance(self, savepath):
         from joblib import dump  # type: ignore   # need to make this onnx or similar
-
-        self.build_index()
-        search = self.search_index
-        del self.search_index  # can't pickle Annoy
+        try:
+            self.build_index()
+            search = self.search_index
+            del self.search_index  # can't pickle Annoy
+        except Exception as e:
+            logger.exception(e)
+            logger.warn(
+                "Could not build index, saving without it. Run g.build_index() to build it later"
+            )
+            search = None
         dump(self, savepath)
         self.search_index = search  # add it back
         logger.info(f"Saved: {savepath}")
@@ -285,5 +291,9 @@ class SearchToGraphMixin(MIXIN_BASE):
         from joblib import load  # type: ignore   # need to make this onnx or similar
 
         cls = load(savepath)
-        cls.build_index()
+        try:
+            cls.build_index()
+        except Exception as e:
+            logger.exception(e)
+            logger.warn("Could not build index, run g.build_index() to build it later")
         return cls


graphistry/text_utils.py
```diff
@@ -272,10 +272,16 @@ def search_graph(
...
         from joblib import dump  # type: ignore   # need to make this onnx or similar
-
-        self.build_index()
-        search = self.search_index
-        del self.search_index  # can't pickle Annoy
+        try:
+            self.build_index()
+            search = self.search_index
+            del self.search_index  # can't pickle Annoy
+        except Exception as e:
```
lmeyerov, line 279: is there a specific exn we can zero in on here for the expected case to OK?

graphistry/text_utils.py
```diff
@@ -272,10 +272,16 @@ def search_graph(
...
+            logger.exception(e)
+            logger.warn(
```
lmeyerov, line 281: should this be `import warnings; ...` ? i'm not that versed on python, but to enable structured unit testing, safer sw, etc

graphistry/text_utils.py
```diff
@@ -272,10 +272,16 @@ def search_graph(
...
-
-        self.build_index()
-        search = self.search_index
-        del self.search_index  # can't pickle Annoy
+        try:
+            self.build_index()
+            search = self.search_index
+            del self.search_index  # can't pickle Annoy
+        except Exception as e:
+            logger.exception(e)
```
lmeyerov, line 280: include a message

graphistry/text_utils.py
```diff
@@ -285,5 +291,9 @@ def load_search_instance(self, savepath):
         from joblib import load  # type: ignore   # need to make this onnx or similar
 
         cls = load(savepath)
-        cls.build_index()
+        try:
+            cls.build_index()
+        except Exception as e:
```
lmeyerov, line 296: is there a way for us to know whether to build the index or not vs just doing it? current form is unclear how to maintain and handle

graphistry/text_utils.py
```diff
@@ -285,5 +291,9 @@ def load_search_instance(self, savepath):
...
+            logger.exception(e)
+            logger.warn("Could not build index, run g.build_index() to build it later")
```
lmeyerov, line 298: `import warnings; ...` ?

lmeyerov:
ci failed due to spurious pypi bugs, kicked it off again

graphistry/text_utils.py
```diff
@@ -272,10 +272,16 @@ def search_graph(
...
+        try:
+            self.build_index()
+            search = self.search_index
+            del self.search_index  # can't pickle Annoy
+        except Exception as e:
+            logger.exception(e)
+            logger.warn(
+                "Could not build index, saving without it. Run g.build_index() to build it later"
+            )
+            search = None
```
lmeyerov, line 284: why aren't we raising?

graphistry/text_utils.py
lmeyerov: why aren't we raising?

graphistry/text_utils.py
lmeyerov: Can we detect the expected-to-fail scenario ahead of time, skip building in that scenario, and still raise for unexpected errors? This seems to catch too many exn cases otherwise and thus is known-wrong..

graphistry/text_utils.py
lmeyerov: Can we detect the expected-to-fail scenario ahead of time, skip building in that scenario, and still raise for unexpected errors? This seems to catch too many exn cases otherwise and thus is known-wrong..

lmeyerov (changes requested):
Exn swallowing logic is unsafe, should instead avoid builds in known-bad scenarios, allowing it to still raise in unexpected scenarios

PR #512, opened 2023-10-12
fix (sso): In databricks, the HTML is not displayable, print out the …

vaimdev:
Display another message to show the SSO login URL because databricks is not able to display the HTML login link.


DataBoyTX (commented):
LGTM!

PR #513, opened 2023-10-28
Dev/more graph

lmeyerov:


silkspace:
I've tested it. 
While each works separately, If I run one and then the other, it fails. Ie

`import pandas as pd
import numpy as np

n_edges = 10
source_nodes = np.random.choice(range(1, 11), n_edges)
target_nodes = np.random.choice(range(1, 11), n_edges)
weights = np.random.uniform(1, 10, n_edges)

# Create a DataFrame
edge_df = pd.DataFrame({
    'source': source_nodes,
    'target': target_nodes,
    'weight': weights
})`

then 

`g = graphistry.edges(edge_df, 'source', 'target')
g = g.materialize_nodes()  # have to do this or it fails

g2 = g.compute_igraph('articulation_points')   # works
g2.plot()  # works

g3 = g.compute_igraph('community_optimal_modularity')  #works
g3.plot()

#fails:
g3 = g2.compute_igraph('community_optimal_modularity')
`

This seems to be due to igraph counting vertices like so:

`igraph/io/objects.py:494, in _construct_graph_from_dataframe(cls, edges, directed, vertices, use_vids)
    492 else:
    493     if not edges.iloc[:, :2].isin(vertices.index).all(axis=None):
--> 494         raise ValueError(
    495             "Some vertices in the edge DataFrame are missing from vertices DataFrame"
    496         )
    497     nv = vertices.shape[0]
    498     g = cls(n=nv, directed=directed)

ValueError: Some vertices in the edge DataFrame are missing from vertices DataFrame`

While the nodes dataframes have been enriched with the new columns from the previous step.

lmeyerov:
Hmm is that specific to the new methods here, or a separate issue impacting previous ones too? (Trying to understand if we have to block on landing)

silkspace:
```
g4 = g3.compute_igraph('pagerank')   #fails with same message as above
```

lmeyerov:
Tx, I'll play a bit on the train, tho sounds like not a blocker for this pr as an issue even w/out it

lmeyerov:
ok it looks like something specific to a pathology in your test set, I added several chaining tests for built-in datasets and they passed... will dig further...

silkspace:
Interesting, I did this and now I don't get anything uncommon:

```
from graphistry.plugins.igraph import compute_algs

g = graphistry.edges(df_edges, 'source', 'destination').materialize_nodes()

success ={}
failure ={}
for alg in compute_algs:
    try:
        g = g.compute_igraph(alg)
        success[alg] = 1
    except Exception as e:
        failure[alg] = e
```

The 'failures' are actually okay. 
```
failures = {'community_leiden': igraph._igraph.InternalError('Error at src/community/leiden.c:989: Leiden algorithm is only implemented for undirected graphs. -- Invalid value'),
 'community_multilevel': ValueError('input graph must be undirected'),
 'community_spinglass': igraph._igraph.InternalError('Error at src/community/spinglass/clustertool.cpp:287: Cannot work with unconnected graph. -- Invalid value'),
 'gomory_hu_tree': igraph._igraph.InternalError('Error at src/flow/flow.c:2507: Gomory-Hu tree can only be calculated for undirected graphs. -- Invalid value')}

success = {'articulation_points': 1,
 'authority_score': 1,
 'betweenness': 1,
 'bibcoupling': 1,
 'harmonic_centrality': 1,
 'closeness': 1,
 'clusters': 1,
 'cocitation': 1,
 'community_edge_betweenness': 1,
 'community_fastgreedy': 1,
 'community_infomap': 1,
 'community_label_propagation': 1,
 'community_leading_eigenvector': 1,
 'community_optimal_modularity': 1,
 'community_walktrap': 1,
 'constraint': 1,
 'coreness': 1,
 'hub_score': 1,
 'eccentricity': 1,
 'eigenvector_centrality': 1,
 'k_core': 1,
 'pagerank': 1,
 'spanning_tree': 1}
```

PR #511, opened 2023-10-10
smart dependency manager

dcolinmorgan:
just needed to fix counterfactual of tanmoy's toy code, change order to fit into existing scheme, also add fourth variable for __version__

PR #517, opened 2023-11-28
Dev/depman gpufeat

dcolinmorgan:
to test gpu-featurization (cu_cat) with new dependency manager


graphistry/feature_utils.py
```diff
@@ -81,6 +81,7 @@ def assert_imported_cucat():
         logger.warning(  # noqa
                 "cuml and/or cudf not found, trying running"  # noqa
                 "`pip install rapids`"  # noqa
+                "or `pip install --extra-index-url=https://pypi.nvidia.com cuml-cu11 cudf-cu11`"  #noqa
```
lmeyerov: looks good

however... why is this `assert_...` except never actually asserts, just warns? is this being used incorrectly, or being called incorrectly, or it's just incorrectly named?

lmeyerov (approved):

lmeyerov (commented):
(editing, did not mean to accept)

graphistry/feature_utils.py
dcolinmorgan: 1. for cu_cat I have [removed requirement for cudf/cuml ](https://github.com/graphistry/cu-cat/blob/master/setup.py#L22)so user can install how they want
2. advantage being [cpu fallback of cu_cat can be tested](https://github.com/graphistry/pygraphistry/blob/707b404c366bfb38e4dee5c934ce36fa8f8a316f/.github/workflows/ci.yml#L161)(called `test-gpu-umap` sadly since no gpu runner)
3. so here I reordered so warning if no cuml > loads sklearn, and only if that cant load will it error

- can we abandon #486 ? I've merged that and #511 here already

- fresh colab runs great on many demo datasets (used for cu_cat pytests), apart from strange phenomena on heavy text columns where the `degree` distribution becomes skewed narrower. @tanmoyio loading via `cudf.from_pandas()` seems to partially remedy this despite `make_safe_gpu_dataframes` also applied in `cu_cat._gap_encoder`

.github/workflows/ci.yml
```diff
@@ -157,6 +157,54 @@ jobs:
...
+      run: |
+        source pygraphistry/bin/activate
+        ./bin/test-features.sh
+
+    - name: Core umap tests (weak featurize)
+      run: |
+        source pygraphistry/bin/activate
+        ./bin/test-umap-learn-core.sh
+      
+
```
tanmoyio, line 207: Repetitive(same as `test-core-umap`), lets remove this

lmeyerov:
I think https://github.com/graphistry/pygraphistry/pull/548 solves some of the issues here , maybe rebase and update around it?

graphistry/ai_utils.py
```diff
@@ -422,7 +422,10 @@ def infer_self_graph(res,
         assert (
             emb.shape[0] == df.shape[0]
         ), "minibatches emb and X must have same number of rows since h(df) = emb"
-        df = df.assign(x=emb.x, y=emb.y)  # add x and y to df for graphistry instance
+        try:
+            df = df.assign(x=emb.x, y=emb.y)  # add x and y to df for graphistry instance
```
lmeyerov, line 426: instead of try/catch, can we do value inspection?

graphistry/ai_utils.py
```diff
@@ -447,7 +450,14 @@ def infer_self_graph(res,
...
         diff = X_previously_fit - X_new.iloc[i, :]
-        dist = np.linalg.norm(diff, axis=1)  # Euclidean distance
+        try:
+            diff = np.array(diff, dtype = 'float')
+        except TypeError:
+            pass
+        try:
+            dist = np.linalg.norm(diff, axis=1)  # Euclidean distance
+        except TypeError:
+            dist = np.linalg.norm(diff.to_pandas(), axis=1)  # Euclidean distance
```
lmeyerov, line 460: same, instead of try/catch, can we do value inspection?

graphistry/dep_manager.py
```diff
@@ -0,0 +1,30 @@
+import importlib
+
+class DepManager:
+    def __init__(self):
+        self.pkgs = {}
+
+    def __getattr__(self, pkg:str):
```
lmeyerov, line 7: instead call this `def import(...)` or something to make it explicit that code is running

graphistry/dep_manager.py
```diff
@@ -0,0 +1,30 @@
+import importlib
```
lmeyerov, line 1: `from importlib import __import__, import_module` ?

graphistry/dep_manager.py
```diff
@@ -0,0 +1,30 @@
...
+
+class DepManager:
```
lmeyerov, line 3: move `dep_manager.py` to `utils/dep_manager.py`

graphistry/dep_manager.py
```diff
@@ -0,0 +1,30 @@
...
+        try:
+            pkg_val = importlib.import_module(pkg)
+            self.pkgs[pkg] = pkg_val
+            setattr(self, pkg, pkg_val)
+        except:
+            pass
+
+    def import_from(self, pkg:str, name:str):
+        try:
+            module = __import__(pkg, fromlist=[name])
```
lmeyerov, line 24: importlib docs recommend using `importlib.import_module` instead

graphistry/dep_manager.py
```diff
@@ -0,0 +1,30 @@
+import importlib
+
```
lmeyerov, line 2: this class benefits from a comment on its design and the problem it solves

lmeyerov: ex: maybe it does a fast & cached check of a package being in the path before actually importing, so there should be a testable and observable speedup?

graphistry/dep_manager.py
```diff
@@ -0,0 +1,30 @@
...
+class DepManager:
+    def __init__(self):
+        self.pkgs = {}
+
+    def __getattr__(self, pkg:str):
+        self._add_deps(pkg)
+        try:
+            return self.pkgs[pkg]
+        except KeyError:
+            return None
```
lmeyerov, line 12: instead of hiding the exception from the user, it's probably more informative and unsurprising to:

* cache the exception during `_add_deps`
* fetch the cached exception, instead of None, and rethrow here, instead of return

lmeyerov:
Plan is we'll split this into 2 PRs, one on depman (see comments) and a separate on featurE_uitls fixes

PR #519, opened 2023-12-01
feat(is_in)

lmeyerov:
### Added

* chain/hop: `is_in()` membership predicate, `.chain([ n({'type': is_in(['a', 'b'])}) ])`


lmeyerov:
ci fails fixed in https://github.com/graphistry/pygraphistry/pull/520

PR #520, opened 2023-12-04
Dev/hop predicates

lmeyerov:
Adds:

* hop: optional df queries - `hop(..., source_node_query='...', edge_query='...', destination_node_query='...')`
* chain: optional df queries:
  - `chain([n(query='...')])`
  - `chain([e_forward(..., source_node_query='...', edge_query='...', destination_node_query='...')])`
* `IsIn` now implements an ASTPredicate (starting to prep for more predicate types)


graphistry/compute/ast.py
```diff
@@ -2,7 +2,9 @@
 import pandas as pd
 
 from graphistry.Plottable import Plottable
-from .filter_by_dict import is_in, IsIn
+from .predicates.ASTPredicate import ASTPredicate
```
aucahuasi, line 5: We don't use this import in the file

aucahuasi: nvm is being used in the other predicate PR

PR #527, opened 2023-12-20
databricks: fix register to display SSO login link prior to timer expiration, fixes: #539

DataBoyTX:
…HTML to allow viewing for databricks users


lmeyerov:
@DataBoyTX see ci fails

graphistry/pygraphistry.py
```diff
@@ -270,7 +270,8 @@ def _handle_auth_url(auth_url, sso_timeout, sso_opt_into_type):
         if in_ipython() or in_databricks() or sso_opt_into_type == 'display':  # If run in notebook, just display the HTML
             # from IPython.core.display import HTML
             from IPython.display import display, HTML
-            display(HTML(f'<a href="{auth_url}" target="_blank">Login SSO</a>'))
+            display(HTML(f'<a href="{auth_url}" target="_blank">old: Login SSO</a>'))
+            display(Markdown(f'[new: Login SSO]({auth_url})"))
```
lmeyerov, line 274: missing import? + ci fails

does  this work in both jupyter + databricks?

lmeyerov: also, `old/:new:` seems kind of funny for the text label, i think that's just for testing

graphistry/pygraphistry.py
```diff
@@ -2,6 +2,7 @@
 from typing_extensions import Literal
 from graphistry.Plottable import Plottable
 from graphistry.privacy import Mode, Privacy
+from IPython.display import DisplayObject, TextDisplayObject
```
lmeyerov, line 5: this shouldn't be a top-level import as it'd break many installs; note the dynamic import once we've sniffed

graphistry/pygraphistry.py
```diff
@@ -99,6 +100,19 @@ def strtobool(val: Any) -> bool:
     else:
         raise ValueError("invalid truth value %r" % (val,))
 
+# used to print Login link databricks register() call 
+class Markdown(TextDisplayObject):
+
+    def __init__(self,TextDisplayObject):
+        import markdown as md
```
lmeyerov, line 107: i'm worried about using `markdown` package as an unmet dependency

graphistry/pygraphistry.py
```diff
@@ -270,7 +284,8 @@ def _handle_auth_url(auth_url, sso_timeout, sso_opt_into_type):
         if in_ipython() or in_databricks() or sso_opt_into_type == 'display':  # If run in notebook, just display the HTML
             # from IPython.core.display import HTML
             from IPython.display import display, HTML
-            display(HTML(f'<a href="{auth_url}" target="_blank">Login SSO</a>'))
+            display(HTML(f'<a href="{auth_url}" target="_blank">old: Login SSO</a>'))
+            display(Markdown(f'[new: Login SSO]({auth_url})'))
```
lmeyerov, line 288: should we split this to something like

```python
if in_ipython() or sso_opt_into_type=='display': ...
elif in_databricks() or sso_opt_into_type=='displayHTML': ...
else ...
```

and then we can use the right render for each?

also, should double check databricks notebooks vs dashboards

lmeyerov (changes requested):
see comments

lmeyerov:
@vaimdev I think this is really 2 different issues:

1. Ensure in databricks + jupyter that the link immediately appears (vs after 30s) . That includes when used in a databricks dashboard view.

2. Respect overrides. User should be able to force a display mode, and sniffers should not override that decision if provided. Currently, databricks/python sniffers override any explicit decision.

lmeyerov:
is this closing in favor of https://github.com/graphistry/pygraphistry/pull/543 ?

DataBoyTX:
> is this closing in favor of #543 ?

yes

CHANGELOG.md
```diff
@@ -78,6 +84,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Breaking 🔥
 
 * GFQL `e()` now aliases `e_undirected` instead of the base class `ASTEdge`
+>>>>>>> 2506b798ec723e906c1c5279f613fe0c37bdbad2
```
lmeyerov, line 87: @DataBoyTX bad merge?

graphistry/pygraphistry.py
```diff
@@ -242,12 +245,63 @@ def sso_login(org_name=None, idp_name=None, sso_timeout=SSO_GET_TOKEN_ELAPSE_SEC
...
+            certificate_validation=PyGraphistry.certificate_validation(),
+        ).sso_login(org_name, idp_name)
+
+        try:
+            if arrow_uploader.token:
+                PyGraphistry.api_token(arrow_uploader.token)
+                PyGraphistry._is_authenticated = True
+                arrow_uploader.token = None
+                return PyGraphistry.api_token()
+        except Exception:  # required to log on
```
lmeyerov, line 299: can we be selective on the Exception for the one we expect to be thrown?

graphistry/pygraphistry.py
```diff
@@ -242,12 +245,63 @@ def sso_login(org_name=None, idp_name=None, sso_timeout=SSO_GET_TOKEN_ELAPSE_SEC
...
             PyGraphistry.sso_state(arrow_uploader.sso_state)
-
             auth_url = arrow_uploader.sso_auth_url
             # print("auth_url : {}".format(auth_url))
             if auth_url and not PyGraphistry.api_token():
                 PyGraphistry._handle_auth_url(auth_url, sso_timeout, sso_opt_into_type)
-                return auth_url
+  
+    @staticmethod
+    def databricks_notebook_sso_login(
```
lmeyerov, line 254: is this for databricks notebooks or databricks dashboards?

Where in dashboards, I think the idea is first cell is a non-blocking print, and second cell is a blocking poll on successful sso

Where in notebooks... ?

graphistry/pygraphistry.py
```diff
@@ -670,6 +727,12 @@ def register(
                     import graphistry
                     graphistry.register(api=3, protocol='http', server='nginx', client_protocol_hostname='https://my.site.com', token='abc')
 
+        **Example: Databricks notebook and dashboard users need to call the following before register to correctly display the HTML link:
```
lmeyerov, line 730: I think in the dashboard case, it needs to be in 2 different cells, one is a non-blocking url print, and the other the blocking register for whichever cell will plot

I'm not sure what's needed, if anything,  for notebooks?

graphistry/pygraphistry.py
```diff
@@ -266,25 +320,26 @@ def _handle_auth_url(auth_url, sso_timeout, sso_opt_into_type):
...
-            print("After, if you get timeout error, run graphistry.sso_get_token() to complete the authentication")
+        if not in_databricks():
+            # elif in_ipython() or sso_opt_into_type == 'display':  # If run in notebook, just display the HTML
+            if in_ipython() or sso_opt_into_type == 'display':  # If run in notebook, just display the HTML
+
+                from IPython.display import display, HTML
+                display(HTML(f'<a href="{auth_url}" target="_blank">Login SSO</a>'))
+                print("Please click the above URL to open browser to login")
+                print(f"If you cannot see the URL, please open browser, browse to this URL: {auth_url}")
+                print("Please close browser tab after SSO login to back to notebook")
```
lmeyerov, line 331: mixing `display()` and `print()` suggests an incomplete case analysis & handling, can we handle more carefully?

graphistry/pygraphistry.py
```diff
@@ -242,12 +245,63 @@ def sso_login(org_name=None, idp_name=None, sso_timeout=SSO_GET_TOKEN_ELAPSE_SEC
...
+        try:
+            if arrow_uploader.token:
+                PyGraphistry.api_token(arrow_uploader.token)
+                PyGraphistry._is_authenticated = True
+                arrow_uploader.token = None
+                return PyGraphistry.api_token()
+        except Exception:  # required to log on
+            PyGraphistry.sso_state(arrow_uploader.sso_state)
+            auth_url = arrow_uploader.sso_auth_url
+            if auth_url and not PyGraphistry.api_token():
```
lmeyerov, line 302: in the notebook case (but not needed in dashboard), we risk having an expired token (e.g., been up for an hour)

normally there's an auto-refresh that kicks in for some of the dispatched methods, so this gets auto-handled at plot

BUT when an explicit public call like this one, it makes sense to show the url either way, or to make this an option?

graphistry/pygraphistry.py
```diff
@@ -266,25 +320,26 @@ def _handle_auth_url(auth_url, sso_timeout, sso_opt_into_type):
...
-            import webbrowser
-            input("Press Enter to open browser ...")
-            # open browser to auth_url
-            webbrowser.open(auth_url)
-        else:
-            print(f"Please open a browser, browse to this URL, and sign in: {auth_url}")
-            print("After, if you get timeout error, run graphistry.sso_get_token() to complete the authentication")
+        if not in_databricks():
+            # elif in_ipython() or sso_opt_into_type == 'display':  # If run in notebook, just display the HTML
+            if in_ipython() or sso_opt_into_type == 'display':  # If run in notebook, just display the HTML
```
lmeyerov, line 325: let's make sure overrides get followed before sniffers, so sniffers don't supercede them

graphistry/pygraphistry.py
lmeyerov: we can make print() vs display() options, and sniffers pick the right one

PR #529, opened 2023-12-21
Dev/gfql serialization

lmeyerov:
Adding serialization of GFQL queries in preparation for exposing as a public FEP endpoint on datasets

```python
pattern = [n(), e(), n()]
o = graphistry.compute.chain.to_json(pattern)
pattern2 = graphistry.compute.chain.from_json(o)
g.chain(pattern2).plot()
```

---

### Added

* GFQL `Chain` AST object
* GFQL query serialization - `Chain`, `ASTObject`, and `ASTPredict` implement `ASTSerializable`
  - Ex:`Chain.from_json(Chain([n(), e(), n()]).to_json())`
* GFQL predicate `is_year_end`

### Docs

* GFQL in readme.md

### Changes

* Refactor `ASTEdge`, `ASTNode` field naming convention to match other `ASTSerializable`s (`e.edge_query` vs `e._edge_query`)

### Breaking 🔥

* GFQL `e()` now aliases `e_undirected` instead of the base class `ASTEdge`

### DIFF
--- a/graphistry/compute/predicates/numeric.py
+++ b/graphistry/compute/predicates/numeric.py
@@ -20,18 +20,6 @@ class GT(NumericASTPredicate):
     def __call__(self, s: pd.Series) -> pd.Series:
         return s > self.val
 
-    def to_json(self, validate=True) -> dict:
-        if validate:
-            self.validate()
-        return {'type': 'GT', 'val': self.val}
-    
-    @classmethod
-    def from_json(cls, d: dict) -> 'GT':
-        assert 'val' in d
-        out = GT(val=d['val'])
-        out.validate()
-        return out
-
 def gt(val: float) -> GT:
     """
     Return whether a given value is greater than a threshold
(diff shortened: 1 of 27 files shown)


graphistry/compute/predicates/numeric.py
```diff
@@ -1,80 +1,162 @@
...
     def __init__(self, val: float) -> None:
         self.val = val
 
     def __call__(self, s: pd.Series) -> pd.Series:
         return s == self.val
 
+    def to_json(self, validate=True) -> dict:
+        if validate:
+            self.validate()
+        return {'type': 'EQ', 'val': self.val}
```
aucahuasi, line 126: There are many calls with this pattern at the end:
```python
return {'type': 'EQ', 'val': self.val}
```
I wonder if we can add some util method in the base class NumericASTPredicate to create that map using self.val internally, so we would have something like this:
```python
return self._to_json('EQ')
```
and the implementation in the base would be:
```python
def _to_json(self, op:string):
  return {'type': op, 'val': self.val}
```

aucahuasi (approved):
LGTM!

graphistry/compute/predicates/numeric.py
lmeyerov: ah good idea

like auto-mixin `self.__class__.__name__`

aucahuasi: Yes it can be a mixing as well.

lmeyerov: thanks, i was able to cut like 300+ loc this way

PR #532, opened 2023-12-25
Dev/cugfql

lmeyerov:
GFQL can now run on GPUs!

On a colab T4 GPU (16GB):
- runs a 5 hop traversal on a 100M edge benchmark (returning 3M nodes) in < 4s
- avg 18.4X speedup across 3 SNAP social networks (88K edges - 30M)
- highest speedup of 43X

Adds support to GFQL hop, chain, + helpers to run on cudf when a cudf.DataFrame and matching engine

TODO:

- [x] local tests
- [x] ci
- [x] benchmarks where GPU is better

---

## Infra

* GPU tester threads through LOG_LEVEL

## Refactor

* GFQL methods use generic dataframe methods instead of hardcoding pandas ones
* thread through engines
* add `AbstractEngine` pattern to `engine.py::Engine` enum

---

* Benchmark spreadsheet: https://docs.google.com/spreadsheets/d/1Mb4WqvavZbQPCCpi2H8VxSRTG2BDo2862aqPyDEcEOs/edit#gid=0
* Benchmark ipynb draft: https://colab.research.google.com/drive/1iuH9YWd3VLSALR-3z1Jt35MXELI3Sjk_?usp=sharing

### DIFF
--- a/graphistry/compute/chain.py
+++ b/graphistry/compute/chain.py
@@ -146,6 +146,8 @@ def chain(self: Plottable, ops: Union[List[ASTObject], Chain], engine: Union[Eng
 
     For direct calls, exposes convenience `List[ASTObject]`. Internal operational should prefer `Chain`.
 
+    Use `engine='cudf'` to force automatic GPU acceleration mode
+
     :param ops: List[ASTObject] Various node and edge matchers
 
     :returns: Plotter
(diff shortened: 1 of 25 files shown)


graphistry/compute/chain.py
```diff
@@ -13,6 +14,12 @@
 ###############################################################################
 
 
+if TYPE_CHECKING:
```
aucahuasi, line 17: Maybe we can encapsulate this into another module, so other parts of the code can just import the type alias from that module; something like this (in `type_aliases.py`):
```python
from typing import Type, Any
import pandas as pd

if TYPE_CHECKING:
    DataFrameT = Type[pd.DataFrame]
else:
    DataFrameT = Any
```
And in the client code we can do something like this:
```python
from type_aliases import DataFrameT
def foor(arg: DataFrameT):
```

lmeyerov: The underlying challenge I'm facing is how to do type checking across multiple DF libs (cudf, dask, ...) when we only have pandas installed... Maybe they have stubs?

aucahuasi: Maybe we can use `typing.Union`:
```python
from typing import Type, Union, Any
import pandas as pd

_DataFrameCudf = None

try:
    import cudf
    _DataFrameCudf = cudf.DataFrame
except ImportError:
    pass

if TYPE_CHECKING:
    DataFrameT = Union[pd.DataFrame]
    if _DataFrameCudf:
        DataFrameT = Union[pd.DataFrame, _DataFrameCudf]
    # ... for dask, etc.
else:
    DataFrameT = Any
```

graphistry/compute/typing.py
dcolinmorgan, line 1: seems in combo with smart `dep_manager` this would simplify a lot

PR #542, opened 2024-01-31
Dev/add dataset validation, including those for encodings

aucahuasi:
This PR validates the edge encodings format, checking that the attribute for the complex encoding must exists as part of the edge schema.


graphistry/arrow_uploader.py
```diff
@@ -380,7 +380,17 @@ def verify(self, token=None) -> bool:
             json={'token': token})
         return out.status_code == requests.codes.ok
 
-    def create_dataset(self, json):  # noqa: F811
+    def validate_encoding(self, json):
+        edge_attribute = json['edge_encodings']['complex']['default']['edgeColorEncoding']['attribute']
```
lmeyerov, line 384: this would exn if the encoding is not defined, right?

graphistry/arrow_uploader.py
```diff
@@ -380,7 +380,17 @@ def verify(self, token=None) -> bool:
...
+        edge_attributes = self.edges.column_names
+        if not edge_attribute in edge_attributes:
+            err = f'Invalid edge encoding: attribute {edge_attribute} does not exists in {str(edge_attributes)}'
+            logger.error(err)
+            raise Exception(err)
+
+    def create_dataset(self, json, validate: bool = True):  # noqa: F811
+        if validate:
+            self.validate_encoding(json)
```
lmeyerov, line 393: I'd think:
-  we move the ~entire FEP validator in here (and keep there too for now)
-  add a pass for the new checks introduced here that check against the node/edge tables

lmeyerov (changes requested):
- CI fails
- see 2 comments, this feels like a bigger PR for what it is

graphistry/arrow_uploader.py
lmeyerov: AFAICT we can move all of them, making this PR about:

- `g.validate()`: Entrypoint that, for now, validate metadata -- bindings, simple encodings, complex encodings, settings, ... 
  - And motivating all this, when `._nodes` / `._edges` is available, against that schema. This should be skippable somehow, so that FEP can switch to using PyGraphistry, and run validate even tho the tables aren't available.
- `g.plot(..., validate=True)`: Running that (default-on) for api=3 uploads

Let's defer dataset validations for another PR, e.g., null ID value checks, as that's a big item to itself

aucahuasi: Awesome, that makes sense.  I'll move the logic from `FEP/validate_encodings.py`; I'll create another file (`validate_encodings.py`) just to keep clean the `arrow_uploader.py` module.

graphistry/arrow_uploader.py
aucahuasi: I think I fixed this by incorporating the server-side logic here, so done, thanks!

graphistry/arrow_uploader.py
aucahuasi: Done!

lmeyerov:
awesome --

* CI seems to fail?
* Are there tests we can port in?
* FEP would need to be able to call w/out passing in column names, can we have an entrypoint that still supports that?

aucahuasi:
> * CI seems to fail?

I've fixed most of the errors; however, it seems there are some that are unrelated to the changes in this PR.

> * Are there tests we can port in?

Good catch, I ported the tests from FEP and added a couple (`test_validate_encodings_with_attributes_bad` and `test_validate_encodings_with_attributes_good`)

> * FEP would need to be able to call w/out passing in column names, can we have an entrypoint that still supports that?

Good point, I've converted the function signatures to support an empty list of attributes, and the validation only applies when the attribute list is non-empty.

graphistry/validate_encodings.py
```diff
@@ -309,7 +309,7 @@ def validate_complex_encoding_badge(kind, mode, name, badge):
 # { 
 # }
 # NOTE: auto-adds graphType, encodingType based on name
-def validate_complex_encoding(kind, mode, name, enc, attributes: list):
+def validate_complex_encoding(kind, mode, name, enc, attributes: list = []):
```
lmeyerov, line 312: @aucahuasi this may be clearer as an `attributes: Optional[List] = None` , as `[]` vs `None` connote different things

lmeyerov (changes requested):
* can we get CI passing? maybe @tanmoyio / @vaimdev have ideas

* can we move `validate_encodings.py` to `validate/validate_encodings.py` ? presumably we'll be adding more

* signature change request

otherwise about there

graphistry/validate_encodings.py
aucahuasi: good catch, done!

lmeyerov (approved):
Approving b/c umap fails aren't related to this, and core tests w/ this pass

Are you good going through the publish process in the dev docs? This is probably work being a major bump b/c it'll start throwing exns earlier in people's invalid data pipelines where before there'd be none

aucahuasi:
> @aucahuasi on that note... CHANGELOG :)

I've updated the changelog, thanks!

lmeyerov:
https://github.com/graphistry/pygraphistry/pull/548 is landing, may help w/ CI fails

aucahuasi:
> #548 is landing, may help w/ CI fails

Awesome, I've just update this PR with those changes.

lmeyerov:
LGTM!

Want to try going through the release flow in the develop.md?

aucahuasi:
> Want to try going through the release flow in the develop.md?

I'll check the release flow, I'll comment here in case I have questions, thanks for the review!

PR #543, opened 2024-02-09
Fix: SSO link display iPython

tanmoyio:


lmeyerov (commented):
@tanmoyio  :

- can you test Jupyter vs colab vs databricks, incl databricks dashboard view? @DataBoyTX may have guidance

- right now , the environment sniffing ovverrides any user overrides. Can we make default-off overrides where the api caller can force a path, and if no explicit mode ovverrides, the inference picks the mode?

DataBoyTX:
@tanmoyio - so close!  Works in notebook view, but for some reason, it's not working with a dashboard, I am unable to add the cell for the login html to the dashboard.   I thought simply adding a line of blank HTML to the output might help, but I tried that live at the end of the video and it did not work..  so need to make sure we can add this cell to a dashboard, so a dashboard user can use as well.   thanks!  

https://www.loom.com/share/048f60d13c02480492a06de097de57b1?sid=e06bedc4-ccce-403b-abb2-e1d01a1cc8e2

lmeyerov:
In a session w/ thomas:

* Databricks notebooks will print() live, and I forgot if display() clears old display() or just displays at end
* Databricks dashboards will show stale display() from a previous session, not show print(), and then only shown new display() at the end
* Databricks dashboards can use ipywidgets starting 11.3
* TODO: @DataBoyTX sees if ipywidgets just call python callbacks, or can also cause same-cell / multi-cell / all-cell reexecution . Ideally, also if dashboard re-runs are new (empty) python sessions or continued.


So:

* Databricks notebooks: "Cosmetic" cleanup can use print()s early and display() at the end . Can be done now.
* Databricks dashboards: TBD
  * Idea 1: Users create a first cell that's a solicitation to login (link), ideally skips if already logged in. Then subsequent cells can block until login succeeds.
  * Idea 2: Assuming ipywidgets enable some form of resumption, we can emit the trigger button, and upon completion, users can click a retry/resume button, which restarts things.
* Refactoring: This is all brittle, so we need to be careful about implementation
  * Explicit overrides for these various modes to opt-in
  * ... Environment sniffing only if no overrides are provided
  * Especially if we're messing with blocking/non-blocking, we may want to consider exposing new entrypoints for these things, and corresponding documentation.

lmeyerov:
After analysis, it seems for databricks dashboards, we need to support a split mode:

* dashboard login cell: a non-blocking sso initiation call emits the link and returns
* dashboard usage cell: upon a subsequent plot (or an explicit resumption call), it blocks (polling loop) for SSO link to be successfully used and IdP url acknowledging that

This is a bit strange and finnicky, but I can imagine other envs also doing it, so maybe:

1. default auto sso mode:
  - if we can sniff out that it's a databricks + sso...
  - link emit: we can make register()  call & immediately return a new `graphistry.login_sso_link()` that doesn't block on successful handling
  - Implicit use calls like from plot() would track that it's in login_sso_link mode and know to block / pool, calling some internal `.sso_login_resume()`
3. overrides:
  - register (and plot?) should have an explicit override to force this irrespective of it being databricks
  -  maybe even an explict different route, like `.sso_login_link()` and `.sso_login_resume()`, not sure if there's a better name

PR #547, opened 2024-02-23
add gfql colab links to readme

dcolinmorgan:


setup.py
```diff
@@ -42,7 +42,7 @@ def unique_flatten_dict(d):
 }
 
 base_extras_heavy = {
-  'umap-learn': ['umap-learn', 'dirty-cat==0.2.0', 'scikit-learn>=1.0'],
+  'umap-learn': ['umap-learn', 'dirty-cat==0.2.0', 'scikit-learn<=1.3.2'],
```
lmeyerov, line 45: maybe we keep `>=1.0` here, and make just ci require `<=1.3.2`? I don't want to mess up people's installs as we're just one lib of many

also, do we know what's going on here?

lmeyerov: ok i moved this guidance to test deps in https://github.com/graphistry/pygraphistry/pull/548, so you can rebase as that lands

lmeyerov:
I'm not sure what this improves, and notebooks should be on org accounts?

Issue #552, opened 2024-03-11
[BUG] Validator errors on valid graphs


lmeyerov:
**Describe the bug**

The new validator assumes graphs have nodes, while many graphs just have the edge list defined, and thus it throws exceptions when running this line: https://github.com/graphistry/pygraphistry/blob/c0dbcac3ef32a8b64cec234b1431dab814ba5d7b/graphistry/arrow_uploader.py#L389

lmeyerov:
cc @aucahuasi for visibility

PR #553, opened 2024-03-11
Dev/fix validate node

lmeyerov:
- Handle case of graphs with no nodes table - invocation assumed it
- Add more types to make the nullables explicit


lmeyerov:
Fixes https://github.com/graphistry/pygraphistry/issues/552

PR #555, opened 2024-03-15
Attempt to fix refresh() for SSO

vaimdev:
Refresh in SSO will not display the auth_url due to condition error.
This PR enhance the existing code to throw Token Expire exception and print the correct print out to tell the user that token is expired and required to relogin via SSO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Development]
 
+## [0.33.7 - 2024-04-06]
+
+* Fix refresh() for SSO
+
 ## [0.33.6 - 2024-04-05]
 
 ### Added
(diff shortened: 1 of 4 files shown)


lmeyerov:
AFAICT this mostly just changes the exn to being structured, and gives a clearer exn

lmeyerov (changes requested):
see comment

vaimdev:
> AFAICT this mostly just changes the exn to being structured, and gives a clearer exn

I remembered the changes in the PR that fix the bug, this part of code, in the if condition, I removed the   `and not Pygraphistry.api_token()` , which will print the login URL in the case of refresh. Previously, refresh won't enter the `if` block because `Pygraphistry.api_token()` contains old api_token value.  

image

lmeyerov (approved):
LGTM

To test:
- Run a nexus with this shortened: https://github.com/graphistry/graphistry/blob/30b64d1ed97b8e6b3b634eeb49039bfd558bfeec/compose/dockerfiles/graphistry/nexus/nexus.Dockerfile#L245C5-L245C25
- Setup SSO, including pygraphistry (`graphistry.register(...)`)
- Without patch:
  - first time of `g.plot()`, shows url to login, works
  - after expiry, `g.plot()` should reshow login link... but doesn't

PR #567, opened 2024-05-23
suppress cudf on cpu warning

dcolinmorgan:


dcolinmorgan:
this will pass the initial import warning, but umap still complains

PR #569, opened 2024-06-12
coerce float if string or tuple

dcolinmorgan:
for rare cases first encountered in avr-59k['Source_IP4_Subnet_16'] dataset, for example

`graphistry.nodes(avr59k_splunk['Source_IP4_Subnet_16'].dropna()).umap()`

would throw type error

[see end of this thread](https://graphistry.slack.com/archives/C033H1S0DA6/p1716308253862969?thread_ts=1716307584.556219&cid=C033H1S0DA6)

### DIFF
--- a/graphistry/feature_utils.py
+++ b/graphistry/feature_utils.py
@@ -402,18 +402,16 @@ def try_coerce_to_numeric(ndf: pd.DataFrame):
             if num_floats > len(nndf[j]) / 2:  # most of column is float
                 try:
                     nndf[j] = [float(value) if not isinstance(value, float) else value for value in nndf[j]]
-                    exploded = False
                     logger.info("Coerced strings to floats")
                 except:
                     nndf[j] = nndf[j].apply(lambda x: str(x).split() if isinstance(x, str) and ' ' in x else x)
                     nndf = nndf.explode(j)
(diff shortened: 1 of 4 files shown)


graphistry/feature_utils.py
```diff
@@ -393,6 +393,28 @@ def convert_money_string_to_float(money: str):
     mask = where_is_currency_column(df, col)
     df[col, mask] = df[col, mask].apply(convert_money_string_to_float)
 
+def try_coerce_to_numeric(ndf: pd.DataFrame):
+    try:
+        nndf = ndf.copy()
```
lmeyerov, line 398: :(

graphistry/feature_utils.py
```diff
@@ -393,6 +393,28 @@ def convert_money_string_to_float(money: str):
...
+        object_columns = nndf.select_dtypes(include=['object']).columns
+        for j in object_columns:
+            num_floats = sum(isinstance(x, float) for x in nndf[j].dropna())
```
lmeyerov, line 401: This is slow, and weird to check for floats in a col marked as object?

lmeyerov (commented):
@silkspace @tanmoyio this seems slow and surprising to put in the critical path for most gpu umap calls?

PR #570, opened 2024-06-26
gfql remote calling on FEP

dcolinmorgan:
as stand alone `serial_gfql` function atm, not `g.gfql` yet

### DIFF
--- a/graphistry/gfql_utils.py
+++ b/graphistry/gfql_utils.py
@@ -1,7 +1,7 @@
 import requests
 import json
 import logging
-from typing import TYPE_CHECKING, List, Any
+from typing import TYPE_CHECKING, List, Any, Optional
 from inspect import getmodule
 
 from .feature_utils import FeatureMixin
@@ -44,7 +44,7 @@ class GFQLMixin(MIXIN_BASE):
 
     def gfql(self,
              operations: List[Any] = {"type": "Edge","filter_dict": {}},
-             dataset_id: str = None):
+             dataset_id: Optional[str] = None):
         response = None
         import re
         import graphistry
--- a/graphistry/gfql_utils.py
+++ b/graphistry/gfql_utils.py
@@ -17,7 +17,7 @@ class GFQLMixin(MIXIN_BASE):
     def __init__(self, *args, **kwargs) -> None:
         pass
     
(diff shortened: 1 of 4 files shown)


graphistry/gfql_utils.py
```diff
@@ -0,0 +1,53 @@
+import requests
+import json
+import logging
+from inspect import getmodule
+
+from graphistry import server, api_token
+
```
lmeyerov, line 7: Please best modules near related modules or as part of them, vs adding to top level, when avoidable

graphistry/gfql_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+    
+    return response_data
+
+
+def serial_gfql(dataset_id,operations={"type": "Edge","filter_dict": {}}):
+    if 'frame' in str(getmodule(dataset_id)):
+        import graphistry as g
+        try:
+            dataset_id = g.edges(dataset_id).materialize_nodes()
```
lmeyerov, line 43: Materialize_nodes needs edge bindings available

Afaict you are trying to allow a DF to be passed in - this can make sense, but afaict, needs to be rethought in both concept and implementation

graphistry/gfql_utils.py
```diff
@@ -0,0 +1,53 @@
...
+        server = server(),
+        auth_token = api_token(),
+    )
+    
+    return response_data
+
+
+def serial_gfql(dataset_id,operations={"type": "Edge","filter_dict": {}}):
+    if 'frame' in str(getmodule(dataset_id)):
+        import graphistry as g
```
lmeyerov, line 41: Normally 'g' refers to a bound Plottable instance, not the graphistry module, please use the regular conventions

lmeyerov (commented):
I strongly recommend starting with writing out some examples of how you would like user code to work as an RFC, similar to how.we did the aggregates and other feature specs. The implementation and type signatures here are surprising to me, and I'm guessing because i don't see how they lead to a fluent interface that meets typical use case workflows I would have predicted for pygraphistry & Louie users. Writeup and share, and we work through that first?

Cc @mj3cheun @aucahuasi @silkspace

lmeyerov:
closing, we should restart after the spec is done

PR #575, opened 2024-07-08
Ring layouts

lmeyerov:
This starts adding 3 ring layouts: time-based, generic continuous, and generic categorical 

Out of scope:
* Later PRs should add modes like ordinal, including with user-defined ring mappings, and IT network ring layouts
* Ability to control the axis ring & text colors, and when text color is default, have the text color auto-invert based on BG luminosity as we do with normal labels. That'd take extra streamgl-viz extensions, so out of scope for now imo

Time is almost there. Continuous is not tested.

This seems to have some glitchy behavior on month/year datasets, but second/minute/hours seem to work. Part of the subtlety here is nodes should get positioned on a linear time->radius mapping, e.g., following seconds, while axis rings may have different sizes when in month/year/+ because each month has a different number of days.

## Demos

```python
# infer everything
g = g.time_ring_layout() 

# control things
g0 = g.ring_continuous_layout('t_min', play_ms=0, num_rings=200, time_unit='D')
g1 = g.ring_continuous_layout('t_min', play_ms=0, num_rings=6, time_unit='M')
g2 = g.ring_continuous_layout('t_min', reverse=True, play_ms=0)
```

Similar for:

* `g.ring_continuous_layout()`
* `g.ring_categorical_layout()`

## TODO

- [x] time
- [x] continuous
- [x] categorical 
- [x] demo notebooks
- [x] update related docs
- [x] CI
- [x] finalize default encodings
- [x] cudf

Screenshot from 2024-07-08 16-45-57


graphistry/layout/ring/time.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+@lru_cache
+def unit_to_timedelta(unit: TimeUnit) -> np.timedelta64:
+    map: Dict[TimeUnit, np.timedelta64] = {
+        's': np.timedelta64(1, 's'),
+        'm': np.timedelta64(60, 's'),
+        'h': np.timedelta64(3600, 's'),
+        'D': np.timedelta64(86400, 's'),
+        'W': np.timedelta64(7 * 86400, 's'),
```
aucahuasi, line 30: Maybe we can put some of these computed numbers as constants:
`WEEK = 604800  # 7 * 86400`

graphistry/layout/ring/time.py
```diff
@@ -0,0 +1,391 @@
...
+    r_end: float,
+    scalar: float,
+    label: Optional[Callable[[np.datetime64, int, np.timedelta64], str]] = None,
+    reverse: bool = False
+) -> List[Dict]:
+    
+    def offset_time_to_r(
+        offset: pd.DateOffset,
+    ) -> float:
+        #return (
```
aucahuasi, line 173: Is this something we need when debugging?

Issue #583, opened 2024-09-13
[BUG] - GFQL


SinsBre:
**Describe the bug**

GFQL not filtering on degrees

**To Reproduce**
Code, including data, than can be run without editing:
```python
g = graphistry.nodes(nodes).bind(node="Vertex")
                   .edges(edges).bind(source='Vertex 1', destination='Vertex 2')
         )

g_gfql = g_gqfl_1.get_degrees().nodes(lambda g: g._nodes.reset_index(drop=True)).edges(lambda g: g._edges.reset_index(drop=True))
          .chain([
              n({'degree': gt(1)}),
              e_undirected(),
              n({'degree': gt(1)})
          ])

```

**Expected behavior**
Should only return nodes with degree gt 1

**Actual behavior**
It is not filtering out based on degree.

**Graphistry GPU server environment**
 
AWS - 2.40.74

**Additional context**

Contact for datasample

lmeyerov:
Some minimal reproductions

### Preamble

```python
import graphistry
import pandas as pd
from graphistry import n, e_forward, e_reverse, e_undirected, is_in, gt
```

### 1.

```python
edf = pd.DataFrame({
    's': ['a1', 'b3', 'b3'],
    'd': ['b3', 'b3', 'c1']
})
g = graphistry.edges(edf, 's', 'd').materialize_nodes().get_degrees()
```

where

```
	id	degree_in	degree_out	degree
0	a1	0	1	1
1	b3	2	2	4
2	c1	1	0	1
```

Then unexpected output:

```python
g2 = (g.get_degrees()
  .chain([
      n({'degree': gt(1)}),
      e_undirected(),
      n({'degree': gt(1)})
  ])
)
```
=>

```python
	id	degree	degree_in	degree_out
0	b3	4	2	2
1	a1	1	0	1
```

We don't expect edge `(a1)->(b3)` as `(a1 {degree: 1})`

```
	s	d
0	a1	b3
1	b3	b3
```

lmeyerov:
@SinsBre published as 0.34.5 -- lmk if still happening

I added a test case on a minimal dataset with the same basic query structure, among others: https://github.com/graphistry/pygraphistry/blob/ecea21df3fff4bd4fadcf7f51add1c45ac0df3d7/graphistry/tests/compute/test_chain.py#L383

the bigger dataset may have more interesting shapes going on, so am curious

PR #589, opened 2024-09-21
fix(gfql): missing checks 583

lmeyerov:
Potential fix for https://github.com/graphistry/pygraphistry/issues/583

### Fixes

* We had a disabled filtering check that was allowing too many results to return. Reenabled, and new test case added that triggers it

* Multihop variants of `hop()` and `chain()` ( `e(hops=n)` / `e(to_fixed_point=True)`) was incorrectly handling reverse path logic around: wavefront target, g._nodes

* `hop()` was not handling multihop scenarios around source & destination node matching

Fixes are just these commits:

* https://github.com/graphistry/pygraphistry/pull/589/commits/6e809d6f17cd1a53a94b3add090926988689ad0b

* https://github.com/graphistry/pygraphistry/pull/589/commits/511c9e8fef324e89840b219945c342fdc0c01835

### Infra

* Added more `chain()` tests

* Added `hop()` tests that reflect how `chain()` (vs users) call it

* Added tests around the gfql fixes

@SinsBre can you reinstall, confirm this branch (`graphistry.__version__`), and report whether the bug is still manifesting?

the fix should apply to running in both cpu + gpu modes

@aucahuasi assuming this works, we should prioritize for inclusion in the coming release

### DIFF
--- a/graphistry/compute/hop.py
+++ b/graphistry/compute/hop.py
@@ -243,8 +243,7 @@ def hop(self: Plottable,
                 logger.debug('--- direction in [reverse, undirected] ---')
                 logger.debug('hop_edges_reverse basic:\n%s', hop_edges_reverse)
 
-            #FIXME: What test case does this enable? Disabled to pass shortest path backwards pass steps
-            if False and target_wave_front is not None:
+            if target_wave_front is not None:
                 assert nodes is not None, "target_wave_front indicates nodes"
                 if hops_remaining:
                     intermediate_target_wave_front = concat([
--- a/graphistry/compute/hop.py
+++ b/graphistry/compute/hop.py
@@ -51,8 +51,8 @@ def hop(self: Plottable,
     source_node_query: dataframe query to match nodes before hopping (including intermediate)
     destination_node_query: dataframe query to match nodes after hopping (including intermediate)
     edge_query: dataframe query to match edges before hopping (including intermediate)
-    return_as_wave_front: Only return the nodes/edges reached, ignoring past ones (primarily for internal use)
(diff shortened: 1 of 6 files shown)


lmeyerov:
hmm ci failing some reason, investigating

lmeyerov:
And the fix is more complicated than expected =/ TBD

lmeyerov:
@aucahuasi ok yeah including this pr is a key one for the next release

graphistry/compute/hop.py
```diff
@@ -141,7 +141,11 @@ def hop(self: Plottable,
     matches_edges = edges_indexed[[EDGE_ID]][:0]
 
     #richly-attributed subset for dest matching & return-enriching
-    base_target_nodes = target_wave_front if target_wave_front is not None else g2._nodes
+    if target_wave_front is None:
+        base_target_nodes = g2._nodes
+    else:
+        base_target_nodes = concat([target_wave_front, g2._nodes], ignore_index=True, sort=False).drop_duplicates(subset=[g2._node])
```
mj3cheun, line 147: this confuses me a bit, what is `g2._nodes`? is that all nodes in the graph? or is it something else?

if it is all nodes than this line would be equivalent of `base_target_nodes = g2._nodes` i think...

lmeyerov: `g2._nodes` is the materialized `self`._nodes`:

* on forwards-pass of chain, that's all graph nodes, even if user didn't materialize first
* on reverse-pass, nodes are already materialized, so this is the nodes that were reached on the forwards pass
  *  it does NOT include the starting nodes from the forwards pass, so we include those later

PR #600, opened 2024-10-10
feat(layout): fa2 with circles, incl for gib

lmeyerov:
### Added

* Layout: `circle_layout()` that moves points to a ring based on an ordinal field such as degree
* Layout: `fa2_layout()` that uses the ForceAtlas2 algorithm to lay out the graph, with some extensions.
* Layout: `group_in_a_box_layout()` is significantly faster in CPU and GPU mode
* Layout: `group_in_a_box_layout()` exposes attribute `_partition_offsets`
* Compute methods `g.to_pandas()` and `g.to_cudf()`

### Fix

* cugraph methods now handle numerically-typed node/edge index columns
* repeat calls to `fa2_layout` and `group_in_a_box_layout` now work as expected

### Docs

* Add group-in-a-box layout tutorial

### Infra

* Expose `scene_settings` in `Plottable`


lmeyerov:
Close -- adding tqdm would likely help for gib as slow even w/ GPU

lmeyerov:
Switched demo to blockchain transactions to better show its utility

PR #604, opened 2024-10-12
Dev/readme redux

lmeyerov:
- New readme.md <-- feedback welcome!
  - Only other big change I can imagine is updating the gallery, but rather not do myself in this round

- Defer `squarify` import (used in treemap layout) to allow `pip install graphistry --no-deps` => `import graphistry` to work


lmeyerov:
I want to add a quickref.md as well (on the doc site?) as part of this, b/c that was super handy in the old

lmeyerov:
The original readme had a nice start on who this is for - ds, devs, etc.  Should repeat on both homes.

PR #615, opened 2024-11-28
Dev/gfql endpoint

lmeyerov:
Some very exciting things happening here -- maybe time to call pygraphistry `1.x.x` ??

### Remote dataset binding

Bind to remote (lazy):

```python
g1 = graphistry.bind(dataset_id="abc123")
```

Decouple upload from plot(), and bind the upload:

```python
g_uploaded = g_unuploaded.upload()
print(g_uploaded._dataset_id)
print(g_uploaded._nodes_file_id)
print(g_uploaded._edges_file_id)
print(g_uploaded._url)
```

### Remote GFQL

```python
# + auto-uploads if not already
g2 = g1.chain_remote([....], engine='gpu')

# no need to download results if you just care whether matched vs not
df_meta = g1.chain_remote_shape([...])
```

### Remote Python too!

```python
json_obj = g1.python_remote_json("""
from graphistry import Plottable

def task(g: Plottable):
  return {'status': True}
""")

def identity(g):
  return g
g2 = g1.python_remote_g(identity)

def edges(g):
  return g._edges
e_df = g1.python_remote_table(edges)
```
---

### Changes

* Now tracks `dataset_id`, `nodes_file_id`, `edges_file_id`, and `url`, including clearing them out whenever the nodes/edges gets re-bound with new data

* `plot(render=` expanded from `bool` to `Union[bool, Literal["auto", "g", "ipython", "databricks", "browser"]]` to make more predictable


graphistry/compute/chain_remote.py
```diff
@@ -0,0 +1,297 @@
...
+    elif output_type == "all" and format in ["csv", "parquet"]:
+        zip_buffer = BytesIO(response.content)
+        with zipfile.ZipFile(zip_buffer, "r") as zip_ref:
+            nodes_file = [f for f in zip_ref.namelist() if "nodes" in f][0]
+            edges_file = [f for f in zip_ref.namelist() if "edges" in f][0]
+
+            nodes_data = zip_ref.read(nodes_file)
+            edges_data = zip_ref.read(edges_file)
+
+            nodes_df = read_parquet(BytesIO(nodes_data)) if format == "parquet" else read_csv(BytesIO(nodes_data))
```
mj3cheun, line 123: no action needed now but i imagine when we want to support > 2 binary formats we will need to refactor anywhere format string is being checked inline

graphistry/compute/python_remote.py
```diff
@@ -0,0 +1,103 @@
...
+import requests
+
+from graphistry.Plottable import Plottable
+
+
+def python_remote(
+    self: Plottable,
+    code: str,
+    api_token: Optional[str] = None,
+    dataset_id: Optional[str] = None,
```
mj3cheun, line 14: dataset_id is not optional yet

graphistry/compute/python_remote.py
```diff
@@ -0,0 +1,103 @@
...
+    
+    Uses the latest bound `_dataset_id`, and uploads current dataset if not already bound. Note that rebinding calls of `edges()` and `nodes()` reset the `_dataset_id` binding.
+
+    :param code: Python code that includes a top-level function `def task(g: Plottable) -> Union[str, Dict]`.
+    :type code: str
+
+    :param api_token: Optional JWT token. If not provided, refreshes JWT and uses that.
+    :type api_token: Optional[str]
+
+    :param dataset_id: Optional dataset_id. If not provided, will uplaod current data, store that dataset_id, and run GFQL against that.
```
mj3cheun, line 28: comment references gfql here

mj3cheun: also could clarify that if a dataset_id exists on the plottable it will be reused, as opposed to reuploading every time python_remote is called with it omitted

graphistry/compute/python_remote.py
```diff
@@ -0,0 +1,103 @@
...
+            import graphistry
+            from graphistry import n, e
+            es = pandas.DataFrame({'src': [0,1,2], 'dst': [1,2,0]})
+            g1 = graphistry
+                .edges(es, source='src', destination='dst')
+                .upload()
+            assert g1._dataset_id is not None, "Successfully uploaded"
+            out_json = g1.python_remote(
+                code='''
+from typing import Any, Dict
```
mj3cheun, line 49: it is not necessary to start from 0 indents here, the python endpoint should normalize tabs automatically even if the remote code snipped starts from n > 1 indents.

just a nice thing for readability of snippet in context of other code

graphistry/compute/chain_remote.py
lmeyerov: i think the bigger thing will be getting arrow downloads, i suspect that will give a noticeable speedup over parquet on many connections...

graphistry/compute/python_remote.py
lmeyerov: so this is a bit problematic in that this would trigger re-upload, and not sure if that's 'right':

```python

g2 = g1....

# auto-upload 1
res1 = g2.python_remote(script1)

# auto-upload 2
res2 = g2.python_remote(script2)
```

A few ideas:

* inplace update of `g2` , breaking purity of `v3 = g2.xyz()`
* Some sort of global memoizable trick, so even if `g2._dataset_id` is not mutated on first upload, we can detect `g2` did have a recent upload? Ex: some sort of global weak reference lookup table of last n `g` objects?
* Don't worry about it, and encourage users to do an explicit `g2 = g1.upload()...`

mj3cheun: didnt consider this... this cant happen in python_remote because that never modifies the plottable, but chain_remote definitely does.

my opinion is that perhaps we should avoid self mutation of plottable and instead return a "clone" (with identical metadata etc) with the updated node and edge lists? this should also better mirror the behaviour of non-remote chain anyway?

lmeyerov: Does the python endpoint return a Plottable, JSON, or  any? How would I know ahead of time, or how would the python client sniff?

mj3cheun: for now the python endpoint returns either a string or a json. which one it returns is completely dependent on the return type of the function defined (by the user) in `execute`

lmeyerov: hmm, it might not be so bad to extend the endpoint to reuse the gfql endpoint's machinery here, i'll take a look

lmeyerov: (mimetype is cool, that helps!)

lmeyerov: I agree on avoiding mutation on the base obj, maybe we do some variants like:

```python
# generic helper
_ : Any = g1.remote_python("...", output_mode=...)

# explicit mypy-friendly
g2 = g1.remote_python_g("...")
g2, g1_bound = g1.remote_python_g2("...")
o = g1.remote_python_json("...")
o, g1_bound = g1.remote_python_json2("...")
```

PR #619, opened 2024-12-13
Docs/metadata

lmeyerov:
Some basic metadata, noticed when auditing snyk

Also includes https://github.com/graphistry/pygraphistry/pull/623


aucahuasi (approved):
Thanks for this PR! I wonder if we can mention our community slack as part of the contact info (at least for `FUNDING.md`), other than that this PR looks good.

PR #622, opened 2024-12-21
add support for google spanner graph

DataBoyTX:
Add Google Spanner GQL integration and demo notebooks showing how to connect to google spanner graph database, and visualize the results


graphistry/database_clients/SpannerGraph.py
```diff
@@ -0,0 +1,287 @@
+import os
+import pandas as pd
+import json
+import time
+import logging
+from typing import Any
+from google.cloud.spanner_dbapi.connection import connect
+from google.cloud.spanner_v1.data_types import JsonObject
+from google.cloud.spanner_v1 import KeySet
```
lmeyerov, line 9: this can be an optional dep in setup.py: https://github.com/graphistry/pygraphistry/blob/7f5b2db8f54e3763dbe8a67793c5d707f935626b/setup.py#L48

moving to dynamic imports may help, not sure

graphistry/database_clients/SpannerGraph.py
```diff
@@ -0,0 +1,287 @@
+import os
```
lmeyerov, line 1: graphistry/plugins - https://github.com/graphistry/pygraphistry/tree/master/graphistry/plugins

graphistry/PlotterBase.py
```diff
@@ -38,6 +38,7 @@
 from .arrow_uploader import ArrowUploader
 from .nodexlistry import NodeXLGraphistry
 from .tigeristry import Tigeristry
+from .database_clients.spannergraph import spannergraph
```
lmeyerov, line 41: this top-level import would force all users to import, so we need to be lazy/dynamic somewhere

graphistry/database_clients/SpannerGraph.py
```diff
@@ -0,0 +1,287 @@
...
+        __connect: Establish a connection to the Spanner database.
+        execute_query: Run a GQL query against the database.
+        parse_spanner_json: Convert raw Spanner JSON results into structured data.
+        get_nodes_df: Extract and structure graph nodes into a pandas DataFrame.
+        get_edges_df: Extract and structure graph edges into a pandas DataFrame.
+        gql_to_graph: Combine the above methods to generate a Graphistry visualization.
+        get_schema: Retrieve the schema of the database.
+        validate_data: Validate input data for queries or updates.
+        dump_config: Returns the configuration of the spannergraph instance.
+    """
```
lmeyerov, line 65: see the docs conventions in more modern code in pygraphistry, this is important for sphinx/readthedocs . likewise, for llm use, important to have examples.

graphistry/pygraphistry.py
```diff
@@ -1824,6 +1824,23 @@ def tigergraph(
...
+
+        Args:
+            project_id (str): Google Cloud project ID.
+            instance_id (str): Spanner instance ID.
+            database_id (str): Spanner database ID.
+
+        Returns:
+            PlotterBase: A PlotterBase instance configured with SpannerGraph.
+        """
+        return Plotter().spannergraph(project_id, instance_id, database_id)
```
lmeyerov, line 1841: re:interface --


## Connection

i wonder if we want something similar to the bolt connectors, where we can do streamlined auth + passing in a client


```python
CONN_CFG = { ... }
g1 = graphistry.spanner_connect(**CONN_CFG)
g2 = g1.spanner_query("...")
```

+

```python
native_client = google.spanner.client(...)
g1 = graphistry.spanner_connect(native_client)
g2 = g1.spanner_query("....")
```

## Type queries

Separately, a pattern I'm finding to be type-friendly with remote-mode gfql has been separating return-shape-typed methods:

```python
g2 = g1.spanner_g("get a graph")
df = g1.spanner_df("get a table"")
```

lmeyerov: Also, in the neo4j & gremlin cases, when building real examples:

* hydration: I found scenarios like we'd get either node IDs or edge IDs and would want to then hydrate them against the db
* stored procedures: does GSQL have some notion of passing inputs
* reads vs creates vs writes: any notions here? a good flow would be an example of going through uploading some data, querying it, doing some local enrichment, and then writing back the updates

Part of this makes me want to have some bigger examples to motivate helper functions on top. I suspect `query_g()` and `query_df()` can be good starts, but we need these examples to make more realistic

(The lack of bulk reads/writes such as via arrow is a bit puzzling, but that can wait?)

lmeyerov:
@DataBoyTX looks like ci issues?

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+import time
+from typing import Any, List, Dict
+
+from graphistry.util import setup_logger
+logger = setup_logger(__name__)
+
+import logging
+logging.basicConfig(level=logging.INFO)
+
+from google.cloud.spanner_v1.data_types import JsonObject
```
lmeyerov, line 13: make a dynamic import in the methods using this instead of top-level so when `PlotterBase` imports this file, things don't break in typical installs

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+import pandas as pd
+import json
+import time
+from typing import Any, List, Dict
+
+from graphistry.util import setup_logger
+logger = setup_logger(__name__)
+
+import logging
+logging.basicConfig(level=logging.INFO)
```
lmeyerov, line 11: i assume these will go away before landing (important)

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def __init__(self, data: List[Any], execution_time: float):
+        """
+        Initializes a SpannerQueryResult instance.
+
+        :param data: The raw query results.
+        :param execution_time: The time taken to execute the query.
+        """
+        self.data = data
+        self.execution_time = execution_time
```
lmeyerov, line 36: include unit in name, e.g., `execution_time_s` or `execution_time_ms`

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+        self.record_count = len(data)
```
lmeyerov, line 37: unnecessary?

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def summary(self) -> Dict[str, Any]:
```
lmeyerov, line 39: unnecessary?

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+class SpannerQueryResult:
+    """
+    Encapsulates the results of a query, including metadata.
+
+    :ivar list data: The raw query results.
+    :ivar float execution_time: The time taken to execute the query.
+    :ivar int record_count: The number of records returned.
+    """
+
+    def __init__(self, data: List[Any], execution_time: float):
```
lmeyerov, line 28: I'm guessing we can't get any more granuarly nested typing, like `List[Dict[str,...` or `List[Union[Kind1, Kind2, ...]]` ?

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+            cursor.execute(query)
+            results = cursor.fetchall()
+            execution_time = time.time() - start_time
+            logging.info(f"Query executed in {execution_time:.4f} seconds.")
+            return SpannerQueryResult(results, execution_time)
+        except Exception as e:
+            raise RuntimeError(f"Query execution failed: {e}")
+
+    @staticmethod
+    def convert_spanner_json(data):
```
lmeyerov, line 126: this to_dataframe gets tricky to maintain, having some unit tests here would help

i get the time crunch.. maybe we get at least a sample dump in comments, and follow-up pr with actual tests?

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+        from google.cloud.spanner_v1.data_types import JsonObject
+        json_list = []
```
lmeyerov, line 128: mypy type? ideally structured, like `List[Dict[...` or `List[Union[Kind1, ...`, and method's return type

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+                nodes_df['type'] = nodes_df['label']
+            else:
+                nodes_df['type'] = None  # Assign a default value if 'label' is missing
+        else: 
+            logger.warn("unable to assign 'type' from label, column exists\n")
+        
+        return nodes_df
+
+    @staticmethod
+    def get_edges_df(json_data: list) -> pd.DataFrame:
```
lmeyerov, line 192: include expected output column names in pydoc

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+
+        # if 'type' property exists, skip setting and warn
+        if "type" not in edges_df.columns:
+            # check 'label' column exists before assigning it to 'type'
+            if "label" in edges_df.columns:
+                edges_df['type'] = edges_df['label']
+            else:
+                edges_df['type'] = None  # Assign a default value if 'label' is missing
+        else: 
+            logger.warn("unable to assign 'type' from label, column exists\n")
```
lmeyerov, line 221: generally better as a `warnings.warn`

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+                                json_entry["nodes"].append(node_data)
+                        elif element_dict.get('kind') == 'edge':
+                            labels = element_dict.get('labels', [])
+                            for label in labels:
+                                edge_data = {
+                                    "label": label,
+                                    "identifier": element_dict.get('identifier'),
+                                    "source": element_dict.get('source_node_identifier'),
+                                    "destination": element_dict.get('destination_node_identifier'),
+                                    "properties": element_dict.get('properties')
```
lmeyerov, line 152: probably too much for this PR, but thinking these would be smoother as:

- if a `Plottable` has node/edge bindings defined, use those
- if not, reuse whatever the graph returns
- take in optional parameters at class constructor or method call time to allow overrides

we're slowly getting better at these, so rough patterns that are emerging


maybe a good Help Wanted / Good First Issue issue to file

graphistry/plugins/spannergraph.py
lmeyerov: (important)

graphistry/pygraphistry.py
DataBoyTX: - I added CONF to register  

>hydration: I found scenarios like we'd get either node IDs or edge IDs and would want to then hydrate them against the db

I plan to take the two additional demo datasets that were shared (transit data, and FinVest) and create notebooks for those next, after that, I will work on the hydration from some other datasets we have

>stored procedures: does GSQL have some notion of passing inputs

Yes, I believe so. But need to check syntax https://cloud.google.com/spanner/docs/reference/standard-sql/graph-query-statements

>reads vs creates vs writes: any notions here? a good flow would be an example of going through uploading some data, querying it, doing some local enrichment, and then writing back the updates

agreed

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+            if "label" in edges_df.columns:
+                edges_df['type'] = edges_df['label']
+            else:
+                edges_df['type'] = None  # Assign a default value if 'label' is missing
+        else: 
+            logger.warn("unable to assign 'type' from label, column exists\n")
+
+        return edges_df
+
+    def gql_to_graph(self, query: str) -> Any:
```
lmeyerov, line 225: more consistent with api to be fluent:

```python
def gql_to_graph(self, g: Plottable, query, ...):
   return g.nodes(...).edges(...)
```

ex:

```
g1 = .... encode_point_colors(...)...settings(...)
g2 = g1.gsql_query(...)
```

lmeyerov: this one might be a bit more important as it's a public interface (will keep reading)

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,294 @@
...
+        g = self.graphistry.nodes(nodes_df, 'identifier').edges(edges_df, 'source', 'destination')
+        return g
+
+    # TODO(tcook): add wrapper funcs in PlotterBase for these utility functions: 
+    
+    def get_schema(self) -> Dict[str, List[Dict[str, str]]]:
+        """
+        Retrieves the schema of the Spanner database.
+
+        :return: A dictionary containing table names and column details.
```
lmeyerov, line 248: document expected cols

graphistry/pygraphistry.py
```diff
@@ -620,6 +650,8 @@ def register(
         :type certificate_validation: Optional[bool]
         :param bolt: Neo4j bolt information. Optional driver or named constructor arguments for instantiating a new one.
         :type bolt: Union[dict, Any]
+        :param spanner_config: Spanner connection information. Named constructor arguments for instantiating a spanner client
+        :type spanner_config: Union[dict, Any]
```
lmeyerov, line 654: `spanner_gsql` ?

(there are non-gsql spanner bindings possible afaict)

graphistry/pygraphistry.py
```diff
@@ -1824,6 +1882,12 @@ def tigergraph(
             protocol, server, web_port, api_port, db, user, pwd, verbose
         )
 
+    @staticmethod
+    def spanner_query(query: str, params: Dict[str, Any] = {}) -> Plottable:    
+        # TODO(tcook): add pydocs 
+        return Plotter().spanner_query(query, params)
```
lmeyerov, line 1888: yeah this is where i think we want 2 entrypoints:

* pygraphistry with a static method
* plotterbase with a chained method

in both case, the spanner_query gets a `self: Plottable`

graphistry/pygraphistry.py
```diff
@@ -1824,6 +1882,12 @@ def tigergraph(
             protocol, server, web_port, api_port, db, user, pwd, verbose
         )
 
+    @staticmethod
+    def spanner_query(query: str, params: Dict[str, Any] = {}) -> Plottable:
```
lmeyerov, line 1886: rename to spanner_gsql_query? i think spanner is something else / more general?

lmeyerov (changes requested):
made a quick pass, see comments

- [x] lint, types, e.g., missing type signatures
- [x] public interface should be 'fluent' wrt plottable, so `PlotterBase::spanner_gsql(self: Plottable, query: str...` , using the passed in plottable instead of a fresh `graphistry.edges(...)`
- [x] remove stray top-level `logging.setLevel`
- [x] move remaining spanner top-level import into method bodies
- [x] rename to `spanner_gql` as spanner can be sql too?

Ideally but I get if hard in this pr, we can do a fast-follow next week?
- [ ] where possible, better to do nested types like `List[Dict[...` or `List[Union[Kind1, ...`
- [ ] unit tests on the json2df
- [ ] document df cols in pydocs where relevant

probably good Help Wanted item:
- [ ] infer source/dest name from passed in plottable or remote graph, vs current hard-coding

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,213 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9418e0e1-648d-4cd7-a264-9bf910cbcfe3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
lmeyerov, line 9: first cell should be a description

later cell should be install command (`pip install graphistry[spanner_gsql]` ?)

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,213 @@
...
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.12.3"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 5
```
lmeyerov, line 212: CTA

maybe more gsql, 10 Minutes to PygRaphistry

lmeyerov (changes requested):
see comments wrt ipynb

graphistry/plugins/spannergraph.py
DataBoyTX: I do want to use the logger though, you meant just take out setting to INFO?  Forgot to remove that

graphistry/plugins/spannergraph.py
DataBoyTX: this was an oversight. good catch, thanks. removed

graphistry/pygraphistry.py
DataBoyTX: renamed this one to `spanner_gql_to_g` and added `spanner_query_to_df` that handles both GQL (non-path) queries and SQL

graphistry/pygraphistry.py
DataBoyTX: Found an example for parameterized queries: [graph_snippets.py](https://github.com/googleapis/python-spanner/blob/32e761b0d4052938bf67cfec63a0e83702a35ada/samples/samples/graph_snippets.py#L273) will have to look at adding this in v2

graphistry/pygraphistry.py
DataBoyTX: I think it's coded and working as expected now wrt to this, can you please confirm this is what you were expecting?

DataBoyTX:
> @DataBoyTX looks like ci issues?

@lmeyerov - think everything is passing, can you re-review pls?

graphistry/pygraphistry.py
```diff
@@ -1030,6 +1076,31 @@ def bolt(driver=None):
         """
         return Plotter().bolt(driver)
 
+    @staticmethod
+    def spanner_init(spanner_config=None):
```
lmeyerov, line 1080: input/output types

graphistry/pygraphistry.py
```diff
@@ -1030,6 +1076,31 @@ def bolt(driver=None):
...
+
+                ::
+
+                    import graphistry
+                    spanner_CONF = { project_id: "my_project", instance_id: "my_instance", database_id: "my_database"}
+                    g = graphistry.spanner_init(spanner_CONF)
+
+        """
+        if spanner_config is None: 
+            return None
```
lmeyerov, line 1099: should this thrown an exn if there is no cfg and no `graphistry.register(..., spanner...)`?

afaict what this does is assures there's a working spanner client for subsequent chained `g.xyz()` calls

graphistry/PlotterBase.py
```diff
@@ -2269,7 +2272,41 @@ def bolt(self, driver):
...
+        # check if valid 
+        required_keys = ["project_id", "instance_id", "database_id"]
+        for key in required_keys:
+            value = spanner_config.get(key)
+            if not value:  # check for None or empty values
+                raise ValueError(f"Missing or invalid value for required Spanner configuration: '{key}'")
+
+        res._spannergraph = SpannerGraph(res, project_id, instance_id, database_id, credentials_file)
+        logger.debug("Created SpannerGraph object: {res._spannergraph}")
+        return res
```
lmeyerov, line 2309: can we have plugin logic stay defined in the plugin, and we either just do a mixin, here, or a proxied call to the plugin?

lmeyerov: (we'll eventually rewrite the core here, but for now, are at least trying to get the plugins encapsulated in their definition)

lmeyerov: this would also let us make spanner more 'lazy' wrt import costs, so just triggered on first use of a spanner method

lmeyerov: ex: https://github.com/graphistry/pygraphistry/blob/bd483a973a084d6a36d82253ce6229e1f39fd579/graphistry/PlotterBase.py#L1607

```python
    def from_igraph(self,
        ig,
        node_attributes: Optional[List[str]] = None,
        edge_attributes: Optional[List[str]] = None,
        load_nodes = True, load_edges = True,
        merge_if_existing = True
    ):
        return from_igraph_base(
            self,
            ig,
            node_attributes,
            edge_attributes,
            load_nodes, load_edges,
            merge_if_existing
        )
    from_igraph.__doc__ = from_igraph_base.__doc__
```

graphistry/Plottable.py
```diff
@@ -62,7 +62,8 @@ class Plottable(object):
     _complex_encodings : dict
     _bolt_driver : Any
     _tigergraph : Any
-
+    _spannergraph: Any = None
```
lmeyerov, line 65: keep `: Any` here as the definition is in `PlotterBase`

graphistry/PlotterBase.py
```diff
@@ -2458,6 +2495,112 @@ def cypher(self, query: str, params: Dict[str, Any] = {}) -> Plottable:
...
+        if res._spannergraph is None: 
+            spanner_config = PyGraphistry._config["spanner"]
+            if spanner_config is not None: 
+                logger.debug(f"Spanner Config: {spanner_config}")
+            else: 
+                logger.warning('PyGraphistry._config["spanner"] is None')
+        
+            res = res.spanner_init(PyGraphistry._config["spanner"])  # type: ignore[attr-defined]
+
+        return res._spannergraph.gql_to_graph(query)
```
lmeyerov, line 2551: this seems a bit off, b/c  instead of `gql_to_graph` working with the settings of `self: Plottable`, it'll be using the `g` at time of `register()`

i think it should be a call like `res._spannergraph.gql_to_graph(self, query)` or `res._spannergraph.gql_to_graph(res, query)`

graphistry/plugins/spannergraph.py
```diff
@@ -0,0 +1,267 @@
...
+    def __init__(self, g: Plottable, project_id: str, instance_id: str, database_id: str, credentials_file: Optional[str] = None):
+        """
+        Initializes the SpannerGraph instance.
+
+        :param graphistry: The Graphistry parent object.
+        :param project_id: The Google Cloud project ID.
+        :param instance_id: The Spanner instance ID.
+        :param database_id: The Spanner database ID.
+        """
+        self.g = g
```
lmeyerov, line 57: see comment https://github.com/graphistry/pygraphistry/pull/622/files#r1922867829

i'm tempted to:

- get rid of member field `g` here, as `SpannerGraph()` seems to persist across `g1, g2, g3,...`compositions, while this `g` would be stale

- instead thread through `g` through the callers using it

- an alternative to achieve this is to make this class a mixin, so `self: SpannerGraph <: Plottable`, but that's more complicated than we want rn imo

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,663 @@
...
+    "  </tr>\n",
+    "</table>\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "814017ae-af4e-4331-b0b0-ed016d010ade",
+   "metadata": {},
+   "source": [
+    "# Demo Notebook - Graphistry and Google Spanner Graph \n",
```
lmeyerov, line 28: intro is quite long...

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,213 @@
+{
+ "cells": [
```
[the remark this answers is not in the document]
lmeyerov: (unresolving)

afaict still needs to happen, esp `notebooks/plugins.connectors.html` => public url we will advertise

graphistry/pygraphistry.py
lmeyerov: close, see https://github.com/graphistry/pygraphistry/pull/622/files#r1922869570

lmeyerov (changes requested):
@DataBoyTX 

really close!!!! see comments


### Landable with

- [ ] we should link into readthedocs as that's the public url we will advertise: https://github.com/graphistry/pygraphistry/pull/622/files#r1917905495

    -  docs build may run & detect issues with ipynb, but dunno

- [ ] the current impl will lead to "stale" `g: Plottable` , leading to confusing chaining bugs, so slight tweak to how we auto-thread `g`: https://github.com/graphistry/pygraphistry/pull/622/files#r1922869570

- [ ] interface/protocol should not have assignments: https://github.com/graphistry/pygraphistry/pull/622#discussion_r1922865345

---

### For marketing

In this or follow-on PR, non-blocking:

- see other noted nits: thin out pygraphistry/plotterbase calls to just call impls in plugin, cleaner docstrs for users/gpt/etc, ...

- ipynb
  
  - shorten intro
  - include pregenerated graphistry viz in output so folks can click-to-watch
  - include output tables in inpyb, and ensure .head() so not full dump
  - various prose seems WIP/TODO notes, like various sentences starting with lowercase letters or parenthicals
  - cta:
    - `10 minutes to`, not `10-mins` 
    - get rid of copyright, if we do that, should be via sphinx theme?

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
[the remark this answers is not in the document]
DataBoyTX: @lmeyerov - not sure what you mean.  I added #title and markdown for various sections.  What else is needed here to get it into `notebooks/plugins.connectors.html` ?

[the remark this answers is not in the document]
lmeyerov: it's not auto-indexed, you have to explicit add it to that .html file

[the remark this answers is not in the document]
lmeyerov: https://github.com/graphistry/pygraphistry/blame/bd483a973a084d6a36d82253ce6229e1f39fd579/docs/source/notebooks/plugins.connectors.rst#L11

graphistry/plugins/spannergraph.py
DataBoyTX: column names are dynamic from the node/edge properties, but I updated the pydocs to say that

graphistry/pygraphistry.py
DataBoyTX: I added this logic and raise an exception in both:  `PlotterBase.spanner_query_to_df` and PlotterBase.spanner_query_to_g`

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,663 @@
...
+    "\n",
+    "g = graphistry.spanner_gql_to_g(query)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0b544ab6-18b4-4fc9-adb1-3137bec62fb7",
+   "metadata": {},
+   "outputs": [],
```
lmeyerov, line 226: I recommend saving the output of `g.plot()` calls so people can look like in the docs

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,663 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fe4f9812-6105-4d22-999a-dea1e62a7820",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "g2 = graphistry.spanner_gql_to_g(query2)\n",
+    "g2.plot()"
```
lmeyerov, line 339: same, save output

demos/demos_databases_apis/spanner/google_spanner_finance_graph.ipynb
```diff
@@ -0,0 +1,663 @@
...
+   "source": [
+    "#  now run again and retrive all the paths \n",
+    "query2a='''GRAPH FinGraph\n",
+    "MATCH\n",
+    "p = (src_accnt:Account )-[transfers:Transfers]->{3,6}\n",
+    "  (dst_accnt:Account ) where 1=1 \n",
+    "RETURN SAFE_TO_JSON(p) as path\n",
+    "'''\n",
+    "g2a = graphistry.spanner_gql_to_g(query2a)\n",
+    "g2a.plot()"
```
lmeyerov, line 357: same, save output

graphistry/PlotterBase.py
```diff
@@ -35,9 +35,11 @@
     end_node_id_key,
     to_bolt_driver)
 
+
 from .arrow_uploader import ArrowUploader
 from .nodexlistry import NodeXLGraphistry
 from .tigeristry import Tigeristry
+#from .plugins.spannergraph import SpannerGraph
```
lmeyerov, line 42: stray line, remove

lmeyerov (approved):
@DataBoyTX looking good --

* please fix CI errors, then landable
* optional nits, and on this pr or a follow up, would recommend saving `g.plot()` calls so live viz appear in docs

otherwise it's ready

Issue #614, opened 2024-11-20
e_forward() not returning results without IDs named 'src' and 'dest'


DataBoyTX:
**Describe the bug**

Customer reported bug with GFQL's e_forward() not returning the correct results unless the edge df IDs are named 'src' and 'dest' 

**To Reproduce**

```
import graphistry
import pandas as pd
from graphistry import (
    # graph operators
    n, e_undirected, e_forward, e_reverse, e,
    # attribute predicates
    is_in, ge, startswith, contains, match as match_re
)
import pandas as pd

graphistry.register(...)

edges_df = pd.read_csv('blueprint_edges.csv')
nodes_df = pd.read_csv('blueprint_nodes.csv')

edges_df['src'] = edges_df['parent_elid']
edges_df['dest'] = edges_df['elid']

# create two different graphs with the same cols, just named differently: 
g = graphistry.edges(edges_df, 'parent_elid', 'elid').nodes(nodes_df, 'elid') 
g2 = graphistry.edges(edges_df, 'src', 'dest').nodes(nodes_df, 'elid') 

# Test: 
# expected behavior: be able to use e_forward to get edges
# observed behavior: get no edges back unless using e() or e_undirected()
def test_eforward(g):
    guc = g.chain([
        n({"elid":"905e3174aa"}),
        e_forward(),
        n()
    ])

    print('nodes:\n', guc._nodes, '\n' )
    print('edges:\n', guc._edges)


test_eforward(g)
test_eforward(g2)

# notice that the first call with 'g' does not return any edges, where the call with 'g2' does 
```

**contact me for notebook**

**Expected behavior**

be able to use e_forward to get edges

**Actual behavior**

get no edges back unless using e() or e_undirected()

**Graphistry GPU server environment**

Hub v2.41.10 

**PyGraphistry API client environment**
 - Where run `Jupyter Lab local`
 - Version  `0.34.17`
 - Python Version `Python 3.8.5`

lmeyerov:
@DataBoyTX any way to know if they have additional surprisingly  named columns? we don't assume hard-coded names, so the issue is generally excess results, not missing...

DataBoyTX:
@lmeyerov - I am not sure what you mean by surprisingly named columns, the column names that don't work are:  `parent_elid` for src ID and `elid` for dest ID, that returns 0 results, but using src and dest IDs named `src` and `dest` does work.  

Also curious what you mean by we don't assume hard-coded names?

[...]

PR #623, opened 2024-12-24
Dev/gfql fixes

lmeyerov:
### Feat

- Edge gfql ast node deserialize as more precise `ASTEdge` subclasses

### Fix

- Detect node/edge binding name collision breaking hop() and raise `NotImplementedError` (https://github.com/graphistry/pygraphistry/issues/614)
- MyPy enum rules


lmeyerov:
Landing as part of https://github.com/graphistry/pygraphistry/pull/619

Issue #634, opened 2025-01-08
[FEA] Dynamic/Programmatic copyright year at footer of Nexus


vaimdev:
In Nexus, at the footer - the Copyright year, is hard code, change it to use system date's year.

Issue #644, opened 2025-01-25
[BUG] spanner queries give KeyError if spanner_config not set via register()


DataBoyTX:
**Describe the bug**

if spanner_config not set, `spanner_gql_to_g` and likely `spanner_query_to_df` errors out.  Need to handle better with error message on how to resolve vs python KeyError 

**To Reproduce**

attempt to run a query without passing spanner_config to register: 

```python
g = graphistry.spanner_gql_to_g(query)
```

**Expected behavior**
Should give descriptive error and how to set with register vs KeyError

**Actual behavior**
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-6-b9aea2069736> in <cell line: 0>()
      2 MATCH p = (a)-[b]->(c) where 1=1 {LIMIT_CLAUSE} return SAFE_TO_JSON(p) as path'''
      3 
----> 4 g = graphistry.spanner_gql_to_g(query)

1 frames
/usr/local/lib/python3.11/dist-packages/graphistry/PlotterBase.py in spanner_gql_to_g(self, query)
   2531 
   2532         if not hasattr(res, '_spannergraph'):
-> 2533             spanner_config = PyGraphistry._config["spanner"]
   2534             if spanner_config is not None:
   2535                 logger.debug(f"Spanner Config: {spanner_config}")
```

Issue #643, opened 2025-01-25
[BUG] spanner errors out when spanner_config does not include credentials file


DataBoyTX:
**Describe the bug**
There are 2 methods for authenticating with Spanner: 
1. gcp service account credentials file, and 
2. web-based interactive authentication 

a new optional parameter was added to register() to allow passing in spanner config: 

```
        spanner_config dict must contain the include the following keys, credentials_file is optional:
            - "project_id": The GCP project ID.
            - "instance_id": The Spanner instance ID.
            - "database_id": The Spanner database ID.
            - "credentials_file": json file API key for service accounts 
```

if using method #2 (web-based interactive auth) which does not require credentials_file, spanner_init errors with:  

``` 
SpannerConnectionError: Failed to connect to Spanner: 'SpannerGraph' object has no attribute 'credentials_file'
```

**To Reproduce**

use option 2, which does not require `credentials_file` to be set, then attempt to call 

```python
import pandas as pd
import graphistry

PROJECT_ID   = "my_project" 
INSTANCE_ID  = "my_instance" 
DATABASE_ID  = "finance-graph-db"  

# google settings, option 1: interactive login using gcloud auth application-default login (below)
SPANNER_CONF = { "project_id":  PROJECT_ID, 
                 "instance_id": INSTANCE_ID, 
                 "database_id": DATABASE_ID }


graphistry.register(api=3, personal_key_id='xxx', personal_key_secret='xxx', server='hub.graphistry.com', spanner_config=SPANNER_CONF) 

# google auth stuff: 

# Set the google project id
!gcloud config set project {PROJECT_ID}
%env GOOGLE_CLOUD_PROJECT={PROJECT_ID}

# web-based interactive login to GCP: 
!gcloud auth application-default login

# spanner graph query: 

query=f'''GRAPH FinGraph
MATCH p = (a)-[b]->(c) where 1=1 {LIMIT_CLAUSE} return SAFE_TO_JSON(p) as path'''

g = graphistry.spanner_gql_to_g(query)
```

**Expected behavior**
should not error out as `credentials_file` is not required with web-based auth

**Actual behavior**

```
AttributeError                            Traceback (most recent call last)
/usr/local/lib/python3.11/dist-packages/graphistry/plugins/spannergraph.py in __connect(self)
     84         try:
---> 85             if self.credentials_file:
     86                 connection = connect(self.instance_id, self.database_id, credentials=self.credentials_file)

[...]

PR #646, opened 2025-01-25
Fix spanner creds issues:   fixes: #634, #644 

DataBoyTX:
changes to address: 

**[BUG] spanner errors out when spanner_config does not include credentials file #643**

**[BUG] spanner queries give KeyError if spanner_config not set via register() #644**


lmeyerov (approved):
thanks!

lgtm -- merge & push tag?

PR #652, opened 2025-02-12
269 fea validator - add validator code.

SinsBre:
Added a graph validator.

Checks:
- Missing nodes
- Nan nodes
- Duplicate node ids


SinsBre:
@aucahuasi - it is failing for test-full-ai, unsure how to fix that. Known issue?

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def check_edge_destinations_exist_in_nodes(g, verbose=True):
+    if not g._edges[g._destination].isin(g._nodes[g._node]).all():
+        if verbose:
+            print("Warning: Contains destination edge IDs that do not exist in the node DataFrame. This can cause unexpected results.")
+    return True
+
+
+def validate_graph(g, verbose=True):
```
lmeyerov, line 47: - add signature to `Plottable` interface: https://github.com/graphistry/pygraphistry/blob/master/graphistry/Plottable.py

- expose for use at https://github.com/graphistry/pygraphistry/blob/320326c72251341920895f87e29ce9a4cd38d955/graphistry/PlotterBase.py#L1428

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
+def check_node_dataframe_exists(g, verbose=True):
+    if g._nodes is None:
+        if verbose:
+            print("Warning: graph was created with only edges. Skipping Node ID check if Node IDs match edge IDs. Use g2 = g.materialize_nodes() to force node df creation. Exiting.")
```
lmeyerov, line 4: switch from `print` statements to `warn` + `logger.info` depending on use

removes needed for `verbose` param

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+
+def check_edge_destinations_exist_in_nodes(g, verbose=True):
+    if not g._edges[g._destination].isin(g._nodes[g._node]).all():
+        if verbose:
+            print("Warning: Contains destination edge IDs that do not exist in the node DataFrame. This can cause unexpected results.")
+    return True
+
+
+def validate_graph(g, verbose=True):
+    if not check_node_dataframe_exists(g, verbose):
```
lmeyerov, line 48: node existence should be more like:
- either _node and _nodes are both defined or neither defined

graphistry/validate/validate_graph.py
lmeyerov, line 47: for validators, instead of doing an immediate error, good to collect errors and report them all

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+def check_node_id_defined(g, verbose=True):
+    if g._node is None:
+        if verbose:
+            print("Invalid graph: Missing Node ID. Did you forget to specify the node ID in the .nodes() function? Exiting.")
+        return False
+    return True
+
+
+def check_nan_node_ids(g, verbose=True):
+    if g._nodes[g._node].isnull().any():
```
lmeyerov, line 18: compute & report `.sum()` vs just `.any()`

a more verbose mode might return the df to help save time

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+        if verbose:
+            print("Invalid graph: Contains NaN Node IDs.")
+        return False
+    return True
+
+
+def check_duplicate_node_ids(g, verbose=True):
+    if g._nodes[g._node].duplicated().any():
```
lmeyerov, line 26: same as https://github.com/graphistry/pygraphistry/pull/652/files#r2004992104

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+        if verbose:
+            print("Invalid graph: Contains duplicate Node IDs.")
+        return False
+    return True
+
+
+def check_edge_sources_exist_in_nodes(g, verbose=True):
+    if not g._edges[g._source].isin(g._nodes[g._node]).all():
```
lmeyerov, line 34: same as https://github.com/graphistry/pygraphistry/pull/652/files#r2004992104

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+        if verbose:
+            print("Warning: Contains source edge IDs that do not exist in the node DataFrame. This can cause unexpected results.")
+    return True
+
+
+def check_edge_destinations_exist_in_nodes(g, verbose=True):
+    if not g._edges[g._destination].isin(g._nodes[g._node]).all():
```
lmeyerov, line 41: same as https://github.com/graphistry/pygraphistry/pull/652/files#r2004992104

graphistry/validate/validate_graph.py
```diff
@@ -0,0 +1,61 @@
...
+    if not check_node_dataframe_exists(g, verbose):
+        return False
+    if not check_node_id_defined(g, verbose):
+        return False
+    if not check_nan_node_ids(g, verbose):
+        return False
+    if not check_duplicate_node_ids(g, verbose):
+        return False
+    check_edge_sources_exist_in_nodes(g, verbose)  # Warnings only
+    check_edge_destinations_exist_in_nodes(g, verbose)  # Warnings only
```
lmeyerov, line 57: why warnings only?

graphistry/validate/validate_graph.py
lmeyerov, line 47: recommend also checking that dtypes of src/dst match, as does node/src if exist

lmeyerov, line 47: as a public function, should have a docstr matching other docstrs format

lmeyerov (changes requested):
see comments

bigger picture:

- [ ] always-on for validated uploads: factor out fast metadata-only checks (e.g., `validate_schema_bindings` binding name <> col name match up) from data-level checks (e.g., `validate_indexes` checking col for nan, value mismatches, etc): update `ArrowUploader::create_dataset` -> `if validate: ...` to include the fast metadata-only checks. Arguably we should do data-level checks here too, but that changes perf characteristics more fundamentally so let's not for now..

- [ ] expose on `Plotter` / `PlotterBase`, ideally calling both this + `validate_encodings`

- [ ] switch to `logging` / `warn` vs `verbse => print()`, and report error count (# hits per violated condition) and ideally all errors vs early-exist on first

- [ ] add an ipynb demo

- [ ] some potential items to warn or error on

graphistry/validate/validate_graph.py
lmeyerov, line 47: recommend also checking that dtypes of src/dst match, as does node/src if exist

type mismatch is more of a warning vs a fail, as we can support bipartite graphs, like imagine `user:str <> day:int`

lmeyerov: likewise, as a warn vs fail, if there is no overlap of src/dst edges, they may have a mismatch needing cleanup, imagine `123` <> `id:123`

graphistry/validate/validate_graph.py
DataBoyTX: Good point, I guess I was considering only graph with edge IDs that could be valid. 

However, to your point, we should error if  g._node and g._nodes exist and there are missing node IDs from g._edges src/dst, that should throw an error.

PR #655, opened 2025-04-17
feat(plottable): add "erase_files_on_fail" option to plot and upload

mj3cheun:


lmeyerov (commented):
@mj3cheun nice!

* Note we are also starting to use `def upload()` as a cleaner upload entrypoint, so would add there too..imagine GFQL uploads.
* Maybe beyond a named flag, this is a place where kwargs would be useful? I'm unsure tho, as kwargs might go to diff places. Or maybe it is a global setting. Kwargs etc may be overreach for this pr, so we do named for now, and watch for more pressure to switch to or add kwargs.

mj3cheun:
thanks! good point with the upload function will add there too

aucahuasi (approved):
Thank you!

PR #659, opened 2025-05-26
Add Kusto DB and kql support

tanmoyio:
wip

todos - metadata creation, types


graphistry/plugins/kustograph.py
```diff
@@ -0,0 +1,82 @@
...
+        self.cluster = kusto_config["cluster"]
+        self.database = kusto_config["database"]
+
+        if all(kusto_config.get(k) for k in ["client_id", "client_secret", "tenant_id"]):
+            self.credential_mode = "aad_app"
+            self.client_id = kusto_config["client_id"]
+            self.client_secret = kusto_config["client_secret"]
+            self.tenant_id = kusto_config["tenant_id"]
+        else:
+            self.credential_mode = "device_code"
```
lmeyerov: Seems a bit error prone

graphistry/plugins/kustograph.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.client = self._connect()
```
lmeyerov: Should this be lazy / on demand ?

lmeyerov: (we had similar in Louie where helps to defer)

graphistry/plugins/kustograph.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def _connect(self) -> KustoClient:
```
lmeyerov: Do we need reconnect / autoreconnect in query fail?

Also a close() ?

lmeyerov: Out of scope, but lifecycle methods for these may help Louie side

Cc @exrhizo for visibility

graphistry/pygraphistry.py
```diff
@@ -609,7 +609,11 @@ def set_spanner_config(spanner_config: Optional[Union[Dict, str]] = None):
 
         if spanner_config is not None: 
             PyGraphistry._config["spanner"] = spanner_config 
-
+    
+    @staticmethod
+    def set_kusto_config(kusto_config):
+        if kusto_config is not None:
+            PyGraphistry._config["kusto"] = kusto_config
```
lmeyerov, line 618: Hmm this pattern is increasingly risky for Louie as assume single tenant, vs cascading off a dynamic plottable instance

exrhizo: Maybe there is a way to allow louie to control lifecycle, like this:
```
client = graphistry.kusto_connect()
graphistry.kusto_query(..., client=client) # otherwise can default to the global one
```

graphistry/plugins/kustograph.py
```diff
@@ -0,0 +1,82 @@
...
+        logger.debug(f"KustoGraph execute_query(): {query}")
+        try:
+            start = time.time()
+            response = self.client.execute(self.database, query)
+
+            rows = list(response.primary_results[0])
+            col_names = [col.column_name for col in response.primary_results[0].columns]
+
+            logger.info(f"Query returned {len(rows)} rows in {time.time() - start:.3f} sec")
+            return KustoQueryResult(rows, col_names)
```
SinsBre: The execute_query works great for basic Kusto, but fails once you introduce kusto graph.

Example: `graph('HoneypotNetwork', 'FirstSnapshot')| graph-match (a)-[edge]->(v) where edge.victimPort != '443.0' project edge` <--- This returns a dataframe containing a dict per row. It is sort of like the way cypher returns graphs, but you don't have the `path=()-[]->()` option in kusto graph.

lmeyerov: Maybe this is pygraphistry's job, to run a Union[DF, Plottable]?

exrhizo: Implemented an unwrap_nested that will be applied based on a hueristic and not disabled

graphistry/plugins/kustograph.py
exrhizo: I changed it to on demand

graphistry/plugins/kustograph.py
exrhizo: I didn't add a reconnect

SinsBre:
```~/Documents/graphistry/Graphistry/pygraphistry/graphistry/PlotterBase.py in PlotterBase()
-> 2606     def kusto_query(self: Plottable, query: str, unwrap_nested: bool | None = None) -> List[pd.DataFrame]:
   2607         """
   2608         Submit a Kusto/Azure Data Explorer *query* and return result tables.

TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'```


Seeing this error when trying to run this branch now.

File: graphistry/umap_utils.py [hunk 6]
```diff
@@ -430,7 +430,7 @@ def transform_umap(self, df: pd.DataFrame,
                                   infer_on_umap_embedding=fit_umap_embedding, merge_policy=merge_policy,
                                   eps=min_dist, sample=sample, n_neighbors=n_neighbors) 
             return g
-        return emb, X, y_
+        return emb, X, y_  # type: ignore
```
exrhizo: I wonder why these are happening now

lmeyerov: @tanmoyio are you behind master, e.g., https://github.com/graphistry/pygraphistry/pull/661 ?

i had fixed these there

lmeyerov: ah i see the master merge :)

graphistry/PlotterBase.py
```diff
@@ -2282,30 +2285,50 @@ def bolt(self, driver):
...
-        :raises ValueError: If any of the required keys in `spanner_config` are missing or have invalid values.
-
         """
-        from .plugins.spannergraph import SpannerGraph
+        self._spanner_config = spanner_config
+        return self
+    
 
-        res = copy.copy(self)
+    def kusto(self: Plottable, kusto_config: KustoConfig) -> Plottable:
```
lmeyerov, line 2311: from an ergonomics, esp forward looking, would it be cleaner to be like

```python
# connect
gk = graphistry.kusto_config(**cfg)

# for users wanting typing
df = gk.kql_df(my_kql)
dfs = gk.kql_dfs(my_kql)
g2 = gk.kql_g(my_g)

# for louie, or users wanting dynamic typing
df_or_dfs_or_g = gk.kql(my_dyn_expr)
```

so:

* `kql()` - main call: short method name `kql(str) -> df | list[df] | plottable`
  * or maybe `kql<T>(str) -> T` ?
* `kql_{df,dfs,g}(str) -> T` - main call, typed variants: when return type is known
* `kusto_config()` - initial call: separate out as `kusto_config()`

lmeyerov: Separately, touching on the whole connection lifecycle question, I'm wondering about connection lifecycles here:

some variants:
* `kusto_config()` sets a config but doesn't connect, so if a bad cfg, we don't know, and connections managed later
* `kusto_config()` creates a connected client (and downstream use maintains?), and so it needs to be closed too
* `kusto_config()` creates a connected client, but if it disconnects, on the user to reconnect somehow?

graphistry/pygraphistry.py
lmeyerov: I think we're good now:

`Plottable` now has:

```python
    _spanner_config: Optional[SpannerConfig]
    _kusto_config: Optional[KustoConfig]
```

and `KustoGraph` does a cascade

```python
class KustoGraphContext:
    def __init__(self, config: Union[KustoConfig, None] = None):
        config = config or PyGraphistry._config.get("kusto")
```

lmeyerov: I do broadly agree about moving to clients, both for graphistry server and db servers. I wonder if we can move some of this stuff to type state (dunno if python supports), like `Plottable<KustoClient,SplunkClient, ...>`

lmeyerov (approved):
w00t

PR #663, opened 2025-06-10
Plugins: Give more flexibility to the user how to init client

exrhizo:
`XyzGraph` as an `XyzClientWrapper` with convenience methods

Changes:
 - Give more flexibility to users how to init (so louie can evolve independently)
 - `XyzGraph` decouple from Plottable - This way we can be more confident about independence
 
 
 Requested by Thomas
  - Client to only stay open during query and then close
    - So I made a context manager & only keep the credentials on the Plottable
    
 KustoGraph
  - `kusto_query` -> List[pd.DataFrame]
  - `query_graph` -> Plottable
  
>  It is very difficult to predict what kusto will output as far as nodes and edges, so we have an ergonomic hook to visualize graphs and the more flexible query, that just returns a list of dfs. `query` also has a hueristic for unpacking the nested tables from a graph project.


graphistry/plugins/spannergraph.py
```diff
@@ -259,8 +247,8 @@ def gql_to_graph(self, res: Plottable, query: str) -> Plottable:
         nodes_df = self.get_nodes_df(json_data)
         edges_df = self.get_edges_df(json_data)
 
-        # TODO(tcook): add more error handling here if nodes or edges are empty
-        return res.nodes(nodes_df, 'identifier').edges(edges_df, 'source', 'destination')
+        return nodes_df, edges_df
```
tanmoyio: This will be problematic for louie. Louie expects plottable from gql_to_graph. Any specific reason for this?

graphistry/PlotterBase.py
```diff
@@ -2298,20 +2298,37 @@ def spanner_init(self: Plottable, spanner_config: Dict[str, str]) -> Plottable:
...
+
+        :returns: Plottable with a Kusto connection 
+        :rtype: Plottable
+        :raises ValueError: If any of the required keys in `kusto_config` are missing or have invalid values.
+        :raises KustoConnectionError: If the connection to Kusto fails.
+        """
         from .plugins.kustograph import KustoGraph
         res = copy.copy(self)
-        res._kustograph = KustoGraph(res, kusto_config)
+        res._kustograph = KustoGraph.from_config(kusto_config)
```
lmeyerov: I'm more of a fan of the previous mode because it doesn't lose current `self` bindings from the `Plottable::copy()` , which is our typical pattern here. The important part for safety is it's a fresh base instance object for client-local state.

lmeyerov: (and if the caller wants empty, they can start with an empty)

graphistry/PlotterBase.py
```diff
@@ -2561,11 +2578,14 @@ def spanner_gql_to_g(self: Plottable, query: str) -> Plottable:
...
-                raise ValueError('spanner_config not defined. Pass spanner_config via register() and retry query.')
+                raise ValueError('spanner_config not defined. Pass spanner_config via register() or .spanner_init() and retry query.')
         
             res = res.spanner_init(spanner_config)  # type: ignore[attr-defined]
 
-        return res._spannergraph.gql_to_graph(res, query)
+        nodes, edges = res._spannergraph.gql_to_graph(query)
+        
+        # TODO(tcook): add more error handling here if nodes or edges are empty
+        return res.nodes(nodes, 'identifier').edges(edges, 'source', 'destination')
```
lmeyerov: hardcoding bindings here is a little weird to me in that it moves control of the schema out from `gql_to_graph`, causing coordination confusion, so maybe it should return bindings too? at that point, i think it should just return a plottable as before...

lmeyerov: if the goal is a fresh one, maybe:

```python
g = res._spannergraph.gql_to_graph(query)
return res.nodes(g._nodes, g._node).edges(g._edges, g._source, g._destination)
```

graphistry/plugins/spannergraph.py
```diff
@@ -38,69 +40,55 @@ class SpannerGraph:
...
+
+    def close(self) -> None:
         """
         Closes the connection to the Spanner database.
         """
-        if self.connection:
-            self.connection.close()
-            logger.info("Connection to Spanner database closed.")
+        self.connection.close()
+        logger.info("Closed Spanner connection.")
```
lmeyerov, line 76: fwiw, does this mean the caller is supposed to do `res_g = g.kusto_query_to_graph(query); res_g._kusto_client.close()` ? Not a blocker for the this PR, but am curious

lmeyerov: (In general, unsure what lifecycle pattern we want for reconnecting, closing connections, etc, getting back to the whole louie discussion)

graphistry/plugins/kustograph.py
```diff
@@ -2,6 +2,7 @@
 import time
 import json
 from typing import Any, Dict, List, Optional
+from typing_extensions import TypedDict, NotRequired
```
lmeyerov: What happens if the user does not have `azure.kusto.*` installed, will pygraphistry fail to import?

should we make these conditional somehow ,like

1. Conditional top-level imports for type safety

```python
if typing.TYPE_CHECKING:
  try:
    from azure.kusto.data import KustoClient, KustoConnectionStringBuilder
  catch ModuleImporterror:
    KustoConnectionStringBuilder = Any
```

2. Lazy runtime imports for clear errors

```python
def _connect(self) -> KustoClient:
    # trigger better runtime exn on use
    from azure.kusto.data import KustoClient, KustoConnectionStringBuilder
```

graphistry/plugins/spannergraph.py
```diff
@@ -241,7 +229,7 @@ def get_edges_df(json_data: list) -> pd.DataFrame:
         return SpannerGraph.add_type_from_label_to_df(edges_df)
 
 
-    def gql_to_graph(self, res: Plottable, query: str) -> Plottable:
+    def gql_to_graph(self, res: Plottable, query: str) -> Tuple[pd.DataFrame, pd.DataFrame]:
```
lmeyerov: 1. I like having these explicit versions to help with explicit consumers, so we can do `gql_df: str -> df` and `gql_graph: str -> plottable` and consumer has known type

2. scenarios like louie, and maybe convenience ones for users, benefit from `gql: str -> df | plottable`, and up to the consumer to handle the union type

graphistry/plugins/spannergraph.py
exrhizo: In the past the clients are never closed -- here I am assuming that the SpannerGraph is basically a client wrapper with convenience functions

Usually a user is using the graphistry object that seems more optimized for notebooks then for memory management

If someone wants memory management, then I think they can instantiate the `SpannerGraph` directly, like I want to do in louie

graphistry/PlotterBase.py
exrhizo: This is the crux of what I am trying to reverse

I don't think these `XyzGraphs` should know anything about what is bound to the Plottable, that is the whole issue we are having in louie with it being unclear if these are multi tenant / thread safe / global values involved

graphistry/PlotterBase.py
exrhizo: I think it could be that spanner has two functions one that returns the nodes, edges and the other that binds it to the current plottable

graphistry/plugins/spannergraph.py
exrhizo: Louie pins the graphistry version so we can change the interface, I don't feel as strongly about this - see comments to Leo

graphistry/plugins/spannergraph.py
exrhizo: The weird thing to me is what if I don't want to chain in my Plottable

graphistry/plugins/spannergraph.py
lmeyerov: I'm a bit lost on internal methods vs public

wrt public:

- see: https://github.com/graphistry/pygraphistry/pull/663#discussion_r2138483855

- Louie afaict benefits from a generic `str -> pd.DataFrame | Plottable`, where louie can do its own `isinstance(...)` and handle appropriately 

- I think beginner/vibes-y python users would benefit from that more dynamic type signature too...
- ... and only when we do more 'serious' typed code would we want the explicit `gql_to_g` / `gql_to_df` variants (or maybe there is magic where we can `gql<pd.DataFrame>(...)`?). So I'd want "both" vs picking

I hit the same thing when we exposed the remote python+gfql endpoints and we wanted to support multiple return types

graphistry/plugins/spannergraph.py
```diff
@@ -4,6 +4,8 @@
 import time
 from typing import Any, List, Dict, Optional
 
+from google.cloud.spanner_dbapi.connection import connect, Connection
```
lmeyerov: same imports issue: https://github.com/graphistry/pygraphistry/pull/663#discussion_r2138476427

lmeyerov: we can release in louie off of this PR sooner than later, but blocks landing as a public graphistry release

graphistry/plugins/kustograph.py
lmeyerov: fwiw, louie can build & release off of a PR version here, so fixing this is just about landing in pyg main ... not great but allows progress

graphistry/PlotterBase.py
exrhizo: see the way I did it this time

exrhizo:
Merged into https://github.com/graphistry/pygraphistry/pull/659

PR #668, opened 2025-06-20
feat: New redteam50k dataset to be used in Microsoft Kusto

SinsBre:
Added a new version of the redteam50k dataset.

It is the two datasets used in the notebook: "redteam-umap-gtc-gpu.ipynb" merged into one csv.
Dataset is used in future Kusto graph notebooks.


SinsBre:
I had missed one data wrangle Alex M does in his redteam-demo. `red_team['feats2'] = red_team['feats']`
It is basically one set with that merges the file where RED = 0 and the file where RED = 1.
The file makes life easier when loading in Azure

Issue #557, opened 2024-03-17
[FEA] First-class client support for easier multitenancy


lmeyerov:
**Is your feature request related to a problem? Please describe.**

Currently, multitenant apps need to lock-and-register at `.plot()` to ensure the right credentials are used at upload time

It'd be easier if we had some sort of first-class functional client model where a client can be configured, with its own creds, and as long as the right client is used, the plot works

**Describe the solution you'd like**

### Internal base interface `Client`

* Mostly around JWT & org manipulations
*  `g.client() : Client` , `g._client.token()`, `g._client.refresh()`, `g._client.switch()`, etc
* Variants: `ClientUnconfigured`, `ClientUserPass`, `ClientAPIKey`, `ClientSSO`, ...

### Global and root chain clients

Recall that `g` objects functionally build up, like `g2 = g1.edges(df)`

Each plottable will either set its client, or a reference to another plottable and use its dynamically resolvable client, to be dynamically determined upon use (e.g., plot). For convenience, the client will be set for legacy convenient global notebook cases + more controllable multitenant scenarios:

1. Backwards compatible global/single-tenant: for notebook etc users + legacy
  * module has a single global client
  * `graphistry.register(); g = graphistry.bind(...); g.plot()`:
    * top-level `graphistry.register()` mutates global client
    * by default, chained items use the latest global client
  * add `graphistry.register(client=...)`
    * potentially deprecate old access flows
    * modifying the client `c` bound in `client=c` impacts all plottables using it
  * access via `c = graphistry.client_root()` and `graphistry.client_root(c)`

2. New: Custom chainable roots, for multitenant
  * Generate a top-level g via explicit client instantiation
    * `g = ClientUserPass().g()`
    * `g = graphistry.client(ClientUserPass())`

  * Mid-chain, setting a new root that impacts all downstream, and optionally the current
    * `g2 = g1.client(Client(...))` <- `g2` uses a diff client
    * `g1.set_client(Client(..)` <- update `g1`'s client, and also downstream like `g2`

### Question

Default behavior during chaining: I can imagine `g.copy()` chaining to automatically do one of these:
    *  `g2._client_root = g1` # so changing g1 will change g2
    * `g2._client_root = g1._client_root` # so changing g1's root will change g2's, but changing g1 will not (unless it has the root)

Basically, is:
*  `g._client_root : Client`, so by ref eq of a client set at time of chain
[...]

PR #670, opened 2025-06-21
graphistry.client() & Iterate on kusto and spanner mixin & +types

exrhizo:
`graphistry.client()` - addresses #557 

Iterating on the Kusto and Spanner to be mixins that can exist standalone.

Also adding the notion of `Session`
 - `ClientSession` - Main graphistry global config, to be detachable
 - `SpannerSession` & `KustoSession` - persist so interact session can be initialized once
 
Making `PyGraphistry` a singleton rather than a Class with static methods
  
  ### TODO
   - [x] Write tests
   - [x] Write docs
   - [x] Think about how to have docs not repeated so much for Kusto and Spanner..


graphistry/PlotterBase.py
```diff
@@ -1397,8 +1504,8 @@ def server(self, v: Optional[str] = None) -> str:
         """
         from .pygraphistry import PyGraphistry
         if v is not None:
-            PyGraphistry._config['server'] = v
-        return PyGraphistry._config['server']
+            PyGraphistry._session.hostname = v
+        return PyGraphistry._session.hostname
```
lmeyerov: For backwards compatibility, maybe we create a proxy object for PyGraphistry._config that implements get/set proxies and does depreciation warnings?

exrhizo: yes, that is my plan, I wanted to break it first so I found all the places to change

lmeyerov: (the edits here seem fine, so more a top-level comment)

graphistry/PlotterBase.py
```diff
@@ -1408,8 +1515,8 @@ def protocol(self, v: Optional[str] = None) -> str:
         """
         from .pygraphistry import PyGraphistry
         if v is not None:
-            PyGraphistry._config['protocol'] = v
-        return PyGraphistry._config['protocol']
+            PyGraphistry._session.protocol = v
```
lmeyerov: I thought self._session?

graphistry/PlotterBase.py
```diff
@@ -1408,8 +1515,8 @@ def protocol(self, v: Optional[str] = None) -> str:
...
+        return PyGraphistry._session.protocol
```
lmeyerov: Same

graphistry/PlotterBase.py
```diff
@@ -1421,8 +1528,8 @@ def client_protocol_hostname(self, v: Optional[str] = None) -> str:
         """
         from .pygraphistry import PyGraphistry
         if v is not None:
-            PyGraphistry._config['client_protocol_hostname'] = v
-        return PyGraphistry._config['client_protocol_hostname']
+            PyGraphistry._session.client_protocol_hostname = v
+        return PyGraphistry._session.client_protocol_hostname or f"{PyGraphistry.protocol()}://{PyGraphistry.server()}"
```
lmeyerov: I think most Pygraphistry._session calls should be self._session now ?

exrhizo: Yes, I wanted to get this singleton working as a sort of way point

lmeyerov: Ah Ok so goal of this PR is to be backwards compatible changes, client mode just for ksuto+spanner, and later PR does for rest?

graphistry/client_session.py
```diff
@@ -0,0 +1,158 @@
...
+from graphistry.privacy import Privacy
+from . import util
+
+
+
+config_paths = [
+    os.path.join("/etc/graphistry", ".pygraphistry"),
+    os.path.join(os.path.expanduser("~"), ".pygraphistry"),
+    os.environ.get("PYGRAPHISTRY_CONFIG", ""),  # user-override path
+]
```
lmeyerov, line 23: fwiw, while we're thinking about better config patterns, dask seems interesting here: https://docs.dask.org/en/latest/configuration.html

graphistry/PlotterBase.py
lmeyerov: maybe useful:

* gak accesses `PyGraphistry._is_authenticated`, `PyGraphistry._config["api_token"]` - https://github.com/search?q=repo%3Agraphistry%2Fgraph-app-kit%20PyGraphistry&type=code

* louie doesn't really inspect the getters, it mostly manages its own tenants and just calls setters & callers -- `register()`, `org_name()`, `privacy()`, `refresh()`, ... and reads back ~fresh call results

graphistry/client_session.py
exrhizo: pydantic has patterns like the nested fields

graphistry/PlotterBase.py
exrhizo: I am going to see how hard it is, maybe claude can finish up the `Self` part

graphistry/client_session.py
lmeyerov: the pydantic 1 -> 2 mass breaking changes left a bad taste in my mouth wrt it being a dependency for frameworks so been resisting add here

it's a diff story in our internals where we have more control

i was wondering if there's a thinner library we can pull in for config mgmt, looks like dynaconf might be good as 0 required deps, popular, etc: https://github.com/dynaconf/dynaconf

but again.. maybe not worth the dependency annoyance for a client framework lib

exrhizo:
hm
```
_pd_hash_to_arrow
_cudf_hash_to_arrow
_umap_param_to_g
_feat_param_to_g
```

Do these need to be segmented?

graphistry/arrow_uploader.py
```diff
@@ -13,6 +14,54 @@
 logger = setup_logger(__name__)
 
 class ArrowUploader:
+
+    def __init__(
+        self,
+        *,
+        client_session: Optional[ClientSession] = None,
```
lmeyerov: can we honor original arg order for this PR, and do new ones at the end?

i believe we have some users relying on this, and it'd be a needlessly breaking change

graphistry/pygraphistry.py
```diff
@@ -316,48 +268,47 @@ def _handle_auth_url(auth_url, sso_timeout, sso_opt_into_type):
...
+            #         return self.api_token()
             #     print("Keep trying to get token ...")
             #     time.sleep(5)
 
             print("Please run graphistry.sso_get_token() to complete the authentication")
             return None
 
-    @staticmethod
-    def sso_get_token():
+    def sso_get_token(self) -> Optional[str]:
```
lmeyerov, line 341: i think moving all these from global static exports to exporting an instance method (see bottom of this file for re-export) is technically a breaking change

these global methods were a bad enough design that i don't think it gets used in practice. supporting backwards compatibility doesn't feel worth solving.

so maybe the answer is the breaking change is ok, and we get rid of the exports at the bottom of this file as make even less sense now?

graphistry/tests/test_arrow_uploader.py
```diff
@@ -156,25 +156,25 @@ def test_au_e_enc_full(self):
         }
 
     def test_cascade_privacy_settings_default_global(self):
-        PyGraphistry._config["privacy"] = None
+        PyGraphistry.session.privacy = None
```
lmeyerov, line 159: as we start pushing the client driver -- not this pr -- probably want more testing of the new setup

graphistry/pygraphistry.py
exrhizo: I am not sure what usages would be broken by having a global singleton w/ methods vs static functions that use a global state variable. I think the interface is the same unless someone is doing funny mucking in internals.

```
# I don't think this looks good
from graphistry import PyGraphistry as graphistry
g = graphistry.bind()
```

in practice people seem to use `register` and `bind`

I could use the magic to have the module resolve to the singleton, until client is used? But that wouldn't be good for code completion

lmeyerov: i think we're good

from slack discussion, i missed:

```
# Create the global PyGraphistry instance
PyGraphistry = PyGraphistryClient()
```

lmeyerov (approved):
approving --

* `ArrowUploader` new optional param should be last to avoid breaking public interface, see: https://github.com/graphistry/pygraphistry/pull/670/files#r2164955768
* please add `changelog.md`, it's mostly internal changes but maybe some external. i wasn't tracking the kusto/spanner stuff here so unsure if anything to say. docs around multitenant client stuff can be a follow-on pr imo, no need to block..
* we should do as a major version, `0.39.0`, as the metaobject protocol etc changes might be real for some folks, even though we tried to make backwards compatible

docs/source/server/concurrency.rst
```diff
@@ -0,0 +1,43 @@
+Multi-Tenancy & Concurrency
+==========================
+
+Graphistry `register()` and `plot()` are stateful. Methods are the concurreny boundary.
+
+This is because on `plot`, the `ArrowUploader` manages user auth token refresh on session
+state. 
+
+`Plottable._pygraphistry.session` = `Graphistry.session`
```
lmeyerov: Discussion of internal details seems confusing for user docs: `ArrowUploader`, `._pygraphistry.session`, ...

docs/source/server/concurrency.rst
```diff
@@ -0,0 +1,43 @@
+Multi-Tenancy & Concurrency
+==========================
+
+Graphistry `register()` and `plot()` are stateful. Methods are the concurreny boundary.
```
lmeyerov: Maybe flip around like

> To safely use pygraphistry in concurrent and multitenant settings, use client objects. Use of top-level calls like `register()` and `plot()` are unsafe in these settings as they use global variables, unless you are using them through client objects.
> Client objects automatically isolate session state like graphistry server and database tokens. Each client and derived plottable objects are safe for use within a single concurrency context like a thread or event loop, and you can have multiple in different threads and event loops.

docs/source/server/concurrency.rst
```diff
@@ -0,0 +1,43 @@
...
+A `Plottable`'s state is tied to the client used to create it.
+
+Example
+-------
+
+.. code-block:: python
+
+   import graphistry
+   # Independent clients
+   alice = graphistry.client().register(api=3, username='alice', password='pw')
```
lmeyerov: to follow convention, maybe

```python
alice_g = graphistry.client()
alice_g.register(...)

...
```

docs/source/server/concurrency.rst
```diff
@@ -0,0 +1,43 @@
...
+state. 
+
+`Plottable._pygraphistry.session` = `Graphistry.session`
+
+A `Plottable`'s state is tied to the client used to create it.
+
+Example
+-------
+
+.. code-block:: python
```
lmeyerov, line 27: While verbose, it's helpful to do examples that we see here frequently so we can just send docs links when people ask (or search for those keywords) --

* Multi-tenant (this example)
* Multiple servers (staging vs prod)
* Multi-threaded (can be this example, and just add a comment like `## thread 1`)

exrhizo: done

docs/source/server/concurrency.rst
exrhizo: done

docs/source/server/concurrency.rst
exrhizo: done

docs/source/server/concurrency.rst
exrhizo: moved it

PR #683, opened 2025-06-27
Lock global caches

exrhizo:
### Possible Race Conditions

1. **Global Cache Dictionary Corruption**
   - `_pd_hash_to_arrow`, `_cudf_hash_to_arrow`, `_umap_param_to_g`, `_feat_param_to_g` are class-level caches shared across ALL instances

2. **Check-Then-Act Race Conditions**
   ```python
   # Before fix - UNSAFE
   if hashed in PlotterBase._pd_hash_to_arrow:  # Thread 1 & 2 both check
       return PlotterBase._pd_hash_to_arrow[hashed].v
   # Both threads see cache miss, both compute conversion, duplicate work
   ```

3. **Global Variable Race in cache_coercion()**
   ```python
   # Before fix - UNSAFE
   _cache_coercion_val = v        # Thread 1 sets value
   # Thread 2 overwrites before Thread 1 uses it
   out = cache_coercion_helper(k) # Thread 1 gets wrong value\!
   ```

4. **`client(inherit=False)` Limitation**
   - While `client(inherit=False)` isolates session state, it does NOT isolate global caches
   - Even "isolated" clients share the same cache dictionaries

###  Changes Made

- **PlotterBase.py**: Added `_cache_lock`, protected pandas/cudf cache operations
- **util.py**: Fixed `cache_coercion()` race condition with dedicated lock
- **utils/plottable_memoize.py**: Protected UMAP/feature cache access

## Benefits

✅ **Prevents State Corruption** - All cache dictionaries now have synchronized access  
✅ **Fixes Global Variable Races** - The `_cache_coercion_val` race condition is resolved  
✅ **Minimal Performance Impact** - Uses `RLock()` which allows recursive locking  
✅ **Simple Implementation** - Single shared lock prevents deadlocks  
✅ **Maintains Existing API** - No changes to public interfaces  

#### 1. **Per-Cache Fine-Grained Locking**
```python
_pd_cache_lock = threading.RLock()
_cudf_cache_lock = threading.RLock() 
_umap_cache_lock = threading.RLock()
_feat_cache_lock = threading.RLock()
```
**Pros**: Better concurrency, separate hot/cold paths  
**Cons**: More complex, potential deadlock scenarios, more locks to manage

#### 2. **Per-Client Instance Caches**
```python
class PlotterBase:
    def __init__(self):
        self._pd_hash_to_arrow = WeakValueDictionary()  # instance-level
        self._cudf_hash_to_arrow = WeakValueDictionary()
```
**Pros**: True isolation, no locking needed, better for `client(inherit=False)`  
**Cons**: Higher memory usage, loss of cross-client cache benefits, major API change

#### 3. **Lock-Free Thread-Safe Data Structures**
```python
from concurrent.futures import ThreadPoolExecutor
# Use thread-safe dict implementations
```
**Pros**: Better performance, no lock contention  
**Cons**: Limited Python options, complex implementation, potential memory overhead

#### 4. **Thread-Local Caches**
```python
import threading
_thread_local_caches = threading.local()
```
**Pros**: No synchronization needed, perfect isolation  
**Cons**: No cache sharing benefits, higher memory usage per thread

🤖 Generated with [Claude Code](https://claude.ai/code)


graphistry/PlotterBase.py
```diff
@@ -102,17 +102,20 @@ class PlotterBase(Plottable):
     _defaultEdgeSourceId = SRC
     _defaultEdgeDestinationId = DST
     
+    # Thread-safe global caches with lock protection
+    _cache_lock = threading.RLock()
     _pd_hash_to_arrow : WeakValueDictionary = WeakValueDictionary()
     _cudf_hash_to_arrow : WeakValueDictionary = WeakValueDictionary()
     _umap_param_to_g : WeakValueDictionary = WeakValueDictionary()
     _feat_param_to_g : WeakValueDictionary = WeakValueDictionary()
```
lmeyerov, line 110: i don't think a client is thread safe to beginwith, so maybe the starting question is what thread safety we are trying to achieve

ex: "multithreading support is by single-threaded clients"
- you can have diff clients access across diff threads/loops
- a client and sub-plottables are only guaranteed safely accessed within a thread / event loop
- access to the client / sub-plottables by other threads / event loops require manual coordination at method boundaries, e.g., locks
  - the only known exception is the special async SSO method as it is essentially transaction across 2 method calls
- you can access multiple clients on different threads without requiring coordination

given such a model, i'm not sure we need this change

lmeyerov: it might be useful to have a client tutorial talking about common client scenarios, like:

- (default: global)
- multiple users / same server
- multiple servers
- thread safety

lmeyerov: separately, for this specific cache sharing scenario, i'm actually thinking about conflicts a bit differently - less on concurrency, more on security & sharing

imagine 2 louie users in the same dthread make diff viz of the same df:

- reusing file uploads would be good (faster UX, less uploads)..
   - one way is a shared global cache across Clients is good: hoist caches to a global and share across clients
   - another way is the loiue case can be handled outside, e.g., we track if a file_id is known for a df, and on louie to supply that!
- either way, server handles file security so we don't worry about the client
- if pygraphistry does the sharing, there is an edge case of independent DFs in the cache from independent contexts, imagine a tutorial, which the server would reject due to violating reuse of the first uploader's df. we can probably protect from this in practice tho w/ better hashing if we start seeing that, eg, better metadata controls

exrhizo:
Added docs on concurrency model: https://github.com/graphistry/pygraphistry/pull/670/commits/0ac651bbf64eedd426912f773cdca61584dfad85

PR #684, opened 2025-06-27
Improve global Arrow Cache Performance

exrhizo:
No longer keep the last 100 tables in memory, instead use id() and hash of table metadata.

```python
_ID_CACHE: Dict[int, "MemoizedFileUpload"] = {}
_SIG_CACHE: Dict[int, "MemoizedFileUpload"] = {}
```

When the table is garbage collected: `weakref.finalize(table, _evict_id_cache, obj_id)`

```python
def _compute_signature(table: pa.Table) -> int:
    """
    Pure structural hash: schema, metadata, column order.
    Avoids storing the table itself.
    """
    schema_str = str(table.schema)
    meta_items = tuple(sorted((table.schema.metadata or {}).items()))
    col_names = tuple(table.column_names)
    # 64‑bit stable hash via sha1 → int
    sig_bytes = (
        schema_str.encode()
        + b"|"
        + str(meta_items).encode()
        + b"|"
        + str(col_names).encode()
    )
    return int.from_bytes(hashlib.sha1(sig_bytes).digest()[:8], "big", signed=False)
```


graphistry/ArrowFileUploader.py
```diff
@@ -109,72 +117,114 @@ def post_arrow(self, arr: pa.Table, file_id: str, url_opts: str = 'erase=true')
...
-                    raise RuntimeError("Erased uploaded file contents upon failure (file_id still valid), see errors", out['errors'])
+                    raise RuntimeError(
+                        "Erased uploaded file contents upon failure "
+                        "(file_id still valid), see errors",
+                        out["errors"],
+                    )
             return out
         except Exception as e:
-            logger.error('Failed uploading file: %s', res.text, exc_info=True)
+            logger.error("Failed uploading file: %s", res.text, exc_info=True)
```
lmeyerov: fwiw lint changes can likely be reverted... maybe we should have a flake setting to help standardize & avoid this kind of review noise?

lmeyerov (changes requested):
this seems close, but i think the signature case is buggy: tables with same schema but diff cell values would incorrectly match


maybe we want:

* `_compute_signature_schema` (fast): keep current schema-level so constant-time, but it doesn't guarantee the content match

* `_compute_signature_full` (slow): slower b/c analyzes schema + entire data payload, but we only need to run it when the schema matches, and can skip otherwise


Separately, we might be able to do another cheap data-sensitive hash in addition to the `_schema` without going full data:

* row count
* number bytes
* maybe: hashes of vals in first/last row (up to some reasonable col count)

graphistry/ArrowFileUploader.py
```diff
@@ -109,72 +109,129 @@ def post_arrow(self, arr: pa.Table, file_id: str, url_opts: str = 'erase=true')
...
 
         if memoize:
-            wrapped = WrappedTable(arr)
-            cache_arr(wrapped)
-            DF_TO_FILE_ID_CACHE[wrapped] = out
-            logger.debug('Memoized arrow->file_id %s', file_id)
-        
-        return out.file_id, out.output
+            # Compute full hash (may already be available)
+            fh = _hash_full_table(arr)
```
lmeyerov: micro opt: reuse :156's  if same metadata / diff val (so `fh is not None`)

i'm unsure how common this case is and how perceptible

graphistry/ArrowFileUploader.py
```diff
@@ -109,72 +109,129 @@ def post_arrow(self, arr: pa.Table, file_id: str, url_opts: str = 'erase=true')
...
+                last_v = col.slice(num_rows - 1, 1).to_pylist()[0]
+            except Exception:
+                first_v = last_v = None
+            digest.update(str(first_v).encode())
+            digest.update(str(last_v).encode())
+
+    return int.from_bytes(digest.digest()[:8], "big", signed=False)
+
+
+def _hash_full_table(table: pa.Table) -> int:
```
lmeyerov, line 211: take optional schema metadata hash, and compute if not passed in?

graphistry/ArrowFileUploader.py
```diff
@@ -109,72 +109,130 @@ def post_arrow(self, arr: pa.Table, file_id: str, url_opts: str = 'erase=true')
...
-            logger.error('Failed uploading file: %s', res.text, exc_info=True)
+            text = res.text if res.text else str(res)
+            logger.error("Failed uploading file: %s", text, exc_info=True)
             raise e
 
     ###
 
     def create_and_post_file(
-        self, arr: pa.Table, file_id: Optional[str] = None, file_opts: dict = {}, upload_url_opts: str = 'erase=true', memoize: bool = True
-    ) -> Tuple[str, dict]:
```
lmeyerov, line 127: changing the return type here may be a breaking change, should we preserve the old return type, or do a more subtle breaking change like `Tuple[str, Optional[dict]]` , where first call returns the response, and subsequent return None?

I see similar elsewhere, like `post_arrow` / `_post_arrow`... these are public methods that breaking changes don't seem too motivated on wrt this pr

graphistry/ArrowFileUploader.py
lmeyerov: (i'm unclear if this is a correctness issue)

lmeyerov (commented):
Spotted 2 more potential correctness / breaking change issues

The perf one is a microopt so don't super care in terms of unblocking this pr

graphistry/ArrowFileUploader.py
exrhizo: reverted, I thought of this as internal

graphistry/ArrowFileUploader.py
exrhizo: it has
```
digest.update(str(table.schema).encode())
```

I want to move on, I suspect not a significant improvement in performance

PR #692, opened 2025-07-08
kusto notebook fixes for GRAPH_NAME and register links fixes 

DataBoyTX:
**commit: `487efd5`**

- fixes a bug in the kusto demo notebook where graph_name was not defined in a cell, and later defined as GRAPH_NAME, switched to use just `GRAPH_NAME`

- replaced register() cell with personal tokens to use standard cell for register to keep consistent across notebooks

- added timestamp to snapshot name to avoid name collisions the 2nd time the notebook is run

**commit: `df632da`**

- fixes a broken link in 47 notebooks to point to updated register() docs


lmeyerov (approved):
tx!

PR #733, opened 2025-08-14
fix(auth): chain_remote and python_remote now use client session instead of global singleton

lmeyerov:
## Summary
Fixes authentication failure when using `client.register()` with API keys for remote operations.

## Problem
`chain_remote()` and `python_remote_g()` were using the global PyGraphistry singleton instead of the client's session, causing "Must call login() first" errors when using client-specific authentication.

## Solution  
- Use `self._pygraphistry.refresh()` and `self.session.api_token` instead of PyGraphistry singleton
- Add comprehensive regression tests (7 tests) 
- Fix Python 3.12 compatibility in tests

## Changes
- `graphistry/compute/chain_remote.py` - Use instance session (lines 30-33)
- `graphistry/compute/python_remote.py` - Use instance session (lines 110-113)  
- `graphistry/tests/test_chain_remote_auth.py` - New regression tests
- `docker/test-cpu-local-minimal.sh` - Support PYTHON_VERSION env var

## Testing
✅ All tests pass on Python 3.8-3.12
✅ CI is green


exrhizo (approved):
LGTM

PR #735, opened 2025-08-22
Kusto graph syntax error fix 

SinsBre:
Issue:

If a Kusto graph has any column called "NodeId", "src" or "dst", Kusto throws an error:

Semantic error: SEM0001: graph-to-table: column named 'src' already exists

Prepend node, source and destination with **g_** to account for syntax errors from Kusto if a graph contains columns called "NodeId", "src" or "dst".


lmeyerov:
ci is red, please fix

lmeyerov:
Rerunning ci, may have been spurious

PR #736, opened 2025-08-24
Louie type feedback

exrhizo:
Surface more in Plottable interface, match types, always return g from umap


graphistry/plugins_types/embed_types.py
```diff
@@ -0,0 +1,24 @@
...
+if TYPE_CHECKING:
+    _, torch, _, _, _, _, _, _ = lazy_embed_import()
+    TT = torch.Tensor
+else:
+    TT = Any
+    torch = Any
+
+
+XSymbolic = Optional[Union[pd.DataFrame, 'np.ndarray', List[str]]]
+ProtoSymbolic = Optional[Union[str, Callable[[TT, TT, TT], TT]]]  # type: ignore
```
lmeyerov: Do we need the type ignore ?

graphistry/umap_utils.py
```diff
@@ -865,8 +866,7 @@ def umap(
         if dbscan:
             res = res.dbscan(min_dist=min_dist, kind=kind, fit_umap_embedding=True)  # type: ignore
 
-        if not inplace:
-            return res
+        return res
```
lmeyerov: Hmm most seemed just refactor but this may be a logic change ?

exrhizo: reverted and added the overloads

graphistry/plugins_types/embed_types.py
exrhizo: Got it working

lmeyerov (approved):
Tx!

Issue #738, opened 2025-08-26
[BUG] unsupported parameter passed to cudf during gfql execution


mj3cheun:
**Describe the bug**
Contains (when inside JSON Chain operator, the same issue may or may not exist using python Chain operator) sends an "na" parameter to cudf which is not supported by cudf (but may be supported by pandas). this results in cudf throwing an error.

it would also interest me to learn what other operators have similar issues (of unsupported arguments).

```
......
forge-etl-python-1  |     return g.chain(ops, engine="cudf")
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/ComputeMixin.py", line 463, in chain
forge-etl-python-1  |     return chain_base(self, *args, **kwargs)
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/chain.py", line 295, in chain
forge-etl-python-1  |     op(
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/ast.py", line 173, in __call__
forge-etl-python-1  |     .filter_nodes_by_dict(self.filter_dict)
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/ComputeMixin.py", line 455, in filter_nodes_by_dict
forge-etl-python-1  |     return filter_nodes_by_dict_base(self, *args, **kwargs)
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/filter_by_dict.py", line 56, in filter_nodes_by_dict
forge-etl-python-1  |     nodes2 = filter_by_dict(self._nodes, filter_dict, engine)
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/filter_by_dict.py", line 48, in filter_by_dict
forge-etl-python-1  |     hits = hits & op(df[col])
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/graphistry/compute/predicates/str.py", line 17, in __call__
forge-etl-python-1  |     return s.str.contains(self.pat, self.case, self.flags, self.na, self.regex)
forge-etl-python-1  |   File "/opt/conda/lib/python3.10/site-packages/cudf/core/column/string.py", line 782, in contains
forge-etl-python-1  |     raise NotImplementedError("`na` parameter is not yet supported")
[...]

PR #739, opened 2025-08-27
fix contains, startswith, endswith, and match string predicates

mj3cheun:
Fixes https://github.com/graphistry/pygraphistry/issues/738


graphistry/compute/predicates/str.py
```diff
@@ -14,8 +14,24 @@ def __init__(self, pat: str, case: bool = True, flags: int = 0, na: Optional[boo
         self.regex = regex
 
     def __call__(self, s: SeriesT) -> SeriesT:
-        return s.str.contains(self.pat, self.case, self.flags, self.na, self.regex)
-    
+        is_cudf = hasattr(s, '__module__') and 'cudf' in s.__module__
+
+        # workaround cuDF not supporting 'case' and 'na' parameters https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.core.accessors.string.stringmethods.contains/
+        if is_cudf:
```
lmeyerov, line 20: Can we have same universal behavior ? Afaict that means just removing the outer if here

lmeyerov (approved):
Awesome, thank you!

I assume you ran GPU test locally? Some reason I thought we had GPU CI set up for pygaphistry but I didn't see it in the GHA run

PR #828, opened 2025-11-04
Adding additional exports from graphistry.compute to top-level graphistry module

ahwitz:
Small doc change from working with Leo on something: in 0.45.7, `let` and `ref` only seem to be accessible through the `graphistry.compute` module, not through base `graphistry`. This updates the doc strings in one of the GFQL pages to point to the new module. Can go back through the other doc pages to clean those up if you want me to as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Development]
 <!-- Do Not Erase This Section - Used for tracking unreleased changes -->
 
+## [0.45.8 - 2025-11-03]
+
 ### Changed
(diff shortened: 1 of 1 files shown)


lmeyerov:
Thank you for catching!

- The immediate fix would be to export these: https://github.com/graphistry/pygraphistry/blob/ab4c5d4fdf0af751bc0c153873f56303aa7d0972/graphistry/__init__.py#L64

- Can you add a changelog.md entry as well (0.45.8), and then I can take through publishing?

We've started getting the pure ipynb's tested by ci, but more to go to catch all these...

ahwitz:
Got the `__init__` file updated, flipped all the lines in the ipynb (there was one that still ref'd `graphistry.compute`, it now refs `graphistry` directly) and added a changelog line.

Did this all using the default Github web editors, so I'mma watch for the tests to finish just in case I messed something up.

lmeyerov (approved):
tx!

PR #832, opened 2025-11-13
Fix per-client org propagation

lmeyerov:
## Summary
- clarify Hub bug [graphistry/graphistry#2933](https://github.com/graphistry/graphistry/issues/2933): `/api-token-auth/` accepts `org_name` and returns that org as active, but server-side session stays on the default/personal org, so paid-tier GFQL requests keep failing the free-tier limits unless the client immediately calls `/api/v2/o/<slug>/switch/`
- shift that switch into ArrowUploader login flows (`_finalize_login()` renamed from `_handle_login_response()`), so username/password, API key, SSO, and token refresh pathways always perform the extra switch call
- introduce `_maybe_switch_org()` guard that tracks the last `(org, token)` on the session to avoid redundant HTTP calls, covering refresh and SSO flows previously skipping the switch
- expand regression tests for ArrowUploader, `GraphistryClient.register`, and refresh flows (with fallback to stdlib `unittest.mock` when `mock` isn’t installed)
- fix GFQL hypergraph `return_as=*` flows to skip the Protocol `isinstance()` check that triggered pandas Styler/Jinja imports when returning DataFrames, plus add a regression test guarding against that dependency

## Testing
- `python -m pytest graphistry/tests/test_arrow_uploader.py -k org`
- `python -m pytest graphistry/tests/test_client_session.py -k org`
- `python -m pytest graphistry/tests/test_pygraphistry.py -k refresh_switches_org`
- `python -m pytest graphistry/tests/compute/test_gfql_hypergraph.py -xvs`
- Hub regression (paid org path) from a Graphistry repo checkout:
  ```bash
  docker run --rm --gpus all \
    -v /path/to/graphistry3/apps/forge/etl-server-python:/workspace:ro \
    -v /path/to/pygraphistry:/pygraphistry:ro \
    -w /workspace \
    -e PYTHONPATH=/workspace:/pygraphistry \
    -e HUB_GFQL_TEST=1 \
    -e HUB_GFQL_USERNAME=$HUB_GFQL_USERNAME \
    -e HUB_GFQL_PASSWORD=$HUB_GFQL_PASSWORD \
    -e HUB_GFQL_TEAM_ORG_SLUG=$HUB_ORG \
    graphistry/etl-server-python:v2.45.0-11.8 \
    pytest test/server/client/processing/test_hub_org_team_limit.py -vv -rs
  ```
- External repro (optional): run any script that binds a ~20k node dataset via `gfql_remote([n()])` while toggling `PYTHONPATH` between patched/origin pygraphistry checkouts to confirm the paid-org request succeeds only after this patch
  ```bash
  PYTHONPATH=/path/to/pygraphistry python repro.py
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Development]
 <!-- Do Not Erase This Section - Used for tracking unreleased changes -->
 
+## [0.45.10 - 2025-11-19]
+
 ### Added
 - **GFQL / Oracle**: Introduced `graphistry.gfql.ref.enumerator`, a pandas-only reference implementation that enumerates fixed-length chains, enforces local + same-path predicates, applies strict null semantics, enforces safety caps, and emits alias tags/optional path bindings for use as a correctness oracle.
 


graphistry/arrow_uploader.py
```diff
@@ -261,14 +283,15 @@ def pkey_login(self, personal_key_id: str, personal_key_secret: str, org_name: O
...
+        return self._finalize_login(out, org_name)
 
-    def _handle_login_response(self, out: requests.Response, org_name: Optional[str]) -> 'ArrowUploader':
+    def _finalize_login(self, out: requests.Response, org_name: Optional[str]) -> 'ArrowUploader':
         from .pygraphistry import PyGraphistry
         json_response = None
         try:
             json_response = out.json()
-            if not ('token' in json_response):
+            token_value = json_response.get('token')
```
vaimdev: suggest to change to `json_response.get('token', None)`, to avoid exception

lmeyerov:
green: https://github.com/graphistry/pygraphistry/actions/runs/19527097029

PR #841, opened 2025-11-17
Add GFQL reference enumerator oracle

### LINKED ISSUE #836: GFQL: Reference path enumerator & oracle
## Why
Establish a definitive semantics source for **linear GFQL patterns**, covering both:
- **local step predicates** (per-node/per-edge filters) and
- **same-path multihop WHERE clauses** over named steps.

This oracle drives correctness for pandas/cuDF engines and freezes null/NaN behavior.

## Deliverables
- `gfql/ref/enumerator.py` that:
  - Enumerates every valid fixed-length path (v0, e1, …, vL) for a given chain of steps
  - Applies local predicates on each node/edge hop
  - Evaluates same-path WHERE predicates over named steps (e.g., `a.x > c.y`, `a.id = c.prev_id`) with no path variables
  - Emits:
    - `V_res`: nodes appearing on ≥1 satisfying path
    - `E_res`: edges appearing on ≥1 satisfying path
    - `tags`: per-step name → set of ids bound at that step
    - `paths` (optional debug output): list of bindings `{name -> node/edge id}`
- Null/NaN semantics:
  - Any comparison touching null/NaN evaluates to False
  - No implicit type coercion
  - Null/NaN never flow into downstream bitsets/state tables
- Note: Dedicated **path predicates** (`length(p)`, `distinctNodes(p)`, etc.) and `MATCH PATH` syntax are out of scope; they will consume this enumerator later, but this issue does not define any new path operators.

## Key Tasks
- Iterative dynamic program:
  - For each step, join on src/dst and apply node/edge predicates
  - Track bindings for named steps so WHERE clauses can reference multiple hops
  - As soon as all names in a WHERE conjunct are bound, evaluate it and prune early
- CI safety caps: |V|≤12, |E|≤40, length≤5, abort if partial rows exceed 200k
- Property tests (Hypothesis): random tiny graphs, random WHERE atoms (<, <=, >, >=, ==, !=), graph shapes (cycles, multi-edges, repeated nodes, mixed directions), explicit null/NaN cases
- Golden fixtures for regression and documentation

## Acceptance
- pandas backend must match the enumerator on all CI inputs
- Future cuDF backend must match the same oracle input-for-input
- Null/NaN behavior is identical between oracle and engines

## References
- GraphFrames motif/path-table semantics as a conceptual baseline (path tables restricted to fixed-length chains for oracle use)

lmeyerov:
## Summary
- add pandas-only GFQL reference enumerator with same-path predicate support + caps
- cover deterministic + Hypothesis tests and parity checks with existing pandas GFQL engine
- document oracle contract and update changelog

## Testing
- python -m pytest graphistry/tests/gfql
- ./bin/mypy.sh graphistry/gfql/ref/enumerator.py
- ruff check graphistry/gfql/ref/enumerator.py graphistry/tests/gfql

Resolves #836

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ dev_extras = {
       'sphinx-copybutton==0.5.2',
       'sphinx-book-theme==1.1.3',
     ],
-    'test': ['flake8>=5.0', 'mock', 'mypy', 'pytest', 'pytest-xdist'] + stubs + test_workarounds,
+    'test': ['flake8>=5.0', 'hypothesis', 'mock', 'mypy', 'pytest', 'pytest-xdist'] + stubs + test_workarounds,
     'testai': [
       'numba>=0.57.1'  # https://github.com/numba/numba/issues/8615
     ],


ai/docs/gfql/oracle.md
lmeyerov, line 1: move to ai/code_notes/docs

Issue #844, opened 2025-11-18
UMAP fails on CuPy 13 / CUDA 13 because cupyx coo_matrix no longer has .get()


lmeyerov:
## Summary
Upgrading our Forge ETL stack to CUDA 13 (RAPIDS 25.10, graphistry==0.45.7) surfaced a regression where every GFQL UMAP call now throws:

```
GFQLTypeError: [invalid-node-reference] Error executing 'umap': 'coo_matrix' object has no attribute 'get'
```

`graphistry.umap_utils.umap_graph_to_weighted_edges()` still calls `coo.get()` when the engine is `"cuml"` (line 186 in v0.45.7). cupyx.scipy.sparse.coo_matrix dropped `.get()` in CuPy 13 (the version bundled with RAPIDS 25.10 / CUDA 13), so any code path that hits UMAP now crashes before weights are pruned.

## Repro
1. Use a CUDA 13 / RAPIDS 25.10 environment (e.g., graphistry/graphistry-nvidia:v2.45.3-13.0) where graphistry==0.45.7, cupy==13.0.*, cuml==25.10.*
2. Run any GFQL operation that invokes UMAP, e.g.
   ```bash
   pytest test/server/client/processing/test_gfql_endpoint.py::test_gfql_umap_chain_should_work_when_pygraphistry_761_fixed --maxfail=1
   ```
3. Observe `GFQLTypeError … 'coo_matrix' object has no attribute 'get'` originating from `umap_graph_to_weighted_edges()`.

Minimal python snippet (runs on CUDA 13):
```python
import cupy as cp
from cupyx.scipy.sparse import coo_matrix
from graphistry.umap_utils import umap_graph_to_weighted_edges

coo = coo_matrix((cp.array([1, 2]), (cp.array([0, 1]), cp.array([1, 0]))), shape=(2, 2))
umap_graph_to_weighted_edges(coo, engine="cuml", is_legacy=False)
```
=> AttributeError because cupyx removed `.get()`.

## Expected
`umap_graph_to_weighted_edges()` should convert the device coo matrix to host without using `.get()` (e.g., via `cupy.asnumpy`/`coo_matrix.toarray()` or `coo.todense()` accessors). After the fix, GFQL UMAP calls should succeed again on RAPIDS 25.10 / CUDA 13.

Let me know if additional logs are needed.

PR #845, opened 2025-11-19
fix(umap): support cupy13 weighted edges

lmeyerov:
## Summary
- Fix CuPy 13 UMAP COO conversion by using cupy.asnumpy() with legacy .get() fallback to keep RAPIDS 25.10 and older stacks working (#844)
- Export core submodules (compute/render/arrow_uploader/pygraphistry) from top-level package to stabilize import order in core tests
- Documented changes in CHANGELOG under [Development]

## Testing
- CI push: CI Tests (feature/umap-cuda13) ✅ (run 19788560877)
- CI PR: CI Tests ⏳ in progress (latest run 19788922665)


lmeyerov:
Status update:\n- CuPy 13 regression is fixed:  now converts device COO arrays via  (with  fallback), so CUDA13 / RAPIDS 25.10 runs pass the new regression test and a real RAPIDS smoke script.\n- Added three focused tests () that cover the CuPy-13 path, legacy  path, and the  pandas path. All pass locally and in containerized GPU/CPU runs.\n- CHANGELOG entry added under .\n- GPU harness (CONFIG
PREP) still fails before running tests because it attempts to import cache , which isn’t published yet.\n- Legacy docker images (, ) exist, but pulls keep failing on the final 5GB layer (). Once the cache image or alternative mirror is available, I’ll rerun the full GPU suite and legacy CUDA regression.\nLet me know if you’d like additional validation before re-running CI.

PR #849, opened 2025-11-19
fix(layout): preserve user-set play URL parameter in ring layouts

dess890:
Fixes issue where &play option does not update in URL param

### DIFF
--- a/graphistry/tests/layout/ring/test_continuous.py
+++ b/graphistry/tests/layout/ring/test_continuous.py
@@ -140,3 +140,46 @@ class Test_continuous_ring(NoAuthTestCase):
         for i, row in enumerate(g._complex_encodings['node_encodings']['default']['pointAxisEncoding']['rows']):
             assert row['r'] == 500 + 100 * i
             assert row['label'] == str(2 + 2 * i)
+
+    def test_play_ms_preserves_url_param(self):
+        lg = LGFull()
+        g = (
+            lg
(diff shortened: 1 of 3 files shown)


graphistry/layout/ring/categorical.py
```diff
@@ -250,13 +250,20 @@ def unique() -> List[Any]:
 
     #print('axis', axis)
 
+    if play_ms != 0:
+        play_value = play_ms
+    elif 'play' in g._url_params:
+        play_value = g._url_params['play']
```
aucahuasi: Can we parse the play value or make sure it has a valid format?

aucahuasi (changes requested):
Thanks, can you add some tests as well?

graphistry/layout/ring/categorical.py
```diff
@@ -250,13 +250,23 @@ def unique() -> List[Any]:
 
     #print('axis', axis)
 
+    if play_ms != 0:
+        play_value = play_ms
+    elif 'play' in g._url_params:
```
lmeyerov, line 255: This looks wrong - if 0 is set, we should respect it

I can imagine changing the type signature to `: Optional[int] = 0`, and having `None` mean "honor url param if already set, otherwise default to 0", but that feels unnecessarily complicated

lmeyerov:
See comment for cleaner way to support cascading - do when None

Also I thought we have 3-4 ring layout variants ?

PR #865, opened 2025-12-06
feat(api)!: remove legacy API v1/VGraph support

aucahuasi:
BREAKING CHANGE: Remove legacy VGraph/protobuf API v1 in favor of Arrow API v3.

Removed:
- `_etl1()`, `_etl_url()`, `_check_url()` methods from pygraphistry.py
- API v1 dispatch path from PlotterBase.py
- `TestPlotterBindings_API_1` and `TestPlotterStylesJSON` test classes

Changed:
- `register(api=...)` type from `Literal[1, 3]` to `Literal[3]`
- `client_session.py` type from `Literal["arrow", "vgraph"]` to `Literal["arrow"]`
- Updated tests

Migration: Users calling `graphistry.register(api=1)` must switch to `graphistry.register(api=3)` or omit the parameter (defaults to v3).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Development]
 <!-- Do Not Erase This Section - Used for tracking unreleased changes -->
 
-## [0.46.2 - 2025-12-15]
+## [0.47.0 - 2025-12-15]
 
 ### Breaking 🔥
 - **API v1 Removal**: Removed legacy VGraph/protobuf API v1 support in favor of API v3.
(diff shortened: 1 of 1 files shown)


graphistry/client_session.py
```diff
@@ -55,9 +55,9 @@ def __init__(self) -> None:
 
         env_api_version = get_from_env("GRAPHISTRY_API_VERSION", int)
         if env_api_version is None:
-            env_api_version = 1
-        elif env_api_version not in [1, 3]:
-            raise ValueError("Expected API version to be 1, 3, instead got (likely from API_VERSION): %s" % env_api_version)
+            env_api_version = 3
+        elif env_api_version != 3:
+            raise ValueError("Expected API version to be 3 (Arrow format). Legacy API versions 1 and 2 are no longer supported. Got: %s" % env_api_version)
```
lmeyerov: Api equal 3. Supports arrow, parquet, CSV, json and orc via file api. Basically anything that cudf takes. I do think we do happen to do arrow uploads for DFs.

aucahuasi: thanks, done!

aucahuasi:
> * does this default to API=3, or registers needs to  be chang to do that too?

Yes, api=3 is defined in `ClientSession.__init__()`

> * If someone calls plot without register, does that still error well?

Yes, with a clear `Exception("Not logged in")`

> * I noticed a version checking code is removed, is that truly legacy?

Yes, that was for `api=1` (aka only used for legacy)

> * any API=1/2 calls left in docs/ipynb's?

Done, I updated this, see commit https://github.com/graphistry/pygraphistry/pull/865/commits/f2524e3193a8b37fd8d10bacc3726d56b0281590

PR #874, opened 2025-12-29
feat: add collections validation and GFQL support

lmeyerov:
## Summary
- Add collections validator with strict/autofix behavior and GFQL wire normalization
- Expose collections API with validate/warn and plot-time URL param validation
- Add collections tests and clarify plan template location
- Move collections helpers to `graphistry/collections.py` and introduce typed models in `graphistry/models/collections.py`

### Key Features
- `g.collections(...)` API for defining subsets via GFQL expressions with priority-based visual encodings
- Helper constructors `graphistry.collection_set(...)` and `graphistry.collection_intersection(...)`
- Support for `showCollections`, `collectionsGlobalNodeColor`, and `collectionsGlobalEdgeColor` URL params
- Automatic JSON encoding and GFQL AST/Chain/wire-protocol normalization

### Validation Behavior
- **Strict mode**: Raises `ValueError` on invalid input
- **Autofix mode** (default): Drops invalid collections with warnings
- Set collections require an `id` field (server requirement) - missing IDs are warned/dropped, not auto-generated
- Intersection collections cross-validate that referenced set IDs exist
- GFQL parsing uses `_wrap_gfql_expr` as canonical implementation with precise exception handling

## Testing
```bash
python -m pytest graphistry/tests/test_collections.py graphistry/tests/test_dataset_id_invalidation.py
./bin/lint.sh
./bin/mypy.sh
```

## Review Comments Addressed
See [comment](https://github.com/graphistry/pygraphistry/pull/874#issuecomment-3866087468) for detailed breakdown of all review comments and their resolutions.


graphistry/PlotterBase.py
```diff
@@ -2191,7 +2244,10 @@ def plot(
             'viztoken': str(uuid.uuid4())
         }
 
-        viz_url = self._pygraphistry._viz_url(info, self._url_params)
+        from graphistry.validate.validate_collections import normalize_collections_url_params
+
+        url_params = normalize_collections_url_params(self._url_params, validate=validate_mode, warn=warn)
+        viz_url = self._pygraphistry._viz_url(info, url_params)
```
lmeyerov, line 2293: back out

graphistry/PlotterBase.py
```diff
@@ -1846,6 +1847,58 @@ def settings(self, height=None, url_params={}, render=None):
         return res
 
 
+    def collections(
+        self,
+        collections: Optional[CollectionsInput] = None,
+        show_collections: Optional[bool] = None,
+        collections_global_node_color: Optional[str] = None,
+        collections_global_edge_color: Optional[str] = None,
+        encode: bool = True,
```
mj3cheun: in the interest of simplifying, would it be difficult to detect pre-encoded strings?

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,410 @@
...
+        return None
+    return {'type': 'intersection', 'sets': sets_list}
+
+
+def normalize_collections(
+    collections: CollectionsInput,
+    validate: ValidationParam = 'autofix',
+    warn: bool = True
+) -> List[Dict[str, Any]]:
+    validate_mode, warn = normalize_validation_params(validate, warn)
```
mj3cheun, line 236: nitpick: this seems to already be called everywhere that uses `normalize_collections`, can either remove this extra call here or remove the `normalize_validation_params` everywhere else

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,410 @@
...
+
+def _parse_collections_input(
+    collections: CollectionsInput,
+    validate_mode: ValidationMode,
+    warn: bool
+) -> List[Dict[str, Any]]:
+    if isinstance(collections, list):
+        return _coerce_collection_list(collections, validate_mode, warn)
+    if isinstance(collections, dict):
+        return _coerce_collection_list(collections, validate_mode, warn)
```
mj3cheun: nitpick: would just get rid of the `_coerce_collection_list` function and put the contents in here, would get rid of checking is list is dict etc twice and make easier to read

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,410 @@
...
+        collection_type = normalized_entry.get('type', 'set')
+        if not isinstance(collection_type, str):
+            _issue(
+                'Collection type must be a string',
+                {'index': idx, 'value': collection_type, 'type': type(collection_type).__name__},
+                validate_mode,
+                warn
+            )
+            if validate_mode == 'autofix':
+                collection_type = str(collection_type)
```
mj3cheun: if collection_type was not already a string, im not sure that just typecasting it is going to produce a valid type string given this string must be either `set` or `intersection`

mj3cheun: actually nevermind, just saw the `if collection_type not in ('set', 'intersection'):` below, im thinking maybe combine this check with the check below, if `collection_type` is not string its all but guaranteed to fail and hit `continue` in the section below

graphistry/PlotterBase.py
lmeyerov: yeah this surprised me: the reason it's like this is bc the base REST api takes AABBCC instead of ~css colors like we started to do elsewhere i believe (js `color(xyz)` normalizes to `rgba('...')`, handling `'#AABBCC'`, `'silver'`, etc) 

so maybe fix is to ship better colors in REST, and then this upgrades to that?

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,410 @@
...
+            if 'gfql' in raw:
+                return raw.get('gfql')
+            if 'chain' in raw:
+                return raw.get('chain')
+            return raw
+        if isinstance(raw, list):
+            return raw
+        return raw
+
+    return _normalize_ops_list(_extract_ops_value(gfql_ops))
```
mj3cheun: i feel this could be simplified by just feeding the GFQL into graphistry.compute.chain and seeing if it parses properly rather than setting up stuff here to determine if GFQL is valid or not

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,410 @@
...
+        return None
+    expr_type = expr.get('type', 'intersection')
+    if expr_type != 'intersection':
+        _issue(
+            'Intersection expr type must be "intersection"',
+            {'index': entry_index, 'value': expr_type},
+            validate_mode,
+            warn
+        )
+        return None
```
mj3cheun, line 220: this function is only called if type is intersection, so this if statement will never return true

graphistry/PlotterBase.py
mj3cheun: i think thats a great idea to ship better colours in REST will add that to the list

i was talking about the `encode: bool = True,` though, which is `If True, JSON-minify and URL-encode collections. Use False for pre-encoded strings.`

i was thinking we could just detect pre-encoded strings (which i assume are URL encoded and can be detected by attempting a parse). then we can remove this parameter

lmeyerov:
the issue with strict true/false is that's closer to what we had before, and users were complaining that they just wanted it to 'work', hence autofix (coerce++). validate=true/false is closer to what you're thinking, while 'autofix' (coerce) is "it'll run, but may not be what you want, but you said wanted soemthing that runs"

We therefore have leeway in what autofix does --- we just  need to warn (if warn=auto/true), and do what. So the q is... what should it do wrt diff collections errors, if strong opinions about any params in any direction?

My default intuition is probably:
* drop collections with invalid gfql
* colors: random-but-deterministic? or neutral grey / transparent? (default black looks buggy)
* others:  ??? disable / see if there's default values ?

mj3cheun:
agree with that, i think we do the 3 bullets listed if strict (or whatever we want to call it) is false and its almost what we are doing right now. theres only 1 point i would make about the current implementation
- instead of just trying to typecast stuff (which a lot of the time doesnt work), i would prefer we do default values or in the case of colours, random colours

in summary its only the approach i think might need to change, not the intention

EDIT: if we do the above it actually gets closer to a true autofix than what it was before, so i dont have an issue with the name anymore

mj3cheun (approved):

graphistry/collections.py
```diff
@@ -0,0 +1,113 @@
...
+    value = metadata.get("node_color")
+    if value is not None:
+        collection["node_color"] = value
+    value = metadata.get("edge_color")
+    if value is not None:
+        collection["edge_color"] = value
+    return collection
+
+
+def _wrap_gfql_expr(expr: CollectionExprInput) -> Dict[str, JSONVal]:
```
aucahuasi: this function still seems to reimplement normalizations (see  [this](https://github.com/graphistry/pygraphistry/pull/874#discussion_r2683617536) and `return [_normalize_op`)

Why we don't just use `validate_collections` (see `_normalize_gfql_ops`)?

Both functions should produce output compatible with the backend's `Chain.from_json({'chain': ops})` (see `compute_collections.py:46`)

I wonder if we can use Chain.from_json() and catch parse errors (like Manfred suggest)

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,420 @@
...
+    sets_value = expr.get('sets', expr.get('intersection'))
+    if sets_value is None:
+        _issue('Intersection expr missing "sets"', {'index': entry_index}, validate_mode, warn)
+        return None
+    sets_list = _normalize_sets_list(sets_value, validate_mode, warn, entry_index)
```
aucahuasi, line 225: I think we should validate `len(sets_list) > 0` too. See backend code `compute_collections.py:186` , there we try to access `sets[0]` causing `IndexError`.

The frontend UI enforces `>= 2 sets`. Maybe we should add `if len(sets_list) == 0: _issue(...); return None` here, that would be consistent with the empty GFQL chain check [at line 194](https://github.com/graphistry/pygraphistry/pull/874/changes#diff-e880ecb86dd5ed8e3fba3a9962e5606cd8ae4c13b32ab686ae39f1ec40bccaa3R194).

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,420 @@
...
+            if 'chain' in raw:
+                return raw.get('chain')
+            return raw
+        if isinstance(raw, list):
+            return raw
+        return raw
+
+    return _normalize_ops_list(_extract_ops_value(gfql_ops))
+
+    _issue(
```
aucahuasi: Dead code here, because re already invoked `return`.

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,420 @@
...
+                'Collection type must be "set" or "intersection"',
+                {'index': idx, 'value': collection_type},
+                validate_mode,
+                warn
+            )
+            if validate_mode == 'autofix':
+                continue
+            return []
+
+        for field in ('id', 'name', 'description'):
```
aucahuasi, line 293: Can you confirm whether the backend requires `set['id']` at `compute_collections.py:166`, if so it may raises `KeyError`.

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,420 @@
...
+            _normalize_str_field(normalized_entry, field, validate_mode, warn, idx, autofix_drop=False)
+        for field in ('node_color', 'edge_color'):
+            _normalize_str_field(normalized_entry, field, validate_mode, warn, idx, autofix_drop=True)
```
aucahuasi, line 296: Can we check wheter we need or not cross-validation of intersection set IDs here?: After the `for idx, entry` loop finishes building normalized, there's no check that intersection `expr.sets` IDs reference actual collection `id` fields in the same list. A dangling reference (e.g., `sets: ["nonexistent"]`) passes all validation but causes the backend (see `compute_collections.py:194`) to raise `"Infinite loop detected while computing sets"`.

Maybe we can add a post-loop pass: collect all set IDs, then verify each intersection's sets entries exist. In `strict` mode raise; in `autofix` drop the intersection.

aucahuasi (dismissed):
Thanks for this PR, I left some comments and I think we need to solve this feedback before merge:
https://github.com/graphistry/pygraphistry/pull/874/changes#r2778233409

graphistry/validate/validate_collections.py
lmeyerov: handled on  :307

graphistry/validate/validate_collections.py
lmeyerov: there can be mismatch between the intersection container and the intersection expr

graphistry/validate/validate_collections.py
lmeyerov: done see `_normalize_sets_list`

graphistry/validate/validate_collections.py
lmeyerov: (keeping to avoid reasoning on caller side)

graphistry/validate/validate_collections.py
lmeyerov: correct

Now validates and in autofix mode auto-generates (`set-0`, `intersection-1`) with warning, so collections work rather than crashing the server.

graphistry/validate/validate_collections.py
lmeyerov: tx, done in _validate_intersection_references

graphistry/collections.py
lmeyerov: ok factored a bit more, tx

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,428 @@
...
+    Validate that intersection set IDs reference actual collection IDs.
+
+    Dangling references (e.g., sets: ["nonexistent"]) cause backend errors.
+    In strict mode, raise on first issue. In autofix mode, drop invalid intersections.
+    """
+    # Collect all set IDs (only from 'set' type collections, not intersections)
+    set_ids = {
+        c.get('id')
+        for c in collections
+        if c.get('type') == 'set' and c.get('id')
```
aucahuasi: This filter only collects IDs from `type == 'set'`, but the backend (`compute_all_sets` in `compute_collections.py:162-197`) supports intersection-to-intersection references via a deferred multi-pass loop: an intersection whose deps aren't resolved yet is re-queued until they are. Also, the frontend UI also allows it (the checkbox list at `collections/index.js:239` iterates all `setsArray` entries, not just `sets`). So the missing check at `line 358` will incorrectly reject any intersection that references another intersection, even though it's valid end-to-end.

The simplest fix is `all_ids = {c.get('id') for c in collections if c.get('id')}`, but that would let **circular references** through (`A->B->A`), which the backend catches with its `"Infinite loop detected"` guard at `line 194`. A **proper fix mirrors the backend's resolution**: start with set IDs as resolved_ids, then iteratively mark intersections as resolved when all their sets refs are in `resolved_ids` (anything left after no more progress is either dangling or cyclic). That's ~12 lines and exactly maps to the `compute_all_sets` while-loop.

## Deep Analysis: Intersection-to-Intersection References

### The Backend's Deferred Processing Model

`compute_all_sets` (`compute_collections.py:155-197`) implements a **multi-pass topological resolution loop**:

```
Pass 1: Process what you can, defer what you can't
Pass 2: Process newly-unblocked items, defer remaining
...
Pass N: If nothing new resolved -> "Infinite loop detected"
```

The key mechanism (lines 162-197):

```python
while len(sets_list):                        # keep going until everything resolved
    for set in sets_list:
        match set['type']:
            case 'set':
                subgraphs_dict[set_id] = ...     # resolved immediately
            case 'intersection':
                if not all(id in subgraphs_dict for id in set['expr']['sets']):
                    sets_to_process_list.append(set)   # DEFER -- deps not ready
                    continue
                # ... compute intersection ...
                subgraphs_dict[set_id] = g_a         # resolved

    if len(sets_to_process_list) == len(sets_list):   # nothing resolved this pass
        raise Exception("Infinite loop detected")
    sets_list = sets_to_process_list                   # retry deferred items
    sets_to_process_list = []
```

**What `subgraphs_dict` accumulates:** Both `'set'` and `'intersection'` results. An intersection resolved in pass 1 becomes available for other intersections in pass 2. The backend makes **no distinction** between set-type and intersection-type entries in `subgraphs_dict` -- any resolved ID can be referenced.

### What the Frontend UI Allows

`index.js:239` -- the intersection creation modal shows **all** items in `setsArray`:

```javascript
{setsArray.map((set) => (
    <label key={set.id}>
        <input type="checkbox"
            checked={selectedSets.includes(set.id)}
            onChange={...}
        />
        <span>{set.name || `Set ${set.id}`}</span>
    </label>
))}
```

`setsArray` at line 20 is built from `sets` which includes **all collection types** -- sets, intersections, and global. The UI does not filter by type. A user can select an existing intersection when creating a new intersection. This is an intentional UX choice.

### What pygraphistry Blocks

`validate_collections.py:346-351`:

```python
set_ids = {
    c.get('id')
    for c in collections
    if c.get('type') == 'set' and c.get('id')   # <-- ONLY type='set'
}
```

Then at line 358:

```python
missing = [sid for sid in referenced_sets if sid not in set_ids]
```

Any intersection referencing another intersection will always have `missing` entries because intersection IDs are never in `set_ids`.

### Concrete Example -- Valid on Backend, Rejected by Client

```python
import graphistry

collections = [
    graphistry.collection_set(
        expr=graphistry.n({"department": "engineering"}),
        id="engineers"
    ),
    graphistry.collection_set(
        expr=graphistry.n({"level": "senior"}),
        id="seniors"
    ),
    # Intersection of the two sets -- works everywhere
    graphistry.collection_intersection(
        sets=["engineers", "seniors"],
        id="senior-engineers"
    ),
    # Intersection referencing another intersection -- BLOCKED by client
    graphistry.collection_intersection(
        sets=["senior-engineers", "engineers"],
        id="also-senior-engineers",  # equivalent to senior-engineers, but valid topology
    ),
]

# In autofix mode: "also-senior-engineers" silently dropped
# In strict mode: ValueError raised
g = graphistry.bind().collections(collections=collections)
```

The backend would resolve this in 2 passes:

- **Pass 1:** `engineers` (ok), `seniors` (ok), `senior-engineers` (ok -- deps ready), `also-senior-engineers` (deferred -- dep `senior-engineers` just resolved)
- **Pass 2:** `also-senior-engineers` (ok -- deps now in `subgraphs_dict`)

### Is the Naive Fix (`all_ids`) Sufficient?

The suggested fix:

```python
all_ids = {c.get('id') for c in collections if c.get('id')}
```

This would pass all collections with existing IDs through, which handles the simple case. But there are edge cases:

**Circular references:** Two intersections referencing each other:

```python
collection_intersection(sets=["B"], id="A"),
collection_intersection(sets=["A"], id="B"),
```

With the naive fix, both `"A"` and `"B"` are in `all_ids`, so both pass validation. But the backend's infinite loop detection (line 194-195) would catch this and raise `"Infinite loop detected"`. The client-side validation wouldn't prevent it.

**Self-references:** An intersection referencing itself:

```python
collection_intersection(sets=["self"], id="self"),
```

Same problem -- passes the naive `all_ids` check, crashes on backend.

### The Right Fix

The proper approach mirrors the backend's topological resolution. Since the backend resolves in order and defers unsatisfied dependencies, the client should verify that a valid resolution order exists -- i.e., the dependency graph is a DAG (no cycles):

1. Collect all IDs from all collection types
2. For each intersection, check that its `sets` references are either:
   - A `type='set'` collection (always resolvable), or
   - An intersection that appears before it in topological order
3. Reject cycles (A->B->A)

A simpler pragmatic approach that covers 99% of cases without full topo-sort:

```python
# Build resolvable set progressively (mirrors backend's multi-pass)
resolved_ids = {c.get('id') for c in collections if c.get('type') == 'set' and c.get('id')}
remaining = [c for c in collections if c.get('type') == 'intersection']
changed = True
while changed:
    changed = False
    still_remaining = []
    for c in remaining:
        refs = c.get('expr', {}).get('sets', [])
        if all(r in resolved_ids for r in refs):
            resolved_ids.add(c.get('id'))
            changed = True
        else:
            still_remaining.append(c)
    remaining = still_remaining
# anything left in 'remaining' has unresolvable references (dangling or cyclic)
```

This exactly mirrors `compute_all_sets`'s while-loop logic and would correctly handle:

- Intersection-to-intersection chains
- Dangling references
- Circular references

### Tests

This new tests (`test_intersection_refs.py`) can help to reproduce and understand the issues and proposed fixes:

```python
# -*- coding: utf-8 -*-
"""
Reproducer for: _validate_intersection_references rejects valid
intersection-to-intersection references.

The backend (compute_all_sets in compute_collections.py:162-197) resolves
intersections via a deferred multi-pass loop — an intersection whose deps
aren't ready yet is re-queued until they are.  The frontend UI also allows
selecting existing intersections when creating a new one.

But _validate_intersection_references only collects IDs from type='set',
so any intersection referencing another intersection is treated as a
dangling reference and dropped (autofix) or rejected (strict).
"""

import json
import pytest
from urllib.parse import unquote

import graphistry
from graphistry.collections import collection_intersection, collection_set
from graphistry.validate.validate_collections import normalize_collections


def decode(encoded: str):
    return json.loads(unquote(encoded))


def params(collections, **kwargs):
    return graphistry.bind().collections(collections=collections, **kwargs)._url_params


# ── The core issue: intersection referencing another intersection ──────────


def test_intersection_referencing_intersection_strict():
    """Backend supports intersection→intersection refs; strict mode should accept them."""
    collections = [
        collection_set(expr=graphistry.n({"dept": "eng"}), id="engineers"),
        collection_set(expr=graphistry.n({"level": "senior"}), id="seniors"),
        collection_intersection(sets=["engineers", "seniors"], id="senior-eng"),
        # This references an intersection, not a set — valid on backend
        collection_intersection(sets=["senior-eng", "engineers"], id="senior-eng-v2"),
    ]
    # Should NOT raise — all references are resolvable
    url = params(collections, validate="strict")
    decoded = decode(url["collections"])
    ids = [c["id"] for c in decoded]
    assert "senior-eng-v2" in ids, (
        "intersection→intersection ref was rejected in strict mode but backend supports it"
    )


def test_intersection_referencing_intersection_autofix():
    """In autofix mode the intersection→intersection ref should be kept, not dropped."""
    collections = [
        collection_set(expr=graphistry.n({"dept": "eng"}), id="engineers"),
        collection_intersection(sets=["engineers"], id="int-1"),
        collection_intersection(sets=["engineers", "int-1"], id="int-2"),
    ]
    url = params(collections, validate="autofix", warn=False)
    decoded = decode(url["collections"])
    ids = [c["id"] for c in decoded]
    assert "int-2" in ids, (
        "intersection→intersection ref was silently dropped in autofix but backend supports it"
    )


def test_chained_intersections():
    """A chain A→B→C of intersections should all resolve (backend does multi-pass)."""
    collections = [
        collection_set(expr=graphistry.n({"a": 1}), id="base"),
        collection_intersection(sets=["base"], id="step-1"),
        collection_intersection(sets=["step-1"], id="step-2"),
        collection_intersection(sets=["step-2"], id="step-3"),
    ]
    url = params(collections, validate="strict")
    decoded = decode(url["collections"])
    ids = [c["id"] for c in decoded]
    assert ids == ["base", "step-1", "step-2", "step-3"], (
        f"Chained intersections were not all preserved: {ids}"
    )


# ── Dangling refs: correctly rejected today, should stay rejected after fix ──


def test_dangling_intersection_ref_strict_raises():
    """Intersection referencing a nonexistent ID should raise in strict."""
    collections = [
        collection_set(expr=graphistry.n({"a": 1}), id="base"),
        collection_intersection(sets=["base", "nonexistent"], id="bad"),
    ]
    with pytest.raises(ValueError):
        params(collections, validate="strict")


def test_dangling_intersection_ref_autofix_drops():
    """Intersection referencing a nonexistent ID should be dropped in autofix."""
    collections = [
        collection_set(expr=graphistry.n({"a": 1}), id="base"),
        collection_intersection(sets=["nonexistent"], id="bad"),
    ]
    with pytest.warns(RuntimeWarning):
        url = params(collections, validate="autofix", warn=True)
    decoded = decode(url["collections"])
    ids = [c["id"] for c in decoded]
    assert "bad" not in ids


# ── Circular/self refs: pass today ONLY because _validate_intersection_references
# never collects intersection IDs (line 347-351 filters on type=='set').
# After fixing the intersection→intersection issue above, a naive fix like
# `all_ids = {c.get('id') for c in collections if c.get('id')}` would make
# these two tests FAIL because "int-a"/"int-b"/"self" would all be in the
# allowed set with no cycle detection.  The proper fix (multi-pass resolution
# mirroring compute_all_sets) would keep these passing. ─────────────────────


def test_circular_intersection_refs_strict_raises():
    """Circular refs (A->B->A) should raise — backend hits 'Infinite loop detected'."""
    collections = [
        collection_set(expr=graphistry.n({"a": 1}), id="base"),
        collection_intersection(sets=["int-b"], id="int-a"),
        collection_intersection(sets=["int-a"], id="int-b"),
    ]
    with pytest.raises(ValueError):
        params(collections, validate="strict")


def test_self_referencing_intersection_strict_raises():
    """Self-reference (A->A) should raise — backend hits 'Infinite loop detected'."""
    collections = [
        collection_set(expr=graphistry.n({"a": 1}), id="base"),
        collection_intersection(sets=["self"], id="self"),
    ]
    with pytest.raises(ValueError):
        params(collections, validate="strict")

```

graphistry/validate/validate_collections.py
```diff
@@ -0,0 +1,428 @@
...
+            gfql_ops = json.loads(gfql_ops)
+        except json.JSONDecodeError as exc:
+            _issue('GFQL chain string must be JSON', {'index': entry_index, 'error': str(exc)}, validate_mode, warn)
+            return None
+
+    # Use canonical implementation from compute/ast.py
+    try:
+        from graphistry.compute.ast import normalize_gfql_to_wire
+        return normalize_gfql_to_wire(gfql_ops)
+    except (TypeError, ValueError, GFQLValidationError, GFQLSyntaxError) as exc:
```
aucahuasi: The except clause here catches (`TypeError`, `ValueError`, `GFQLValidationError`, `GFQLSyntaxError`) but misses `AssertionError`. The call chain `normalize_gfql_to_wire -> from_json -> class-level from_json` hits bare assert statements in `ASTLet.from_json` (see `ast.py:1040`), `ASTRemoteGraph.from_json` (see `ast.py:1157`), `ASTRef.from_json` (see `ast.py:1273-1274`), `ASTCall.from_json` (see `ast.py:1415`), and `predicates/from_json.py:34-36`. Any malformed GFQL dict with a recognized type but a missing required field (e.g. `{"type": "Let"}` with no bindings) raises `AssertionError` that flies past this handler, crashing both strict and autofix modes with a raw assertion instead of routing through _issue().

Since this function is the validation boundary for user-provided GFQL input, I'd suggest either adding `AssertionError` to the tuple or broadening to except Exception: this is a pure data-transformation path with no I/O, so catching broadly here is safe and future-proofs against new AST classes that might raise other types. I put together a link-to-gist-or-file covering all five assert sites (`Let`, `RemoteGraph`, `Ref`, `Call`, `predicates`) in both modes if that helps verify the fix.

## The AssertionError Leak -- Full Trace

### The validation boundary

`_normalize_gfql_ops` (`validate_collections.py:149-185`) is the **validation boundary** for GFQL expressions inside collections. Its job is: take any user-provided GFQL input, try to normalize it, and if anything goes wrong, route the error through `_issue()` so it respects the validation mode (`strict` -> raise `ValueError`, `autofix` -> warn + drop).

The try/except at line 174-185 is the gate:

```python
try:
    from graphistry.compute.ast import normalize_gfql_to_wire
    return normalize_gfql_to_wire(gfql_ops)
except (TypeError, ValueError, GFQLValidationError, GFQLSyntaxError) as exc:
    _issue('Invalid GFQL operation in collection', ...)
    return None
```

This catches 4 exception types. The assumption is: "any error `normalize_gfql_to_wire` can raise is one of these four." That assumption is wrong.

### The call chain that breaks it

Here's the full path, step by step:

**Step 1** -- `normalize_gfql_to_wire` (`ast.py:1511`) receives the raw expression. For a dict input, it calls `_normalize_op`:

```python
def _normalize_op(op):
    if isinstance(op, dict):
        return from_json(op, validate=True).to_json()   # <-- step 2
```

**Step 2** -- The top-level `from_json` dispatcher (`ast.py:1453`) inspects `op['type']` and dispatches to the right class. This function itself is well-behaved -- it raises `GFQLSyntaxError` for missing `type`, unknown `type`, etc. But it delegates the actual parsing:

```python
elif o['type'] == 'Let':
    out = ASTLet.from_json(o, validate=validate)    # <-- step 3
```

**Step 3** -- The class-level `from_json` methods. These are the ones that use **bare `assert`**:

| Class | Line | Code | Triggers when |
|---|---|---|---|
| `ASTLet` | `ast.py:1040` | `assert 'bindings' in d` | `{"type": "Let"}` (no bindings key) |
| `ASTRemoteGraph` | `ast.py:1157` | `assert 'dataset_id' in d` | `{"type": "RemoteGraph"}` (no dataset_id) |
| `ASTRef` | `ast.py:1273-1274` | `assert 'ref' in d` / `assert 'chain' in d` | `{"type": "ChainRef"}` (missing ref or chain) |
| `ASTCall` | `ast.py:1415` | `assert 'function' in d` | `{"type": "Call"}` (no function key) |

And there's an entire additional file -- `predicates/from_json.py:33-42` -- that is **nothing but asserts**:

```python
def from_json(d):
    assert isinstance(d, dict)          # line 34
    assert 'type' in d                  # line 35
    assert d['type'] in type_to_predicate  # line 36
    assert isinstance(d['type'], str)   # line 37
    pred = type_to_predicate[d['type']]
    out = pred.from_json(d)
    assert isinstance(out, ASTPredicate)  # line 40
    out.validate()
    return out
```

This gets called when an `ASTNode` or `ASTEdge` has a `filter` predicate that needs to be deserialized.

### Why AssertionError is not caught

`assert` raises `AssertionError`, which inherits directly from `Exception`:

```
BaseException
  +-- Exception
        +-- ValueError
        +-- TypeError
        +-- AssertionError   <-- HERE
        +-- GFQLValidationError
        +-- GFQLSyntaxError
```

`AssertionError` is **not** a subclass of any of the four caught types. It flies straight past the `except` clause.

### What happens when it leaks

The unhandled `AssertionError` propagates up the call stack:

```
_normalize_gfql_ops      (validate_collections.py:176)
  +-- normalize_gfql_to_wire  (ast.py:1538)
        +-- _normalize_op     (ast.py:1538)
              +-- from_json   (ast.py:1488)
                    +-- ASTLet.from_json  (ast.py:1040)
                          +-- assert 'bindings' in d  --> AssertionError!
```

Back in `_normalize_gfql_ops`, the `except` clause doesn't match, so the `AssertionError` **bypasses `_issue()` entirely** and propagates as a raw unhandled exception.

This means:

- **In `strict` mode**: User gets a raw `AssertionError: Let missing bindings` instead of a proper `ValueError` with structured context (`{'index': 2, 'error': '...'}`)
- **In `autofix` mode**: The collection isn't gracefully dropped. Instead the whole `.collections()` call crashes with `AssertionError`. The autofix contract ("we fix what we can, warn, and continue") is broken.

### Concrete reproducer

```python
import graphistry
from graphistry.collections import collection_set

# A "Let" operation missing its required "bindings" field
malformed_let = {"type": "Let"}  # <-- triggers assert at ast.py:1040

collections = [
    collection_set(expr=[malformed_let], id="bad-let"),
]

# EXPECTED (autofix): warning, collection dropped, no crash
# ACTUAL: AssertionError: Let missing bindings
g = graphistry.bind().collections(collections=collections, validate="autofix")
```

Same for:
- `{"type": "RemoteGraph"}` -- missing `dataset_id`
- `{"type": "ChainRef"}` -- missing `ref`
- `{"type": "Call"}` -- missing `function`
- Any node/edge with a malformed predicate filter

### Why this matters

The `from_json` methods using `assert` is pre-existing code -- it was written before collections validation existed, when these methods were only called from internal code paths that guaranteed well-formed input. The `assert` was a dev-time invariant check, not a user-facing validation.

But now that `normalize_gfql_to_wire` is called from the **user-facing validation boundary** in `_normalize_gfql_ops`, those asserts are reachable with arbitrary user input. The except clause was written with the assumption that all errors would be well-typed (`TypeError`, `ValueError`, `GFQLSyntaxError`, `GFQLValidationError`), but it missed the pre-existing assert-based error paths.

### The fix

**Minimal** -- add `AssertionError` to the except clause:

```python
except (TypeError, ValueError, AssertionError, GFQLValidationError, GFQLSyntaxError) as exc:
```

**More robust** -- use `Exception` since this is a validation boundary where *any* parse failure should be handled gracefully, not just the ones we've enumerated:

```python
except Exception as exc:
```

The broader `except Exception` is arguably better here because:
1. It's a **validation boundary** -- its contract is "normalize or report failure"
2. New AST classes added in the future might raise other exception types
3. The error message and structured data are still passed through `_issue()`, so debugging is not lost

The tradeoff is that `except Exception` could mask genuine bugs (e.g., `ImportError`, `RecursionError`). But given that this is a pure data-transformation function with no I/O or side effects, the risk is low.

# Proposed tests

These tests reproduce the issue:

```python
# -*- coding: utf-8 -*-
"""
Reproducer for: AssertionError from AST from_json leaks through
_normalize_gfql_ops exception handler.

_normalize_gfql_ops (validate_collections.py:174-185) wraps
normalize_gfql_to_wire in a try/except that catches:

    (TypeError, ValueError, GFQLValidationError, GFQLSyntaxError)

But several AST from_json class methods use bare `assert` statements
for required-field checks, which raise AssertionError -- a type NOT
in that except clause.  A malformed GFQL dict with a known `type` but
missing a required field bypasses the validation handler and crashes
as an unhandled AssertionError in both strict and autofix modes.

Call chain:
  _normalize_gfql_ops  (validate_collections.py:176)
    -> normalize_gfql_to_wire  (ast.py:1538)
         -> _normalize_op  -> from_json(op)  (ast.py:1538)
              -> ASTLet.from_json  (ast.py:1040)
                   -> assert 'bindings' in d  -->  AssertionError!

The except clause doesn't match, so the error propagates raw instead
of being routed through _issue() for proper mode handling.

Fix: add AssertionError to the except tuple, or use `except Exception`.

NOTE: We pass raw dicts to normalize_collections (not collection_set())
because collection_set() calls normalize_gfql_to_wire eagerly at
construction time -- before the validation boundary exists.  In practice,
users can also reach _normalize_gfql_ops by passing pre-serialized JSON
or raw dicts to g.collections().
"""

import json
import pytest
from urllib.parse import unquote

from graphistry.validate.validate_collections import normalize_collections


# ---- helpers ---------------------------------------------------------------

def _make_set_collection(op_dict, id="test"):
    """Build a raw collection dict with a single GFQL operation."""
    return {
        "type": "set",
        "id": id,
        "expr": [op_dict],
    }


# ---- AST class-level from_json asserts (ast.py) ---------------------------
#
# Each test below provides a dict with a valid `type` field recognized by
# the top-level from_json dispatcher (ast.py:1453-1508), but missing a
# required field that the class-level from_json checks with `assert`.
#
# EXPECTED behavior (after fix):
#   strict  -> ValueError via _issue(), NOT AssertionError
#   autofix -> collection silently dropped, no crash
#
# ACTUAL behavior (current code):
#   AssertionError propagates unhandled in both modes


class TestLetMissingBindings:
    """ASTLet.from_json (ast.py:1040): assert 'bindings' in d"""

    def test_autofix_does_not_crash(self):
        """autofix should drop the collection, not raise AssertionError."""
        result = normalize_collections(
            [_make_set_collection({"type": "Let"}, id="bad-let")],
            validate="autofix",
            warn=False,
        )
        assert isinstance(result, list)
        ids = [c.get("id") for c in result]
        assert "bad-let" not in ids

    def test_strict_raises_valueerror(self):
        """strict should raise ValueError (via _issue), not AssertionError."""
        with pytest.raises(ValueError):
            normalize_collections(
                [_make_set_collection({"type": "Let"}, id="bad-let")],
                validate="strict",
            )


class TestRemoteGraphMissingDatasetId:
    """ASTRemoteGraph.from_json (ast.py:1157): assert 'dataset_id' in d"""

    def test_autofix_does_not_crash(self):
        result = normalize_collections(
            [_make_set_collection({"type": "RemoteGraph"}, id="bad-rg")],
            validate="autofix",
            warn=False,
        )
        assert isinstance(result, list)
        ids = [c.get("id") for c in result]
        assert "bad-rg" not in ids

    def test_strict_raises_valueerror(self):
        with pytest.raises(ValueError):
            normalize_collections(
                [_make_set_collection({"type": "RemoteGraph"}, id="bad-rg")],
                validate="strict",
            )


class TestRefMissingFields:
    """ASTRef.from_json (ast.py:1273-1274): assert 'ref' in d / assert 'chain' in d"""

    def test_autofix_does_not_crash(self):
        result = normalize_collections(
            [_make_set_collection({"type": "ChainRef"}, id="bad-ref")],
            validate="autofix",
            warn=False,
        )
        assert isinstance(result, list)
        ids = [c.get("id") for c in result]
        assert "bad-ref" not in ids

    def test_strict_raises_valueerror(self):
        with pytest.raises(ValueError):
            normalize_collections(
                [_make_set_collection({"type": "ChainRef"}, id="bad-ref")],
                validate="strict",
            )


class TestCallMissingFunction:
    """ASTCall.from_json (ast.py:1415): assert 'function' in d"""

    def test_autofix_does_not_crash(self):
        result = normalize_collections(
            [_make_set_collection({"type": "Call"}, id="bad-call")],
            validate="autofix",
            warn=False,
        )
        assert isinstance(result, list)
        ids = [c.get("id") for c in result]
        assert "bad-call" not in ids

    def test_strict_raises_valueerror(self):
        with pytest.raises(ValueError):
            normalize_collections(
                [_make_set_collection({"type": "Call"}, id="bad-call")],
                validate="strict",
            )


# ---- predicates/from_json.py asserts --------------------------------------
#
# When a Node or Edge has a filter predicate, the predicate's from_json
# (predicates/from_json.py:33-42) uses bare asserts for type validation.
# A Node with a malformed predicate dict triggers the same leak.


class TestPredicateFromJsonAsserts:
    """predicates/from_json.py:34-36: assert isinstance(d, dict), assert 'type' in d, etc."""

    def test_node_with_bad_predicate_autofix(self):
        """Node filter with unknown predicate type -> assert d['type'] in type_to_predicate."""
        bad_node = {
            "type": "Node",
            "filter_dict": {"NotARealPredicate": {"column": "x"}},
        }
        result = normalize_collections(
            [_make_set_collection(bad_node, id="bad-pred")],
            validate="autofix",
            warn=False,
        )
        assert isinstance(result, list)
        ids = [c.get("id") for c in result]
        assert "bad-pred" not in ids

    def test_node_with_bad_predicate_strict(self):
        bad_node = {
            "type": "Node",
            "filter_dict": {"NotARealPredicate": {"column": "x"}},
        }
        with pytest.raises(ValueError):
            normalize_collections(
                [_make_set_collection(bad_node, id="bad-pred")],
                validate="strict",
            )


# ---- A valid collection alongside a bad one: autofix keeps the good one ---


class TestAutofixKeepsValidCollections:
    """Autofix should drop only the broken collection, keep the rest."""

    def test_good_survives_alongside_bad(self):
        good = {
            "type": "set",
            "id": "good",
            "expr": [{"type": "Node"}],  # valid minimal Node
        }
        bad = _make_set_collection({"type": "Let"}, id="bad-let")
        result = normalize_collections([good, bad], validate="autofix", warn=False)
        ids = [c.get("id") for c in result]
        assert "good" in ids, f"valid collection was dropped: {ids}"
        assert "bad-let" not in ids, f"invalid collection was kept: {ids}"

```

graphistry/validate/validate_collections.py
lmeyerov: Oh I didn't know we supported nested composition -- @mj3cheun does the UI allow intersections of intersections? I thought just a flat intersection-of-sets, but not sure

lmeyerov: ... i'm asking codex to investigate fep vs streamgl-viz, b/c i think pyg, as a client-facing, should match what the ui allows . We can in theory add a server-mode vs ui-mode, but I think that complexity is insufficiently motivated rn

aucahuasi (dismissed):
Thanks for fixing the previous issue, here are some others that caught my attention:

https://github.com/graphistry/pygraphistry/pull/874/changes#r2789010065
https://github.com/graphistry/pygraphistry/pull/874/changes#r2789125932

graphistry/validate/validate_collections.py
lmeyerov: yep i missed that, intersections can indeed be nested. i added more checking here, including for being a DAG, good catch!

graphistry/validate/validate_collections.py
lmeyerov: tx added

PR #935, opened 2026-03-09
add sso state ttl

vaimdevs:
```
##Fix
* Added SSO state TTL, added clear error message to hint user
```

Issue: https://github.com/graphistry/clients/blob/dbt-sso-databricks-issue/DBT/sso-databricks-issue/github_issues/pygraphistry-004-sso-state-ttl-enforcement.md

### DIFF
--- a/graphistry/client_session.py
+++ b/graphistry/client_session.py
@@ -71,6 +71,8 @@ class ClientSession:
 
         self.idp_name: Optional[str] = None
         self.sso_state: Optional[str] = None
+        self.sso_state_created_at: Optional[float] = None
+        self.sso_state_ttl_s: int = get_from_env("GRAPHISTRY_SSO_STATE_TTL_S", int, 300)
 
         self.personal_key: Optional[str] = None
         self.personal_key_id: Optional[str] = None
--- a/graphistry/pygraphistry.py
+++ b/graphistry/pygraphistry.py
@@ -19,7 +19,7 @@ from . import util
 from . import bolt_util
 from .plotter import Plotter
 from .util import in_databricks, setup_logger, in_ipython
-from .exceptions import SsoRetrieveTokenTimeoutException, TokenExpireException, SsoStateInvalidException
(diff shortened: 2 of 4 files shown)


graphistry/client_session.py
```diff
@@ -71,6 +71,8 @@ def __init__(self) -> None:
 
         self.idp_name: Optional[str] = None
         self.sso_state: Optional[str] = None
+        self.sso_state_created_at: Optional[float] = None
+        self.sso_state_ttl_s: int = get_from_env("GRAPHISTRY_SSO_STATE_TTL_S", int, 300)
```
lmeyerov, line 75: how do we know this is 300? I couldn't find , closest I saw was this: https://github.com/graphistry/graphistry/blob/eb8aba35feba0fdf85d1c99b163c89d30cd4d816/apps/core/nexus/nexus/allauth_ext/socialaccount/providers/openid_connect/client.py#L56

graphistry/pygraphistry.py
```diff
@@ -352,6 +355,16 @@ def _sso_get_token(self) -> Tuple[Optional[str], Optional[str]]:
         if state is None:
             raise SsoStateInvalidException("[SSO] Invalid SSO state: NoneType encountered")
 
+        created_at = self.session.sso_state_created_at
+        ttl = self.session.sso_state_ttl_s
+        if created_at is not None and (time.time() - created_at) > ttl:
```
lmeyerov, line 360: 1. especially as we're not confident of SSO expiry, instead of being preemptive and risk being wrong, probably better to instead limit this to a nicer error message in the case of a failure

Ex:
- try to do the action against the server
- server raises an exn
- we catch the exn, and if flagged as a potential sso expiry, stack the exns so they get both messages (important to get both, not just potentially incorrectly cloud with this)


2. i'm unsure if this is the right place to detect such an expiry; it'd be good to test that the correct library point is being tested (live test, vs synthetic)

PR #954, opened 2026-03-15
feat(gfql): add typed query AST and remote query transport

lmeyerov:
## Summary

This is the client-only slice of the unified GFQL query rollout.

It adds a first-class `query(...)` AST for local GFQL composition, widens `ref()` to support `ref(name, query(...))`, and updates remote GFQL transport to send canonical `gfql_query` payloads while preserving backward compatibility for legacy-lowerable forms.

## What changed

- add `query(...)` as a first-class GFQL AST and export it from the public API
- allow `let({...})` bindings to use `query(...)`
- allow `ref(name, query(...))` without introducing any new public ref operator
- reject `ASTQuery` inside `Chain([...])` so query programs do not silently go down the chain execution path
- update `gfql_remote()` / `chain_remote()` to send canonical `gfql_query`
- continue sending legacy `gfql_operations` when the query honestly lowers to the old chain transport
- surface a clear upgrade-required error when a typed-only query hits an older server that only understands `gfql_operations`
- add optional remote `output=` plumbing on the client side
- update `CHANGELOG.md`

## Compatibility contract

Direct remote execution with a new client should keep working against older servers for any query that still lowers to legacy `gfql_operations`.

Only genuinely typed-only remote forms should require newer server support. When that happens, the client now fails clearly instead of pretending legacy compatibility.

This PR does not change Forge / server code.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. uv run --no-project --with lark pytest -q graphistry/tests/compute/test_query_ast.py`
- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. uv run --no-project --with lark pytest -q graphistry/tests/compute/test_let.py graphistry/tests/compute/test_graph_operation.py graphistry/tests/compute/test_ast_errors.py graphistry/tests/test_chain_remote_auth.py`

### DIFF
--- a/graphistry/compute/chain_let.py
+++ b/graphistry/compute/chain_let.py
@@ -16,6 +16,19 @@ if TYPE_CHECKING:
 logger = setup_logger(__name__)
 
 
+def _execute_ast_query(
+    g: Plottable,
+    ast_query: ASTQuery,
+    engine: Engine,
+    policy=None,
+) -> Plottable:
+    return g.gfql(
+        ast_query,
+        engine=EngineAbstract(engine.value),
+        policy=policy,
+    )
+
+
 def extract_dependencies(ast_obj: Union[ASTObject, 'Chain', 'Plottable']) -> Set[str]:
     """Recursively find all ASTRef references in an AST object or GraphOperation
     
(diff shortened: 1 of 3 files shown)


graphistry/compute/ast.py
```diff
@@ -1172,6 +1172,115 @@ def reverse(self) -> 'ASTRemoteGraph':
...
+
+    def __init__(
+        self,
+        query: str,
+        language: str = "cypher",
+        params: Optional[Mapping[str, Any]] = None,
+        expect: Optional[str] = None,
+    ) -> None:
+        super().__init__()
+        self.query = query
```
lmeyerov, line 1192: allow more than `str` for `query` ?

graphistry/compute/ast.py
```diff
@@ -1172,6 +1172,115 @@ def reverse(self) -> 'ASTRemoteGraph':
...
+        self.language = language
+        self.params = dict(params) if params is not None else None
+        self.expect = expect
```
lmeyerov, line 1195: what is `expected` ?

lmeyerov:
Superseded by #962 (merged as v0.53.4). Core transport work (dual-field gfql_query + gfql_operations) shipped there with additional Cypher string, GRAPH constructor, and params support. The ASTQuery concept from this PR is interesting future work but not essential — users can pass Cypher strings directly to g.gfql() and gfql_remote() now. See findings at plans/docs-translate-cypher-extensions/findings/pr954-audit.md.

PR #970, opened 2026-03-28
fix(rapids): RAPIDS 26.02 backward-compatible compatibility fixes

aucahuasi:
RAPIDS 26.02 removed several APIs that pygraphistry depends on. All fixes are backward-compatible with RAPIDS 24.12, 25.02, and 26.02.

- cudf: Restore removed DataFrame.from_pandas()/Series.from_pandas() via monkey-patch shim in __init__.py (no call-site changes needed)
- cuml: Guard DBSCAN.fit() calc_core_sample_indices param removal using inspect.signature check
- cugraph: Fallback removed jaccard_w/overlap_w/sorensen_w to base algorithm with use_weight=True
- gfql: Fix cudf tz-aware binary op crash in temporal comparisons by stripping tz after conversion; fix s.dt.tz AttributeError on cudf < 26.02 using getattr
- tigergraph: Replace removed pd.Series.append() with pd.concat()
- tests: Fix contradictory datetime dtype assertion in honeypot_pdf()

Tested across RAPIDS 24.12, 25.02, 26.02, and CPU-only (no cudf). CI-matched: 2584 passed, 0 failed on CPU-only Python 3.10.

### DIFF
--- a/graphistry/__init__.py
+++ b/graphistry/__init__.py
@@ -1,17 +1,5 @@
 """PyGraphistry: A visual graph analytics library for big graphs."""
 
-# RAPIDS 26.02+ removed cudf.DataFrame.from_pandas() and cudf.Series.from_pandas().
-# Restore them so existing code (ours and downstream) keeps working on all versions.
-try:
-    import cudf as _cudf
-    if not hasattr(_cudf.DataFrame, 'from_pandas'):
-        _cudf.DataFrame.from_pandas = staticmethod(_cudf.from_pandas)
-    if not hasattr(_cudf.Series, 'from_pandas'):
-        _cudf.Series.from_pandas = staticmethod(_cudf.from_pandas)
-    del _cudf
-except ImportError:
-    pass
-
 from graphistry.pygraphistry import (  # noqa: E402, F401
     client_protocol_hostname,
     protocol,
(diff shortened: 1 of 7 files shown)


graphistry/compute/predicates/comparison.py
```diff
@@ -49,33 +49,55 @@ def _prepare_temporal_series(self, s: SeriesT, temporal_val: TemporalValue) -> S
...
+                if getattr(s.dt, 'tz', None) is None:
+                    result = s.dt.tz_localize('UTC').dt.tz_convert(tzinfo)
                 else:
-                    return s.dt.tz_convert(tzinfo)
+                    result = s.dt.tz_convert(tzinfo)
+                # cudf does not support binary ops on tz-aware Series (RAPIDS 26.02+);
+                # strip tz after conversion — values are already in the target timezone
+                if 'cudf' in str(type(result)):
+                    result = result.dt.tz_localize(None)
+                return result
```
lmeyerov, line 60: do we need tests for the different version cases here?

lmeyerov (commented):
It'd be helpful if we started some sort of containerized version range test script for supported rapids versions we can run on GPU boxes, to help verify this

graphistry/__init__.py
```diff
@@ -1,5 +1,17 @@
 """PyGraphistry: A visual graph analytics library for big graphs."""
 
+# RAPIDS 26.02+ removed cudf.DataFrame.from_pandas() and cudf.Series.from_pandas().
+# Restore them so existing code (ours and downstream) keeps working on all versions.
+try:
+    import cudf as _cudf
```
lmeyerov: I'm not a fan of importing cudf at start...

lmeyerov: This is also fraught in practice: Imagine something like colab, where often you'll have cudf import , even if no gpu available

There are other files where we silo a lot of the engine-specific stuff, maybe belongs with those, and in their style

Issue #1075, opened 2026-04-06
tooling: explore adding pyright in monitor mode to catch possibly-unbound variables


lmeyerov:
## Problem

During #977 (cudf SIGSEGV fix), a real `UnboundLocalError` in `hop.py` was missed by both ruff and mypy:

```python
# edge_map assigned only inside this branch:
if g_out._edges is not None and edge_hop_col is not None and ...:
    edge_map = ...

# then used outside it — NameError/UnboundLocalError at runtime:
if missing_mask.any() and edge_hop_col is not None and ...:
    ... safe_map_series(..., edge_map)  # edge_map possibly unbound!
```

The bug was masked by an upstream SIGSEGV (so it never surfaced in production), and only caught by writing a test that exercised the specific branch combination (\`label_node_hops=True, label_edge_hops=None\` with a non-empty \`missing_mask\`).

## Why ruff/mypy missed it

- **ruff F821** only catches names with *no* assignment anywhere in scope — not conditionally-assigned names
- **mypy** does not flag "possibly unbound" from conditional branches unless the variable is annotated before the branch (e.g. \`edge_map: pd.Series\` declared before the \`if\`)
- Neither tool does full dataflow / reachability analysis for this pattern

## What would catch it

**pyright** (default mode, not even strict) detects "possibly unbound" variables from conditional assignments and flags them at the use site. This is a well-known difference from mypy.

## Proposal

Explore adding pyright in **monitor / informational mode** to CI as a sequence of stacked PRs:

1. Observability
- Run pyright on the \`graphistry/\` source (excluding tests, matching mypy.ini scope)
- Enforce in CI any rules that already pass
- Initially non-blocking (warnings only) for rest to assess the volume of pre-existing issues
- Clear reporting step in CI for easy auditing

2. Initial refactor
- Create and implement on a metaissue for initial target of rules that the repo can enforce
- There should be a small number of changes that go a long way
- Promote these to blocking, and leave the rest as warning
- For local dev flow, add pyright to the mypy steps as a default-on option

3. Follow-on refactors
- Track a metaissue prioritized rules that fail
- Sequence of PRs that fix each one

This does not replace mypy — they catch different things. Pyright's dataflow analysis is complementary.

## References

- #977 — the bug this would have caught
- \`graphistry/compute/hop.py\` lines 952–974 — the specific pattern
- \`mypy.ini\` — current type-check config

PR #1123, opened 2026-04-10
feat: add pyright script for type checking with configuration

JithinBathula:
## Summary
Adds pyright configuration and local runner script to catch conditionally-assigned variable bugs (like edge_map at hop.py:974) that ruff and mypy miss. No code changes — tooling only.

- [X] Linked issue(s): Ref #1075 
- [X] Scope is explicit (non-goals included when relevant)

Non-goals: CI integration, code fixes, rule enforcement — all follow-up PRs.

## Validation

- [X] Local lint/type/tests run for touched scope
- [ ] CI is green

`bin/pyright.sh graphistry/compute/hop.py` confirms `hop.py:974:122 - warning: "edge_map" is possibly unbound`

## Cypher Frontend CI Evidence (when PR touches cypher frontend / IR scope)

- [ ] `cypher-frontend-strict-typing (py3.12)` passed (strict typing gate)
- [ ] `cypher-frontend-differential-parity (py3.12)` passed (trust-but-verify gate)
- [ ] `cypher-frontend-ci-gates` passed (includes `test-minimal-python` sentinel)
- [ ] PR body includes links/screenshots/log snippets for any non-obvious gate evidence

### DIFF
--- a/bin/pyright.sh
+++ b/bin/pyright.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+set -e
+
+# Run from project root
+# Non-zero exit code on fail
+
+# Resolve pyright command, then delegate with repo config.
+if command -v uvx >/dev/null 2>&1; then
+    PYRIGHT_CMD_ARR=(uvx --from pyright pyright)
+elif command -v pyright >/dev/null 2>&1; then
+    PYRIGHT_CMD_ARR=(pyright)
+elif command -v npx >/dev/null 2>&1; then
+    PYRIGHT_CMD_ARR=(npx pyright)
+else
+    echo "pyright not found. Install pyright, npm, or uvx."
+    exit 1
+fi
+
+PYRIGHT_EXTRA_ARGS_ARR=()
+if [ -n "${PYRIGHT_EXTRA_ARGS:-}" ]; then
+    read -r -a PYRIGHT_EXTRA_ARGS_ARR <<< "$PYRIGHT_EXTRA_ARGS"
+fi
+
+# Get the script directory and repo root
+SCRIPT_DIR="$( cd "$( dirname -- "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+# Change to repo root
+cd "$REPO_ROOT"
+
+# Check if specific files were passed as arguments
+if [ $# -eq 0 ]; then
+    TARGET_ARGS=("graphistry")
+else
+    TARGET_ARGS=("$@")
+fi
+CONFIG_ARGS=(-p pyrightconfig.json)
+
+echo "Running pyright..."
+echo "Checking: ${TARGET_ARGS[*]}"
+
+"${PYRIGHT_CMD_ARR[@]}" --version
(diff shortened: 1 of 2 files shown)


lmeyerov:
The autocomment on flipping precedence order seems right, it'd match the other shell scripts here. Can you change? Thanks!

lmeyerov (changes requested):
Thanks! See https://github.com/graphistry/pygraphistry/pull/1123#issuecomment-4229154629 and then we can merge

lmeyerov:
@codex I agree with your review below, please address

> P2 Badge Fall through to next backend when uvx invocation fails

> Selecting the backend solely by command -v means the script commits to uvx whenever it is installed, and with set -e it exits on the first uvx failure without ever trying npx. This breaks type-check runs in environments where uvx is present but cannot execute pyright (for example, restricted package access) while npx pyright is still usable, so the intended fallback chain is not actually resilient.

lmeyerov:
@codex address that feedback

JithinBathula:
Updated the script based on the comment. Thanks

lmeyerov:
Thank you for this, @JithinBathula — and apologies it sat open so long.

Your work has landed. `bin/pyright.sh` and the `pyrightconfig.json` changes shipped essentially as you wrote them in #2092, with you as `Co-Authored-By` on the commit. The resolution order you set up (local `pyright` → `uvx` → `npx`) is still there, and it turned out to matter more than it looked: when I later tested a broken `include` in the config, the explicit CLI target your script passes was the thing that kept pyright analysing the right 351 files. I nearly "simplified" it away before measuring.

Two things had to be settled before it could gate, both of which you'd flagged as non-goals:

**The tool needed pinning.** The version is now fixed at `pyright@1.1.414` in both fetching fallbacks, because a ratchet baseline is only meaningful against one version — upstream adds and retunes rules between releases.

**The rules needed sorting by whether they're reproducible.** Running your config across five environments (python 3.8/3.11/3.12/3.14, and a workstation with polars + cudf + scipy + scikit-learn) showed most pyright rules move with the environment: `reportAttributeAccessIssue` ranged from 146 to 810 findings on the *same* tree. Only rules decided by a file's own control flow, names and syntax were stable — so those five gate, and the rest are reported but never enforced. `reportPossiblyUnboundVariable`, the one tightening in your config, is among the five, and it's the reason the gate exists.

It's now a real CI job (`python-pyright`) holding findings to a per-file ratchet, so existing debt is grandfathered and new code isn't.

And it found real bugs. The `hop.py` "edge_map is possibly unbound" you cited in your description is exactly the class of thing it catches. Follow-up PRs using it fixed, among others:

- `graphistry.outliers` couldn't be imported at all without matplotlib, which isn't a dependency — so on a default install, `import graphistry.outliers` raised `AttributeError` (#2094)
- two `feature_utils` functions raised `UnboundLocalError` on reachable inputs (#2097)
- `GraphBase.__init__` read a component variable that no edge had bound (#2096)

The ratchet went from 273 findings to 230 across that series. It started with your PR.

Closing this one as superseded, not rejected — the code is in `master`. Thanks again, and sorry for the wait on review.

lmeyerov:
Superseded by #2092, which carries this work with attribution. See the comment above for where each piece landed.

PR #1140, opened 2026-04-17
fix(plugins): handle cuDF DataFrames in igraph and graphviz plugins

mj3cheun:
## Summary

- [x] Linked issue(s): No existing issue
- [x] Scope is explicit (non-goals included when relevant)

CPU-only plugins (`compute_igraph`, `layout_igraph`, `layout_graphviz`) crash with `TypeError` when the graph is backed by cuDF DataFrames, because igraph and pygraphviz only accept pandas. This PR:

- Adds `_ensure_pandas()` to convert cuDF→pandas before entering CPU libraries, using `nullable=True` to preserve nullable integer dtypes
- Adds `_restore_engine()` to convert output back to the original engine (auto-detected, not hardcoded)
- Adds CPU-only docstring notes to all 4 affected functions pointing to GPU alternatives
- Wires conversion into `to_igraph`, `compute_igraph`, `layout_igraph`, `g_to_pgv`, `layout_graphviz`

**Non-goals:**
- Fixing dask-cuDF (`.to_pandas()` triggers `.compute()` — separate concern)
- Adding engine awareness to the Cypher `_execute_backend_call` dispatcher (not needed since `compute_igraph` now handles it internally)

**Files changed:**
- `graphistry/plugins/igraph.py` — `_ensure_pandas`, `_restore_engine`, `_resolve_original_engine`; wired into `to_igraph`, `compute_igraph`, `layout_igraph`; docstring notes
- `graphistry/plugins/graphviz.py` — imports helpers; wired into `g_to_pgv`, `layout_graphviz`; docstring note
- `graphistry/tests/compute/test_call_operations.py` — 9 new tests in `TestCpuOnlyCallsCudfHandling` and `TestEnsurePandasDtypeFidelity`
- `CHANGELOG.md` — Fixed + Tests entries

## Validation

- [x] Local lint/type/tests run for touched scope
  - `pytest graphistry/tests/plugins/ graphistry/tests/compute/` — 2079 passed, 248 skipped, 0 failures
- [ ] CI is green


graphistry/plugins/igraph.py
```diff
@@ -1,12 +1,60 @@
...
+    """
+    from graphistry.Engine import df_to_engine
+
+    engine = _resolve_original_engine(original_nodes) or _resolve_original_engine(original_edges)
+    if engine is None:
+        return g
+
+    try:
+        if g._nodes is not None:
+            g = g.nodes(df_to_engine(g._nodes, engine), g._node)
```
dess890: could we also preserve `__index__` here? right now in one of the hacks in t2g we re-add it after every gfql run because it gets dropped in the pandas roundtrip, and is needed to map filtered nodes back to their positions

graphistry/plugins/igraph.py
```diff
@@ -1,12 +1,60 @@
...
+from typing import Any, List, Optional
 from graphistry.constants import NODE
 from graphistry.Plottable import Plottable
 from graphistry.util import setup_logger
 logger = setup_logger(__name__)
 
-#import logging
-#logger.setLevel(logging.DEBUG)
+
+def _ensure_pandas(df: Any) -> pd.DataFrame:
```
lmeyerov: Can we move this generic utility method to where other dataframe coercion methods are, and if any already doing this, reuse?

graphistry/plugins/igraph.py
```diff
@@ -1,12 +1,60 @@
...
+    """
+    if isinstance(df, pd.DataFrame):
+        return df
+    try:
+        return df.to_pandas(nullable=True)
+    except TypeError:
+        return df.to_pandas()
+
+
+def _resolve_original_engine(df: Any):
```
lmeyerov: Same, keep with other df utility methods

lmeyerov: Or is this unnecessary and just call resolve engine directly ?

graphistry/plugins/igraph.py
lmeyerov: What is index? I don't use indexes on dataframes much, so unsure if/how it relates here

lmeyerov: Cc @mj3cheun

graphistry/tests/compute/test_call_operations.py
```diff
@@ -622,3 +622,283 @@ def test_group_in_a_box_layout_params(self):
...
         assert exc_info.value.code == ErrorCode.E201
+
+
+def _make_sample_graph():
+    edges = pd.DataFrame({'source': [0, 1, 2, 2], 'target': [1, 2, 0, 3]})
+    nodes = pd.DataFrame({'node': [0, 1, 2, 3]})
+    return CGFull().edges(edges, 'source', 'target').nodes(nodes, 'node')
+
+
+class _CudfStubDataFrame:
```
lmeyerov: Currently we catch cudf errors on runs with cudf etc installed vs via CPU testing. I worry a bit on fidelity and surprise issues by introducing indirection abstractions here. In principle I like the idea of being able to test more without a GPU being available, but I do worry if this will cause false confidence and other issues.

graphistry/plugins/graphviz.py
```diff
@@ -51,7 +52,8 @@ def g_to_pgv(
         if x_col and y_col and x_col in g._nodes.columns and y_col in g._nodes.columns:
             pos_cols = (x_col, y_col)
 
-    for _, row in g._nodes.iterrows():
+    nodes_pdf = _ensure_pandas(g._nodes)
+    for _, row in nodes_pdf.iterrows():
```
lmeyerov, line 56: Can we add targeted tests to go along with these?

lmeyerov (changes requested):
Thanks!

See comments -

* Add igraph + graphviz tests as part of their dedicated test files

* Move df utility functions to sit next to the other df utility ones 

Also:

* I'm skeptical about the CPU test mode being done for GPU code, but I won't push hard here
* I have to wonder if there is more surface to audit for the same issue
* I'm unclear on the discussion around `__index__`, maybe that should be a follow on pr / discussion to help this one get through meanwhile ?

graphistry/plugins/igraph.py
mj3cheun: This is an internal graphistry (really collections) column

Not sure this has relevance here?

Cc @dess890

graphistry/plugins/igraph.py
mj3cheun: If I remember resolve engine reads a flag, whereas this function reverses any coercion that might have taken place if needed

graphistry/tests/compute/test_call_operations.py
mj3cheun: True, will change to using real cudf

graphistry/plugins/igraph.py
dess890: yeah disregard, apologies, was thinking of how caller columns can be reattached with` original_nodes/edges` but it is indeed a graphistry side issue to think through (the context being one of the changes i made in fep for the mcp)

graphistry/plugins/igraph.py
lmeyerov: Right, I mean all this does is call resolve with auto, so unsure it's needed

I don't want to block on this discussion, it's a nit

mj3cheun: removing, sorry misunderstood

mj3cheun:
@lmeyerov ready for final review when convenient

graphistry/compute/engine_coercion.py
```diff
@@ -195,3 +195,34 @@ def ensure_engine_match(g: Plottable, requested_engine: Engine) -> Plottable:
             exc_info=True
         )
         return g
+
+
+def ensure_pandas(df: Any) -> pd.DataFrame:
```
lmeyerov, line 200: I don't think this will work on dask_cudf, dask, arrow, etc

lmeyerov: (can handle in follow-on)

graphistry/plugins/igraph.py
```diff
@@ -1,12 +1,29 @@
...
+from typing import Any, List, Optional
 from graphistry.constants import NODE
 from graphistry.Plottable import Plottable
 from graphistry.util import setup_logger
 logger = setup_logger(__name__)
 
-#import logging
-#logger.setLevel(logging.DEBUG)
+
+def ensure_pandas(df: Any) -> pd.DataFrame:
```
lmeyerov, line 9: unsure why circular import, but ok

graphistry/tests/compute/test_call_operations_gpu.py
```diff
@@ -318,3 +318,288 @@ def test_encode_with_gpu(self):
...
+    """Verify CPU-only plugins (igraph, graphviz) handle cuDF input end-to-end.
+
+    These tests use real cuDF DataFrames to validate:
+    1. ensure_pandas converts cuDF to pandas before entering CPU libraries
+    2. restore_engine converts output back to cuDF
+    3. Nullable integer dtypes survive the round-trip via nullable=True
+    """
+
+    @skip_gpu
+    def test_compute_igraph_cudf_round_trip(self):
```
lmeyerov, line 333: these really should be in something like `tests/plugins/test_{graphviz,igraph}.py` to preserve mirror structure

lmeyerov (approved):
See comments, a few style nits, but to avoid blocking progress: accepting, and we can do cleanup in follow-ons

Issue #1194, opened 2026-04-23
gfql/cypher: resolve Lark ambiguity between where_predicates and expr (grammar-level follow-up to #1125)


lmeyerov:
## Context

PR #1193 (issue #1125) hardened WHERE label narrowing by replacing a fragile regex in \`binder.py\` with AST-derived narrowing plus a text-split fallback in \`parser.py::generic_where_clause\`.

The text-split fallback exists because of a Lark grammar ambiguity: \`WHERE n:Admin AND n:Active\` can match BOTH the structured rule \`where_predicates: where_predicate (\"AND\"i where_predicate)*\` (parser.py:127) AND the generic expression path \`where_clause: \"WHERE\"i expr -> generic_where_clause\` (parser.py:126). Lark's ambiguity resolution prefers the generic path, so the AND-joined label predicates land in \`WhereClause.expr\` as raw text rather than as structured \`WhereClause.predicates\`.

The current PR works around this at the transformer level — \`generic_where_clause\` re-parses the raw text and lifts it back into structured predicates via \`_split_top_level_and_terms\` + \`_BARE_LABEL_PREDICATE_RE\`. This is correct and well-tested, but architecturally backward: the grammar already declares the structured path we want; we should let it win.

## Proposed follow-up

Fix the ambiguity at the grammar layer so AND-joined bare label predicates naturally route to \`where_predicates\`:

1. Investigate why \`expr\` currently wins the ambiguity contest — likely because \`bare_label_predicate_expr\` is reachable from \`?predicate\` inside \`?and_expr\`.
2. Either:
   - Remove the \`bare_label_predicate_expr\` alternative from the \`expr\` path, forcing label predicates to go through \`where_predicates\` exclusively, OR
   - Raise the priority of \`where_predicates\` (Lark rule priority annotation), OR
   - Introduce an explicit \`where_clause\` disambiguator rule.
3. Once the grammar disambiguates correctly, the text-split logic in \`generic_where_clause\` becomes dead code for AND-joined labels — simplify to the single-label case (still a valid generic fallback) or remove entirely.

## Tests to keep green

- All 10 label-narrowing tests in \`test_binder.py\` (single, double, triple AND, multi-alias, multi-label, lowercase-and, mixed label+property, XOR/OR/NOT conservative, string-literal false-positive guards).
- All 4 parser-level AST tests in \`test_parser.py\`.

## Constraints

[...]

PR #1198, opened 2026-04-23
fix(gfql): correct \b regex bug + consolidate AND-splitter (#1195)

### LINKED ISSUE #1195: gfql/passes: fix \\b regex bug + DRY consolidation in predicate_pushdown.py (retroactive audit of #1188)
## Context

Retroactive audit of merged PR #1188 (predicate pushdown pass rewrite, commit 23fb3a9f) surfaced two issues missed in review.

## Bug: \`\\\\b\` inside rf-string at \`predicate_pushdown.py:178\`

\`\`\`python
if re.search(rf\"\\\\b{re.escape(alias)}\\\\b\", segment)
\`\`\`

Inside an rf-string, \`\\\\b\` is a literal backslash followed by \`b\`, not a word boundary. Compare to the correct form at line 211 of the same file:

\`\`\`python
if re.search(rf\"\\b{re.escape(alias)}\\b\", expression)
\`\`\`

**Effect:** \`_refs_for_segment\` never matches an alias; the conservative fallback at line 183 (\`return original_refs\`) always fires. Per-conjunct reference narrowing after an AND split is effectively dead code. Tests pass because the fallback is a superset, hiding the bug.

## DRY: Duplicate AND splitter

\`predicate_pushdown.py:129\` defines \`_split_top_level_and\` — a quote/paren-aware top-level AND splitter. The same package already has \`_split_top_level_and_terms\` in \`cypher/parser.py:477\`, which is strictly more robust: handles \`[]\`, \`{}\`, backticks, and escape sequences.

This is the exact same class of miss that PR #1193 fixed during Wave 3 review (see issue #1125 follow-up). Two near-duplicate implementations of \"split WHERE body on top-level AND\" in one package.

## Proposed fixes

1. **Correct the \`\\\\b\` regex** to \`\\b\`. Add a targeted regression test: a multi-alias OPTIONAL MATCH expression where at least one alias must be detected per-segment (i.e., the fallback-to-\`original_refs\` path would incorrectly widen safety guards).
2. **Consolidate the two splitters**: promote \`_split_top_level_and_terms\` to a shared module (e.g., \`graphistry/compute/gfql/common/expr_split.py\`) and delete \`_split_top_level_and\` from \`predicate_pushdown.py\`.

## Missed-in-review gap

The earlier review of #1188 used the builtin \`/review\` tool. The retroactive audit ran the expanded multi-dimension review skill and caught both. Process change: for gfql/cypher PRs, default to the multi-dimension skill.

## Priority

- Bug: **p2** (silent semantic widening; not load-bearing for correctness under current callers, but could become so)
- DRY: **p3** (cleanup, unblocks deletion of the duplicate splitter)

lmeyerov:
## Summary

Closes #1195.

### Bug fix

\`_refs_for_segment\` in \`predicate_pushdown.py\` used \`rf\"\\\\b{re.escape(alias)}\\\\b\"\`. Inside an rf-string, \`\\\\b\` is a literal backslash-b, not a word-boundary assertion. Compare the correct form at line 232 of the same file: \`rf\"\\b...\\b\"\`. Effect: per-segment alias detection never matched and always fell back to \`original_refs\`, silently widening the reference set for every split conjunct. Tests passed because the fallback is a superset — the bug was invisible to existing coverage.

### DRY consolidation

Two near-duplicate \"split WHERE body on top-level AND\" implementations existed in one package:

| Location | Quote-aware | Bracket-aware | Backtick-aware | Escape-aware |
|---|---|---|---|---|
| \`parser.py::_split_top_level_and_terms\` | ✓ | ✓ (paren, \`[]\`, \`{}\`) | ✓ | ✓ |
| \`predicate_pushdown.py::_split_top_level_and\` | partial | paren only | ✗ | ✗ |

Consolidated into \`graphistry/compute/gfql/expr_split.py::split_top_level_and\` with the stricter semantics. Three call sites migrated; two local implementations removed.

### Tests

- \`test_expr_split.py\`: 18 direct unit tests covering empty, single-term, multi-AND, case-insensitivity, word-boundary inside identifiers (\`ANDROID\`, \`BRAND\`), AND inside single/double quotes, backticks, parens, square brackets, curly braces, escape sequences, and malformed edge cases (leading/trailing/consecutive AND).
- \`test_predicate_pushdown_pass.py\`: regression test locking the \`\\b\` fix — multi-alias OPTIONAL MATCH with \`m.x > 1 AND n.y > 2\` where \`m\` is pushable and \`n\` must be held back. Pre-fix, both conjuncts carried \`{m, n}\` references and the m-only conjunct was incorrectly treated as null-rejecting → blocked. Post-fix, per-conjunct narrowing lets the m-only conjunct push through.

### Test plan

- [x] \`pytest graphistry/tests/compute/gfql/test_expr_split.py\` — 18 pass
- [x] \`pytest graphistry/tests/compute/gfql/\` — 1445 pass, 80 skipped, 15 xfailed (+21 from new tests)
- [x] \`ruff check\` — clean on all touched files
- [x] \`mypy\` — no new errors (pre-existing errors in unrelated files)


lmeyerov:
Architectural follow-up filed as #1200: `expr_split.py` introduced here should eventually be removed once Lark's already-parsed `and_op`/`or_op` tree is exposed structurally in IR.  The text-split is a pragmatic intermediate — #1200 + #1194 together eliminate text re-parsing in both parser and pushdown paths.

Issue #1200, opened 2026-04-24
gfql/ir: expose Lark and_op/or_op tree in WhereClause so passes walk structure instead of re-parsing text


lmeyerov:
## Context

\`WhereClause\` currently has two shapes for a WHERE body:

\`\`\`python
class WhereClause:
    predicates: Tuple[Union[WherePredicate, WherePatternPredicate], ...] = ()
    expr: Optional[ExpressionText] = None
\`\`\`

Structured predicates land in \`.predicates\`.  Anything Lark's \`where_predicates\` rule cannot decompose (nested OR/XOR/NOT, mixed boolean forms, parenthesized subexpressions) falls into \`.expr\` as a raw text blob:

\`\`\`python
@dataclass
class ExpressionText:
    text: str
    span: Optional[SourceSpan] = None
\`\`\`

Lark's grammar already parses the expression structurally via dedicated rules (\`parser.py:177-183\`):

\`\`\`
?or_expr: xor_expr
        | or_expr \"OR\"i xor_expr            -> or_op
?xor_expr: and_expr
         | xor_expr \"XOR\"i and_expr         -> xor_op
?and_expr: not_expr
         | and_expr \"AND\"i not_expr         -> and_op
\`\`\`

…but the transformer throws the \`and_op\` / \`or_op\` / \`xor_op\` / \`not_op\` tree away and captures only \`self._slice(span)\`.  Downstream code (binder serialization → \`BoundPredicate.expression\` string → \`predicate_pushdown._split_conjuncts\`) re-parses the text via regex-based top-level-AND splitting (\`graphistry/compute/gfql/expr_split.py\`, introduced by PR #1198 / issue #1195).

## Why this is a smell

Two separate systems re-implement what Lark already computed:

1. \`parser.py::generic_where_clause\` re-parses label-only AND conjunctions.  **Tracked by #1194** (grammar-level disambiguation between \`where_predicates\` and \`expr\`).
2. \`predicate_pushdown._split_conjuncts\` re-parses top-level AND conjuncts in arbitrary boolean expressions.  **This issue.**

Each re-parse is a potential source of divergence (e.g., the \`\\\\b\` vs \`\\b\` bug fixed in #1198 \`_refs_for_segment\`).

## Proposed follow-up

Expose the parsed boolean-expression tree structurally in IR:

\`\`\`python
@dataclass
class BooleanExpr:
    op: Literal[\"and\", \"or\", \"xor\", \"not\", \"atom\"]
[...]

PR #1203, opened 2026-04-24
fix(cypher): grammar-resolve where_predicates vs expr ambiguity (#1194)

### LINKED ISSUE #1194: gfql/cypher: resolve Lark ambiguity between where_predicates and expr (grammar-level follow-up to #1125)

lmeyerov:
## Summary
- resolves #1194 by replacing regex/text-split lifting in `generic_where_clause` with grammar-driven reparsing of ambiguous `WHERE` clauses
- keeps primary Cypher parser on `lalr` (to avoid query-shape regressions), and adds a targeted `earley` `where_clause` subparser for ambiguous label-predicate cases
- hardens binder label narrowing to an explicit all-or-nothing policy: only narrow when the full WHERE predicate set is label-only
- updates parser/binder tests for routing and conservative mixed-predicate behavior

## Why this shape
- a full parser switch to `earley` fixed ambiguity but regressed clause attachment (for example `WITH ... WHERE ...`)
- targeted `where_clause` reparsing preserves existing global parse behavior while removing the regex fallback and still using grammar-level disambiguation

## Validation
- `PYTHONPATH=. uv run --no-project --with lark --with pytest python -m pytest graphistry/tests/compute/gfql/cypher/test_parser.py -q`
- `PYTHONPATH=. uv run --no-project --with lark --with pytest python -m pytest graphistry/tests/compute/gfql/cypher/test_binder.py -q`
- `PYTHONPATH=. uv run --no-project --with lark --with pytest python -m pytest graphistry/tests/compute/gfql/cypher/test_lowering.py -q`
- `PYTHONPATH=. uv run --no-project --with ruff ruff check graphistry/compute/gfql/cypher/parser.py graphistry/compute/gfql/frontends/cypher/binder.py graphistry/tests/compute/gfql/cypher/test_parser.py graphistry/tests/compute/gfql/cypher/test_binder.py`

### DIFF
--- a/graphistry/compute/gfql/cypher/parser.py
+++ b/graphistry/compute/gfql/cypher/parser.py
@@ -1046,8 +1046,27 @@ def _build_transformer(source: str, *, allow_where_reparse: bool = True) -> _Tra
                 try:
                     reparsed_tree = _where_clause_parser().parse(clause_text)
                     reparsed_node = _build_transformer(clause_text, allow_where_reparse=False).transform(reparsed_tree)
-                    if isinstance(reparsed_node, WhereClause) and reparsed_node.expr is None and len(reparsed_node.predicates) > 0:
-                        return reparsed_node
+                    if (
+                        isinstance(reparsed_node, WhereClause)
+                        and reparsed_node.expr is None
+                        and len(reparsed_node.predicates) > 0
+                        and all(
+                            isinstance(term, WherePredicate)
+                            and term.op == "has_labels"
+                            and isinstance(term.left, LabelRef)
+                            for term in reparsed_node.predicates
+                        )
+                    ):
+                        predicates = tuple(
+                            WherePredicate(
+                                left=LabelRef(alias=term.left.alias, labels=term.left.labels, span=span),
+                                op="has_labels",
+                                right=None,
+                                span=span,
+                            )
+                            for term in cast(Tuple[WherePredicate, ...], reparsed_node.predicates)
+                        )
+                        return WhereClause(predicates=predicates, expr=None, span=span)
                 except Exception:
                     # Keep generic expr fallback if sub-parse fails or still
                     # resolves to a generic clause.
--- a/graphistry/compute/gfql/cypher/parser.py
+++ b/graphistry/compute/gfql/cypher/parser.py
@@ -1057,15 +1057,18 @@ def _build_transformer(source: str, *, allow_where_reparse: bool = True) -> _Tra
                             for term in reparsed_node.predicates
                         )
                     ):
-                        predicates = tuple(
-                            WherePredicate(
-                                left=LabelRef(alias=term.left.alias, labels=term.left.labels, span=span),
-                                op="has_labels",
-                                right=None,
-                                span=span,
+                        predicates_list: List[WherePredicate] = []
+                        for term in cast(Tuple[WherePredicate, ...], reparsed_node.predicates):
+                            label_ref = cast(LabelRef, term.left)
+                            predicates_list.append(
+                                WherePredicate(
+                                    left=LabelRef(alias=label_ref.alias, labels=label_ref.labels, span=span),
+                                    op="has_labels",
+                                    right=None,
+                                    span=span,
+                                )
                             )
-                            for term in cast(Tuple[WherePredicate, ...], reparsed_node.predicates)
-                        )
+                        predicates = tuple(predicates_list)
                         return WhereClause(predicates=predicates, expr=None, span=span)
                 except Exception:
                     # Keep generic expr fallback if sub-parse fails or still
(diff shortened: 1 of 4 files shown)


lmeyerov:
Review rerun (wave 3/4 equivalent) found no blocking issues, but one suggestion-level hardening item:

- In `graphistry/compute/gfql/cypher/parser.py` (`generic_where_clause`), the new sub-parse path currently uses a broad `except Exception: pass` around `_where_clause_parser().parse(...)` + transformer.
- Suggest narrowing this to the expected parse/transform failures (for example `LarkError` and known parser/validation exceptions), so unrelated runtime defects do not get silently downgraded into generic raw-expression fallback.

Current implementation is functionally sound for #1194 scope; this is mainly a robustness/observability improvement for future maintenance.

lmeyerov:
Closing in favor of #1209, which addresses #1194 via the cleaner approach now that #1200 slice 1 (#1202) has landed.

**What changed since this PR was opened**:
- #1202 merged on master, exposing `WhereClause.expr_tree` — Lark's parsed boolean tree as a structural value. This was not available when this PR was started.
- #1207 merged on master, teaching the binder to walk `expr_tree` and emit one `BoundPredicate` per top-level AND conjunct.

**Why this PR is superseded**:
- This PR replaces `split_top_level_and` + regex inside `generic_where_clause` with a *secondary Earley parser instance* (`_where_clause_parser`) that re-parses the WHERE clause text. Functionally correct, but it introduces a **second Lark parser path** alongside the primary LALR + the now-available structural `expr_tree`.
- After #1202 + #1207 landed, the right path is to walk `expr_tree` directly — single source of truth, no second parser, no text re-splitting.

**Replacement (#1209)**:
- Adds `_match_bare_label_atom` and `_lift_label_only_and_spine` helpers in `cypher/parser.py`.
- `generic_where_clause` walks `BooleanExpr` (AND-spine) or `_ExpressionSlice.text` (single atom) directly.
- Same I/O contract as this PR for every WHERE shape; verified across the parser/binder/conformance/slice-2 binder_expr_tree/lowering suites (1528 passed).
- Shrinks the duplicate-parsing surface flagged by #1200 by retiring the regex+split path inside `generic_where_clause`.

The branch (`m4-1194-lark-where-ambiguity`) is left undeleted for reference.

PR #1230, opened 2026-04-28
Fall back to personal org when SSO response omits active_organization

vaimdevs:
```
##Fix
* Fall back to personal org when SSO response omits active_organization when new user register
```


lmeyerov:
closing in favor of https://github.com/graphistry/pygraphistry/pull/1228

Issue #1231, opened 2026-04-28
Duplicate 'source'/'target' columns cause pandas merge failure


dess890:
When an igraph edge attribute is named source or target, from_igraph() produces a DataFrame with duplicate column labels. Should be fixed upstream in the igraph conversion step.

Further context found here: https://github.com/graphistry/graphistry/pull/3083/changes/6ed57e52be52241dca37abd8eed17a3811c302d5

PR #1246, opened 2026-05-01
fix(igraph): source/target collision + ensure_pandas fallback

dess890:
Two small robustness fixes that surface when downstream callers drive compute_igraph 
1.` plugins/igraph.py from_igraph`: when an igraph edge attribute is named 'source' or 'target', ig.get_edge_dataframe() returns                    duplicate column labels and the endpoint-remapping merges crash with "column label 'source' is not unique".                                                                                                              
                                                                                                                                                                                                                     2.` compute/engine_coercion.py ensure_pandas`: only caught TypeError on the to_pandas(nullable=True) fallback. Newer cuDF (25.x) raises NotImplementedError for unsupported dtypes like datetime64[ms], which leaked through and crashed compute_igraph.                                                                                                

addresses: #1231


graphistry/compute/engine_coercion.py
```diff
@@ -200,15 +200,15 @@ def ensure_engine_match(g: Plottable, requested_engine: Engine) -> Plottable:
...
+    Tries nullable=True to preserve nullable integers; falls back to
+    plain to_pandas() if rejected (TypeError on older cuDF,
+    NotImplementedError on dtypes like datetime[ms]).
     """
     if isinstance(df, pd.DataFrame):
         return df
     try:
         return df.to_pandas(nullable=True)
-    except TypeError:
+    except (TypeError, NotImplementedError):
```
lmeyerov: we should probably log here fwiw (debug/info/warn?)

lmeyerov (approved):
@dess890  thanks -- see comment, + add chagnelog.md fix entry plz?

lmeyerov (approved):

lmeyerov:
@dess890 merged -- I'll work on publishing later today, and you can of course prototype from the PR branch pin until then

PR #1397, opened 2026-05-09
Security: Insecure JSON deserialization using assert statements

tomaioo:
## Summary

Security: Insecure JSON deserialization using assert statements

## Problem

**Severity**: `Critical` | **File**: `graphistry/compute/predicates/from_json.py:L38`

The from_json function in predicates/from_json.py uses assert statements for security-critical validation (type checking, dict validation). In Python, assert statements can be disabled with the -O flag, which would completely bypass these security checks and allow arbitrary code execution through malicious JSON input.

## Solution

Replace assert statements with proper if/raise validation. Example: if not isinstance(d, dict): raise ValueError('d must be a dict')

## Changes

- `graphistry/compute/predicates/from_json.py` (modified)

### DIFF
--- a/graphistry/compute/predicates/from_json.py
+++ b/graphistry/compute/predicates/from_json.py
@@ -33,12 +33,17 @@ type_to_predicate: Dict[str, Type[ASTPredicate]] = {
 }
 
 def from_json(d: Dict[str, JSONVal]) -> ASTPredicate:
-    assert isinstance(d, dict)
-    assert 'type' in d
-    assert d['type'] in type_to_predicate
-    assert isinstance(d['type'], str)
+    if not isinstance(d, dict):
+        raise ValueError('d must be a dict')
+    if 'type' not in d:
+        raise ValueError("d must have a 'type' key")
+    if not isinstance(d['type'], str):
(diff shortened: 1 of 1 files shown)


lmeyerov (commented):
Thanks for flagging this. I agree the "assert" / "python -O" concern is worth thinking through for runtime validation paths.

I do not think one-off assertion replacement PRs are the right approach here, though. Before changing code, I think we need a broader proposal covering:

- policy: where "assert" is vs is not acceptable
- scope: how to identify load-bearing/runtime validation asserts
- automation: Ruff and/or other tooling, CI under "python -O", regression prevention
- rollout: bounded audit vs automation-first cleanup sequence

I also want to be careful about severity framing. The current PR describes this as critical deserialization/RCE territory, while the actual change is a local assert replacement without a demonstrated exploit path. The underlying concern may still be valid, but I think the framing should match the demonstrated impact.

Rather than reviewing a sequence of narrow fix PRs with unclear coverage/end-state, I’d prefer a broader RFC/proposal issue for how this issue class should be handled across the project. A PR stack doing the same can also work.

I’m going to close this PR as too narrow for the stated issue, but I’d welcome that broader proposal.

PR #1648, opened 2026-06-25
feat(gfql): native lazy Polars engine — collect-once traversals + cypher row pipeline

lmeyerov:
## Summary

**Native CPU Polars execution engine for GFQL** (`Engine.POLARS`, opt-in via `engine='polars'`) — combines the former two-PR split (hop/chain traversals + cypher row pipeline) into one cohesive CPU-engine PR. The production pandas/cuDF paths are untouched; `engine='auto'` with Polars input still coerces to pandas as before.

> Note: this is the **CPU** Polars engine. A GPU engine (`cudf_polars`/`POLARS_GPU`) is a separate PR (#1655) that stacks on top of this. Stacks on #1652 (general perf opts).

### Traversals — `hop()` / `chain()`
Native vectorized BFS via semi/anti joins (no per-row Python). Forward/reverse/undirected single-hop, directed multi-hop chains, node/edge filter dicts and predicates (lowered to `pl.Expr`), `edge_match`/`source_node_match`/`destination_node_match`, `target_wave_front`, alias names. Deferred (honest `NotImplementedError`): variable-length/multi-hop edges, undirected edges in multi-edge chains, hop labels, node `query=`.

### Cypher row pipeline — `MATCH … RETURN`
**NO CHEATING:** every query runs natively on Polars or raises an honest `NotImplementedError` pointing at `engine='pandas'` — never a silent pandas bridge. Native: frame ops (rows/limit/skip/distinct/drop_cols), select/with_/return_ projection (cypher-expr-AST → `pl.Expr`: property/arithmetic/comparison/boolean/literal + `coalesce`/`abs`), `where_rows` (OR/NOT WHERE, Kleene 3-valued), order_by, group_by (count/sum/avg/min/max), unwind (literal cross-join), property/expr result projection, int/string/bool entity-text (`pl.concat_str`). Honestly deferred → NIE: cross-entity same-path WHERE, multi-entity binding_ops, float/temporal/nested entity-text, exotic exprs.

### Conformance hardening
Driven by the cross-repo Cypher TCK differential (pandas-vs-polars). Every fix either matches pandas natively or declines honestly (NO-CHEATING — no silent bridge, no wrong answer):
- **IEEE NaN comparison semantics** — `0.0/0.0 > 1` etc.: Polars treats NaN as the *largest* value; now masked to the IEEE/pandas answer (`& ~is_nan` for ordering/`=`, `| is_nan` for `<>`), driven by the lowered expression's **output dtype** (robustly covers int/int→float division + function results).
- **numeric-vs-string comparison** (`n.val > 'a'`) — would Rust-panic; now NIE (recurses into `AllOf`/`Between`, covers Categorical/Enum + arithmetic on all-null→String columns).
- **ISO-temporal ordering / temporal arithmetic** (`time({...}) > time({...})`, `a.time + duration({...})`) — were lexicographic/string-concat wrong answers; now NIE (narrowed to ordering of two temporal literals; `=`/`<>` stay native).
- **temporal-constructor-string property** + **heterogeneous (mixed-type) column** projections — NIE (clear message naming the column) instead of leaking raw constructor text / cryptic `ArrowInvalid`.
- **`UNION DISTINCT`** — engine-aware `Engine.df_unique` (Polars `unique(maintain_order=True)`) instead of the pandas-only `drop_duplicates` crash.
- **`OPTIONAL MATCH`** — absent whole-entity renders `null` (not `'()'`); null-row-fill alignment shape declines honestly (NIE) instead of a misleading validation error.
- **3-valued boolean over null literals** (`null AND null`, `NOT null`) — AND/OR/NOT cast to `pl.Boolean` so Kleene logic evaluates instead of raising.
- **label match on the reserved `labels` List column** — `list.contains` (membership) instead of a List→String cast crash; user List properties decline honestly.
- **predicate pandas-bridge removed** — an unlowerable `filter_by_dict` predicate now raises NIE (was silently `.to_pandas()`-bridging); native lowering widened (`AllOf`, `IsNull`/`NotNull`, case-insensitive `STARTS`/`ENDS WITH`).
- **`WITH`-scalar `MATCH` re-entry** — engine-aware (`pl.row`/`with_columns`) instead of a pandas `.iloc` crash.

`POLARS_ENGINES = (Engine.POLARS,)` is introduced here (the GPU target PR #1655 extends it) so the engine-aware helpers are self-contained at this layer.

## Validation
Differential parity vs the pandas engine (hop + chain suites + seeded fuzzer + a TCK-style cypher conformance lane with NULL/3-valued-logic + a `DEFERRED` list asserting deferred queries raise rather than bridge). Full `graphistry/tests/compute/gfql/` suite green (incl. the 1610-test cypher dir).

**Cross-repo Cypher TCK, polars arm:** the differential pandas-vs-polars lane is **clean — 0 wrong-answers** across the full TCK (every scenario either matches pandas or honestly declines; **387 honest declines**). This is the headline correctness guarantee: the Polars engine never silently disagrees with pandas.

## Perf (interleaved, 1M nodes, each engine on its native-frame graph, all native)
Polars wins **5.6–38×** across the surface: `RETURN n` ~38×, `ORDER BY` ~17×, traversals 6–7.5×, projections/aggregations/`DISTINCT` 5.6–6.9×. Plus eager fast paths (node-only / single-hop / unconstrained 1-hop) that move the polars>pandas crossover **below ~100K** for real viz/crossfilter shapes.

(Supersedes the former PR2 #1649, folded in here.)


lmeyerov:
Superseded by #1660 — this PR was auto-closed when its base branch `dev/gfql-opt-base` was deleted on the #1652 merge (mechanical side-effect, not merged). Same head branch `dev/gfql-polars-engine`, rebased onto master. Continue review at #1660.

PR #1649, opened 2026-06-25
feat(gfql): native Polars cypher row pipeline (PR2, stacked on #1648)

lmeyerov:
Stacked on #1648 (PR1: native Polars hop/chain traversals). **Base = `dev/gfql-polars-engine`** — review/merge after PR1.

## Summary
Second PR in the 3-PR stack: a **natively vectorized Polars cypher row pipeline**. Extends the polars engine from traversals (PR1) to the `MATCH … RETURN` row surface — selection, filtering, ordering, aggregation, unwind, distinct, limit/skip — running on polars expressions with **no pandas round-trip**.

**NO CHEATING (see plans/gfql-polars-engine):** the polars engine never silently falls back to the pandas engine. Every query either runs natively on polars or raises an honest `NotImplementedError` telling the caller to use `engine='pandas'`. Falling back to pandas would misrepresent the pandas engine as polars performance; only a human may consent to a bridge. This PR is about seeing how far honest native polars goes.

- `chain_polars` splits boundary `call()` ops (mirroring the pandas `_handle_boundary_calls`): traversal runs natively, then trailing row-pipeline calls run **per-op native, or raise `NotImplementedError`**.
- **Native polars entity-text projection**: whole-entity `RETURN n` renders `({prop: val, ...})` via `pl.concat_str` for int/string/bool single-entity nodes — full-table `RETURN n` goes from ~1.0x to ~38x. Float/temporal/nested/labels/multi-entity entity-text is **deferred** (raises NIE), not bridged.
- **Native polars lowering** (`engine_polars/row_pipeline.py`): a conservative cypher-expr-AST → `pl.Expr` lowering (property access, arithmetic/comparison/boolean, literals) drives native `select`/`with_`/`return_` projection, `order_by` (`.sort`), `group_by` (count/sum/avg/min/max), `unwind` (literal-list cross-join), and **`where_rows` filtering** for single-entity WHERE that doesn't fold into the matcher (OR/NOT). Result projection for property/expr columns is native (`.select`).
- **Native 3-valued WHERE**: cypher's `WHERE` keeps only TRUE rows (NULL and FALSE both dropped) — polars `DataFrame.filter` has exactly that semantics, and polars boolean `|`/`&` use Kleene logic, so NULL handling matches the pandas engine with no special-casing.
- **Honestly deferred** (raise `NotImplementedError`, no pandas fallback): cross-entity / same-path `WHERE`, multi-entity bindings (`RETURN n, m`), whole-entity entity-text over float/temporal/nested/label columns, and exotic expressions. These are the forward native-engineering targets for PR3.

## Correctness
- Differential parity vs pandas: `test_engine_polars_row_pipeline.py` + a broad **TCK-style differential conformance lane** `test_engine_polars_cypher_conformance.py` (curated **native-only** corpus + nullable/3-valued-logic graph + seeded fuzzer, every query run on both engines and asserted identical). A `DEFERRED` list explicitly asserts deferred queries raise `NotImplementedError` rather than silently bridging. The polars counterpart of the cross-repo Cypher TCK harness (graphistry/tck-gfql).
- Pandas/cuDF paths untouched; full gfql suite unregressed.
- mypy + ruff clean; changed-line-coverage + gfql coverage audit green.
- **dgx-spark (`--gpus all`): full gfql suite `2860 passed, 0 failed`; polars lane `376 passed, 29 skipped`.**

## Performance (`benchmarks/gfql/cypher_row_pipeline.py`, dgx-spark, CPU, 1M nodes / 5M edges)
Interleaved (≥9 reps; median ≈ min), each engine on a graph already in its native frame type. All workloads run **fully native** on polars (no bridge):

| workload | pandas_ms | polars_ms | speedup |
|---|---|---|---|
| RETURN n (whole-entity, int/str/bool props) | 3595 | 95 | **38x** |
| ORDER BY n.score DESC | 2082 | 120 | **17.3x** |
| WHERE + ORDER BY + LIMIT | 1265 | 90 | **14.0x** |
| RETURN n.score | 474 | 69 | **6.9x** |
| group_by sum/avg | 607 | 95 | **6.4x** |
| DISTINCT n.kind | 518 | 83 | **6.3x** |
| RETURN n.score, n.kind | 487 | 78 | **6.2x** |
| group_by count | 536 | 92 | **5.9x** |
| RETURN n LIMIT 10 | 466 | 83 | **5.6x** |

## Stack
- PR1 (#1648) — core traversals (hop/chain).
- **PR2 (this)** — cypher row pipeline, natively vectorized in polars; deferred features raise honest `NotImplementedError`.
- PR3 — pandas/polars/cuDF/kuzu out-of-the-box comparison on pyg-bench + forward native engineering (native cross-entity WHERE, multi-entity bindings, float/temporal entity-text).


lmeyerov:
Folded into #1648 (auto-marked merged: these commits now live directly in #1648's branch). The hop/chain traversals + this cypher row pipeline are one cohesive **CPU** Polars engine, so they're a single PR now. Full gfql suite 2885 passed, 0 failed. A future GPU/cudf-polars engine will stack on #1648.

Issue #1650, opened 2026-06-26
GFQL: avoid spurious entity-text stringification of returned entities (return structured/Arrow frames)


lmeyerov:
## Summary

GFQL's Cypher result path materializes node/edge entities as **Cypher *display strings*** ("entity-text", e.g. `(:Label{id: 1, val: 61, kind: 'b'})`) instead of as **structured/Arrow values**. Benchmarks show this row-wise stringification is the single largest framework cost for whole-entity returns — and worse, it runs **spuriously in paths whose output never contains entity-text at all** (e.g. `where_rows`). This issue proposes design-level fixes so we stop stringifying entities we don't need to.

## Evidence

Benchmarks on a quiet box, synthetic graph (nodes = edges/10), median-of-5 wall-clock; per-layer split via cProfile self-time. Full data: variant ladder + layer decomposition (`MATCH ... RETURN ...` vs direct API).

**1. `RETURN a` returns a single string column, not a node table.**
```python
g.gfql([n({"val": gt(50)})])                       # -> (980, 3) node TABLE: id, val, kind   (cheap)
g.gfql("MATCH (a) WHERE a.val>50 RETURN a")         # -> (980, 1) STRINGS: "({id:1, val:61, kind:'b'})"
g.gfql("MATCH (a) WHERE a.val>50 RETURN a.val")     # -> (980, 1) plain val column            (cheap)
```
Entity-text rendering (str-render + temporal/escape regex) is the dominant layer of `RETURN a`:

| engine | `RETURN a` wall @1M | entity-text share | engine-compute share |
|---|---:|---:|---:|
| pandas | 181 ms | **56%** | 36% |
| cudf | 100 ms | 27% | 48% |
| polars | 27 ms | **~0%** (native `pl.concat_str`) | 46% |

`RETURN a.val` (no entity-text) is ~6× cheaper than `RETURN a` on pandas (28 vs 181 ms @1M).

**2. Spurious stringification: `where_rows` renders entity-text it then discards.**
[...]

PR #1651, opened 2026-06-26
perf(gfql): dtype-gate temporal-text detection to avoid spurious stringification (#1650)

lmeyerov:
## Summary

Fixes the **spurious-stringification half of #1650**. GFQL ran row-wise entity/temporal stringification in comparison paths whose output never contains entity-text (e.g. `where_rows`). Root cause: `order_detect_temporal_mode` (`ordering.py`) runs `astype(str)` + up to **6 regex `fullmatch` passes** on **both operands of every comparison** — including `int64`/`bool` columns that can never hold temporal *text*. A 10k-row `where_rows(val > 50)`, which returns a plain 3-column table, fired **~160,006 spurious `re.fullmatch`** calls, all discarded.

> Note: the where_rows cost is `order_detect_temporal_mode`, **not** `entity_text.py`/`entity_props.py` as the issue body speculated (cProfile-confirmed), and the issue's ~10× over-render does **not** reproduce — entity-text already renders on the deduped node table. This PR targets the confirmed spurious path; whole-entity `RETURN a` structured/Arrow returns are tracked separately.

## Fix

`order_detect_temporal_mode` early-returns `None` when `series.dtype.kind in {i, u, f, b, c}` (numeric/bool/complex). Object/string/datetime columns are unaffected, so temporal-text detection is preserved.

## Results (byte-identical output, gated == ungated)

| `where_rows(val>50)` | before | after | speedup |
|---|--:|--:|--:|
| pandas @10k | 48.9 ms | 15.6 ms | **3.1×** |
| cuDF @10k | 37.4 ms | 8.5 ms | **4.4×** |
| cuDF @100k (1M edges) | 191.0 ms | 14.4 ms | **13.3×** |

Speedup scales with row count (spurious scan is O(rows)). cuDF measured on a GB10 box with `graphistry/test-gpu:latest` (cudf 25.12).

## Tests

- pandas: temporal + ordering + row suites — **91 pass**
- cuDF (`TEST_CUDF=1` container): temporal + ordering + row suites — **83 pass**
- Added: `order_detect_temporal_mode` unit coverage (numeric/bool skip; object temporal still detected) + end-to-end numeric `where_rows` filter assertion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Development]
 <!-- Do Not Erase This Section - Used for tracking unreleased changes -->
 
+### Performance
+- **GFQL temporal-detection dtype gate (#1650)**: `order_detect_temporal_mode` now short-circuits for numeric/bool/complex columns, which can never hold temporal *text*, instead of running an `astype(str)` + multi-regex `fullmatch` scan on every comparison. Eliminates spurious row-wise stringification in `where_rows`/comparison paths whose output never contains entity-text. Byte-identical results; measured `where_rows` speedups ~3.1× (pandas) and ~4.4–13.3× (cuDF, scaling with row count). Does not address whole-entity `RETURN a` text rendering, which is tracked separately.
+
 ## [0.56.1 - 2026-05-27]
 
 ### Added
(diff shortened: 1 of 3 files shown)


lmeyerov:
Folded into the PR0 general-optimizations base (#1652, commit `b539212c`, original authorship preserved) so it sits beneath the polars-engine stack (#1648 → #1649). Validated there: full gfql suite 2883 passed, 0 failed. Superseded by #1652.

PR #1654, opened 2026-06-26
feat(gfql): Polars-GPU engine (engine='polars-gpu', cudf_polars) [PR3, stacked on #1648]

lmeyerov:
## Summary

**GPU execution mode of the native Polars engine** (`Engine.POLARS_GPU`, opt-in via `engine='polars-gpu'`) — turns the validated PR4 spike/loop-probe into a formal engine, **stacked on the CPU Polars engine (#1648)**. The same vectorized ops run, but the hot traversal joins materialize on GPU via the RAPIDS **cudf_polars** backend (`LazyFrame.collect(engine=pl.GPUEngine(raise_on_fail=False))`).

## Design (contained execution mode — not a frame type)
- `Engine.POLARS_GPU = 'polars-gpu'`, **explicit opt-in only** — `engine='auto'` never selects it.
- Frames stay `pl.DataFrame` (handled exactly like `POLARS` in all frame ops via `POLARS_ENGINES`); only the **collect boundary** changes.
- GPU intent is carried by a **context var** (`engine_polars/gpu.py`) set at the chain/hop dispatch boundary, so the engine internals don't thread a `gpu` flag. When GPU is inactive the join helper is the *ordinary eager join* → **`engine='polars'` (CPU) is byte-for-byte unchanged**.
- `raise_on_fail=False` keeps any GPU-incapable node on CPU **in Polars** — NOT a pandas bridge (still honest/native; NO-CHEATING).

## Scope (first slice)
GPU-routes the hop semi-joins + the final edge/node materialization (the loop-probe's proven win). The cypher row-pipeline ops run on CPU-Polars for now (still correct, still native) — extending GPU coverage there is the next increment.

## Validation (dgx, RAPIDS container `--gpus all`)
- Differential parity **`engine='polars-gpu'` == `engine='polars'`** across the cypher conformance corpus + core traversals — `test_engine_polars_gpu.py`, **36 passed** (skips when no cudf_polars/GPU). GPU semi-join confirmed running on GPU (`raise_on_fail=True` probe).
- Full `graphistry/tests/compute/gfql/` suite unchanged: **2885 passed, 0 failed** (`engine='polars'` untouched).

Stacks on #1648 (CPU Polars engine) → #1652 (general opts) → master.


lmeyerov:
Superseded by the lazy-GPU PR (GPU as a *target* of the lazy engine on #1648). The per-op approach here was a perf regression (repeated H2D); the lazy collect-once version is a 2.84× single-hop GPU win @1M with CPU parity. Branch kept for reference.

PR #1658, opened 2026-06-28
feat(gfql): physical adjacency indexes for O(degree) seeded traversal

lmeyerov:
## Stack position
Stacks on **#1667** (polars-engine followups). Full spine:
`master → #1666 (3VL) → #1660 (polars engine + GPU target; #1655 folded in) → #1667 (followups) → **this PR**`.
Restacked 2026-07-02: the followups PR proved independent of this index layer, so the most
experimental PR (this one) moved to the top of the spine — reviewable against the final
engine surface its hooks land on. Additive — no change to default (un-indexed) behavior
on any engine.

## What this does
Opt-in, **pay-as-you-go** CSR adjacency / node-id indexes → seeded traversal becomes
**O(degree)**, flat in graph size. Sidecar over edge **row positions** (never reorders
`.edges`/`.nodes`), fingerprint-validated by object IDENTITY (a `.edges()` rebind safely
invalidates → scan fallback, never a wrong answer). Three uniform surfaces (Python /
Cypher DDL `CREATE/DROP/SHOW GFQL INDEX` / JSON wire) + `gfql(..., index_policy=...)`.
Engine-polymorphic: numpy host arrays for pandas/polars, **cupy on-device for cudf**.
Hooked at every seeded scan site; falls back to scan for uncovered features
(edge/source/dest match, `target_wave_front`, `min_hops>1`, labeling, missing/duplicate
node ids, missing endpoints).

## ⚠️ Correctness hardening (takeover)
A 2-agent adversarial review reproduced **wrong-answer divergences from the un-indexed
scan oracle** that the original parity scenarios missed (the cost gate masks them unless
≥3 distinct seed keys hit the index path). All fixed — the index now matches scan or
falls back, never a wrong answer (+6 engine-parametrized differential regression tests):
- **`max_hops` was silently ignored** (`hop(max_hops=3)` ran 1 hop) → now honored.
- **duplicate node ids corrupted the node-id index** → first-row-per-key + refuse/fallback.
- **edge endpoints absent from the node table were dropped** (scan synthesizes them) → fallback.
- node-row order, `id(df)` fingerprint recycling, int32/int64 seed narrowing, and a
  polars/cudf seed-conversion crash under `force` → all fixed.
- review wave: polars `is_in` deprecation → order-preserving semi-join; `DROP GFQL INDEX
  <custom-name>` silent no-op → resolve-by-name + raise.

## Results (dgx-spark, deg-8, warm median) — OUR reproduced numbers
[...]

PR #1660, opened 2026-06-29
feat(gfql): native lazy Polars engine — collect-once traversals + cypher row pipeline + followups (multi-hop, to_fixed_point, undirected, more predicates, NIE→native)

lmeyerov:
> **Note:** Supersedes #1648, which GitHub auto-closed when its base branch `dev/gfql-opt-base` was deleted on the #1652 merge (the close was a mechanical side-effect, not a rejection — `mergedAt=null`). This is the same head branch `dev/gfql-polars-engine`, rebased onto `master` (the merged #1652/#1657/#1656 commits dropped). Reopen was not possible (base branch gone + head rebased).

---

## Summary

**Native CPU Polars execution engine for GFQL** (`Engine.POLARS`, opt-in via `engine='polars'`) — combines the former two-PR split (hop/chain traversals + cypher row pipeline) into one cohesive CPU-engine PR. The production pandas/cuDF paths are untouched; `engine='auto'` with Polars input still coerces to pandas as before.

> Note: this PR now includes the **GPU execution target** (`engine='polars-gpu'`, cudf_polars) — former stacked PR #1655 was folded in by fast-forward (GitHub shows it merged): the engine PR already contained the lazy collect-once machinery the GPU target completes ("one engine, two targets"), so the boundary cut a single feature in half. **Also folds the former #1667** (native followups — see the Followups section). Base: #1666 (3VL null-semantics fix).

### GPU target — `engine='polars-gpu'` (folded #1655)
The same lazy plan, executed on the RAPIDS cudf_polars backend via ONE `collect_all` per hop (collect-once is what makes GPU pay off: per-op eager GPU collect was a measured regression). Explicit opt-in only (AUTO never selects it). **GPU-or-error** (`raise_on_fail=True`): a GPU-incapable plan node raises (pointing at `engine='polars'`) — never silent-CPU mislabeled as GPU. Includes opt-in CPU streaming collect (`GFQL_POLARS_CPU_STREAMING=1`) for large batch traversals. Validated by `engine='polars-gpu' == engine='polars'` differential parity across the conformance corpus + traversals (2.84× single-hop GPU win @1M vs CPU).

### Followups — native completions (folded #1667)
The former stacked PR #1667 was folded in (GitHub shows it merged) — it *completes the native surface* rather than adding a separable feature:
- **Native variable-length `min_hops>1`** forward/reverse traversals (no pandas bridge) — the subtle min_hops NODE-output rule proven against the pandas oracle (500/500 chain fuzz on both polars and polars-gpu).
- **Native `get_degrees` / `get_indegrees` / `get_outdegrees`** (parity vs `ComputeMixin`), incl. via the Cypher `CALL` surface.
- **Wider native cypher coverage**: `toFloat`, `collect`/`collect(DISTINCT)`, `WHERE … IN`, `size`/`substring`, `to_fixed_point`, more predicates — each parity-validated or honest-NIE.
- **`count(*)` short-circuit**: a lone `RETURN count(*)` reads table height (or sums a boolean mask) in one reduction instead of materializing + group_by — engine-polymorphic across all four engines.

### Traversals — `hop()` / `chain()`
Native vectorized BFS via semi/anti joins (no per-row Python). Forward/reverse/undirected single-hop, directed multi-hop chains, node/edge filter dicts and predicates (lowered to `pl.Expr`), `edge_match`/`source_node_match`/`destination_node_match`, `target_wave_front`, alias names. Deferred (honest `NotImplementedError`): variable-length/multi-hop edges, undirected edges in multi-edge chains, hop labels, node `query=`.

### Cypher row pipeline — `MATCH … RETURN`
**NO CHEATING:** every query runs natively on Polars or raises an honest `NotImplementedError` pointing at `engine='pandas'` — never a silent pandas bridge. Native: frame ops (rows/limit/skip/distinct/drop_cols), select/with_/return_ projection (cypher-expr-AST → `pl.Expr`: property/arithmetic/comparison/boolean/literal + `coalesce`/`abs`), `where_rows` (OR/NOT WHERE, Kleene 3-valued), order_by, group_by (count/sum/avg/min/max), unwind (literal cross-join), property/expr result projection, int/string/bool entity-text (`pl.concat_str`). Honestly deferred → NIE: cross-entity same-path WHERE, multi-entity binding_ops, float/temporal/nested entity-text, exotic exprs.

### Conformance hardening
Driven by the cross-repo Cypher TCK differential (pandas-vs-polars). Every fix either matches pandas natively or declines honestly (NO-CHEATING — no silent bridge, no wrong answer):
- **IEEE NaN comparison semantics** — `0.0/0.0 > 1` etc.: Polars treats NaN as the *largest* value; now masked to the IEEE/pandas answer (`& ~is_nan` for ordering/`=`, `| is_nan` for `<>`), driven by the lowered expression's **output dtype** (robustly covers int/int→float division + function results).
- **numeric-vs-string comparison** (`n.val > 'a'`) — would Rust-panic; now NIE (recurses into `AllOf`/`Between`, covers Categorical/Enum + arithmetic on all-null→String columns).
- **ISO-temporal ordering / temporal arithmetic** (`time({...}) > time({...})`, `a.time + duration({...})`) — were lexicographic/string-concat wrong answers; now NIE (narrowed to ordering of two temporal literals; `=`/`<>` stay native).
- **temporal-constructor-string property** + **heterogeneous (mixed-type) column** projections — NIE (clear message naming the column) instead of leaking raw constructor text / cryptic `ArrowInvalid`.
- **`UNION DISTINCT`** — engine-aware `Engine.df_unique` (Polars `unique(maintain_order=True)`) instead of the pandas-only `drop_duplicates` crash.
- **`OPTIONAL MATCH`** — absent whole-entity renders `null` (not `'()'`); null-row-fill alignment shape declines honestly (NIE) instead of a misleading validation error.
- **3-valued boolean over null literals** (`null AND null`, `NOT null`) — AND/OR/NOT cast to `pl.Boolean` so Kleene logic evaluates instead of raising.
- **label match on the reserved `labels` List column** — `list.contains` (membership) instead of a List→String cast crash; user List properties decline honestly.
- **predicate pandas-bridge removed** — an unlowerable `filter_by_dict` predicate now raises NIE (was silently `.to_pandas()`-bridging); native lowering widened (`AllOf`, `IsNull`/`NotNull`, case-insensitive `STARTS`/`ENDS WITH`).
- **`WITH`-scalar `MATCH` re-entry** — engine-aware (`pl.row`/`with_columns`) instead of a pandas `.iloc` crash.

`POLARS_ENGINES` (POLARS + POLARS_GPU) is introduced here so the engine-aware helpers are self-contained at this layer.

## Validation
Differential parity vs the pandas engine (hop + chain suites + seeded fuzzer + a TCK-style cypher conformance lane with NULL/3-valued-logic + a `DEFERRED` list asserting deferred queries raise rather than bridge). Full `graphistry/tests/compute/gfql/` suite green (incl. the 1610-test cypher dir).

**Cross-repo Cypher TCK, polars arm:** the differential pandas-vs-polars lane is **clean — 0 wrong-answers** across the full TCK (every scenario either matches pandas or honestly declines; **387 honest declines**). This is the headline correctness guarantee: the Polars engine never silently disagrees with pandas.

## Perf (interleaved, 1M nodes, each engine on its native-frame graph, all native)
Polars wins **5.6–38×** across the surface: `RETURN n` ~38×, `ORDER BY` ~17×, traversals 6–7.5×, projections/aggregations/`DISTINCT` 5.6–6.9×. Plus eager fast paths (node-only / single-hop / unconstrained 1-hop) that move the polars>pandas crossover **below ~100K** for real viz/crossfilter shapes.

## Recent refinements (review-driven)
Post-fold polish — ergonomics, correctness, and typing:
- **cuDF→polars conversion is dtype-lossless**: converts via **Arrow** (cuDF's native interchange) instead of cuDF→pandas→polars, which double-converted **and was lossy** (nullable `Int64`→`float64`+NaN). Verified on dgx; also removes a device→host→device round trip for `polars-gpu`.
- **`validate`/`warn` convention on engine conversion**: `Engine.df_to_engine(..., validate=, warn=)` — `strict` (polars default) raises on an un-representable mixed-type column, `autofix` coerces+warns (matching cuDF + the `plot()`/`upload()` boundary). Both `Engine.py` from_pandas converters unified onto the protocol.
- **Polars execution config is Python-settable + live**: `set_cpu_streaming` / `set_gpu_executor` (+ the public `GPU_EXECUTORS` options) — previously env-only (`GFQL_POLARS_*`) and frozen at import; documented in the engine-selection guide (docs PR).
- **Typing + dead-code sweep**: removed the `getattr(g, "_nodes", None)` no-op anti-pattern on **declared** `Plottable` attrs (result_postprocess / reentry / row pipeline / `gfql_unified` / call procedures); typed the polars-engine dtype/expr helpers (`pl.DataType` / `pl.Expr` / `ExprNode` instead of `Any`); removed verified-redundant `cast()`s. (The systemic `Plottable._nodes: Any` root — ~517 casts repo-wide — is filed separately as #1678, out of scope here.)
- **Housekeeping**: degree helpers moved out of `chain.py` into `degrees.py`; `bin/test-polars.sh` de-duplicated (one test-file list, coverage via env toggle).

- **Off-engine `call()` analytics under Polars — `call_mode='auto'` (default) / `'strict'`**: a `call()` running a whole-graph analytic with **no Polars kernel** (`umap`, `hypergraph`, `compute_cugraph`/`compute_igraph`, `*_layout`, `collapse`, …) previously hard-raised `NotImplementedError` under a Polars engine — forcing users off Polars for the whole pipeline. It now runs as a **mode-gated, warned modality switch**: `auto` bridges off-engine (`polars`→pandas, `polars-gpu`→**cuDF on-device**), runs the analytic, and coerces the result back to Polars **losslessly via Arrow** (warn once per analytic); `strict` keeps the parity-or-`NotImplementedError` decline (benchmark integrity / memory ceiling). `polars-gpu` is **GPU-or-error** (declines rather than silently dropping to host). **Traversal / filter / row ops stay parity-or-NIE** — the split is mechanical (`is_row_pipeline_call`), and the chain and DAG surfaces bridge consistently. Python-settable (`set_call_mode`) + env `GFQL_POLARS_CALL_MODE`. dgx-verified: `compute_cugraph` PageRank byte-parity with the pandas oracle under both `polars` and `polars-gpu`; full polars CPU lane green. Documented in the engine guide (docs PR).
- **More typing (review batch)**: typed `lazy/engine/polars/predicates.py` (`pl.Expr`/`pl.DataType` + a documented scalar union; audited every `is`/`getattr`); a `RowPipelineCtx` `Protocol` replaces `ctx: Any` in `row/frame_ops.py` (checked interface, zero runtime); explicit `__all__` on the `lazy` public surface.

(Supersedes the former PR2 #1649, folded in here.)

## Review notes
- History: many conventional commits across the engine + folded #1655/#1667 + review-driven refinements (repo lands merge commits); an early-PR module rename (`engine_polars/` → `lazy/engine/polars/`) means the first commits touch pre-rename paths — **review by diff**, not commit-by-commit.
- Team-polish pass applied (commit `fe8218c5`): stale rename-era docstring paths fixed, internal plan/`NO-CHEATING` citations scrubbed from user-facing error strings ("no pandas fallback; parity-or-error by design"), `Engine.py` error strings truthful about polars support, inline engine tuples → `POLARS_ENGINES`.
- Follow-ups (non-blocking, tracked): the repo-wide `Plottable._nodes` typing (#1678); unify the 3 mixed-type coercion impls under one protocol-aware helper.
- **Cleanup pass (2026-07-04, commits `4db768a9..33908cde`) — zero-behavior refactor, verified**: consolidated the previously-tracked duplication (shared `endpoint_ids()`/`col_dtype()` in `dtypes.py`, one `_project_polars` body for select/with_columns, `_endpoint_counts` in degrees, `_alias_true_mask` in frame_ops, comparison-op whitelists replacing 6-way operator ladders, degree-dispatch collapse) and re-encoded the test suite without shrinking coverage (shared `polars_test_utils.py`; conformance-ledger axes and the matrix's native-ok/honest-NIE families are table-driven with per-case ids; one chain-parity body with per-table strictness flags; +1 pandas-oracle canary; the chain-vs-DAG surface check is now STRICTER — a non-NIE error on either surface fails). Evidence: dgx full lane + GPU parity 1703 passed at every batch (collected conformance cases 231→232, none lost); polarity/waiver mutation checks fail loudly; ruff+mypy clean. Diff 8,515→8,220 insertions. A follow-up owner-directed **prose pass** (commit `f862657e`) tightened the ~1,745 comment/docstring lines (NO-CHEATING contract once per module, per-site one-liners, every fact/citation/perf number kept) — mechanically gated as code-identical (AST-equal with docstrings stripped) + the same dgx lane (1703 passed). **Final diff: 7,747 insertions** (prod ~3,850 / tests 3,361 / benchmarks+infra 534). A further experiment-gated refactor (commit `b956d115`) unified the lazy single-hop into `hop_polars` — ONE hop implementation instead of the twin `hop.py`/`hop_eager.py` pair — proven by a 3-round interleaved A/B bench at 1M+10M edges (within-noise-or-better on all workloads; the lazy plan boundary is a measured perf contract) + full parity lane (1703) + the #1658 index-hook stack re-verified (1724+68). Follow-up typing pass (`7acdd411`) fully annotates the hop helpers (surfacing and fixing a latent lazy-vs-eager variable name collision). CHANGELOG: unchanged by design — the cleanup/unification is zero-behavior, and the existing feature entries (incl. the collect-once single-hop + eager multi-hop design they describe) remain accurate.


lmeyerov:
**Scope update (2026-07-03):** folded #1667 into this PR at Leo's request so the native polars engine and its parity followups are reviewed as one combined PR. This branch (`dev/gfql-polars-engine`, tip `ba8d360c`) now contains both — combined diff ~7.6K insertions / 40 files / 69 commits. #1658 (adjacency index) retargets onto this PR; #1675 (viz cypher ops) still forks off this PR, its own diff unchanged. Note for merge time: this branch carries a duplicate 3VL CHANGELOG entry (near-identical to the one already landed via #1666) — dedup to one when this merges.

bin/test-polars.sh
```diff
@@ -10,4 +10,10 @@ set -ex
...
 
 python -B -m pytest -vv \
-    graphistry/tests/compute/test_polars.py
+    graphistry/tests/compute/test_polars.py \
+    graphistry/tests/compute/gfql/test_engine_polars_hop.py \
+    graphistry/tests/compute/gfql/test_engine_polars_chain.py \
+    graphistry/tests/compute/gfql/test_engine_polars_row_pipeline.py \
+    graphistry/tests/compute/gfql/test_engine_polars_cypher_conformance.py \
+    graphistry/tests/compute/gfql/test_engine_polars_conformance_matrix.py \
+    graphistry/tests/compute/gfql/test_conformance_ledger.py
```
lmeyerov, line 22: DRY w/ ci.yml

graphistry/Engine.py
```diff
@@ -209,17 +221,92 @@ def df_to_engine(df, engine: Engine):
...
+        if isinstance(df, pl.DataFrame):
+            return df
+        if isinstance(df, pl.LazyFrame):
+            return df.collect()
+        if isinstance(df, pa.Table):
+            return pl.from_arrow(df)
+        if isinstance(df, pd.DataFrame):
+            return _pl_from_pandas(df)
+        # cudf/dask/spark and anything else: route through pandas first
+        return _pl_from_pandas(df_to_engine(df, Engine.PANDAS))
```
lmeyerov: avoid D->H->D round trip

graphistry/Engine.py
```diff
@@ -209,17 +221,92 @@ def df_to_engine(df, engine: Engine):
...
+    cryptic ``pyarrow.lib.ArrowInvalid`` from deep inside construction, raise a
+    clear ``NotImplementedError`` pointing at ``engine='pandas'`` (NO-CHEATING: no
+    silent string-coercion, which would change comparison semantics)."""
+    import polars as pl
+    try:
+        return pl.from_pandas(df)
+    except Exception as e:
+        bad = _mixed_type_object_columns(df)
+        hint = f" (mixed-type column(s): {bad})" if bad else ""
+        raise NotImplementedError(
```
lmeyerov, line 377: follow strictness setting

graphistry/compute/gfql/lazy/engine/polars/chain.py
lmeyerov, line 1: most of these should be polymorphic in `graphistry/compute/gfql/lazy/*` not `graphistry/compute/gfql/lazy/engine/polars/*`

graphistry/compute/gfql/lazy/engine/polars/chain.py
```diff
@@ -0,0 +1,851 @@
...
+    # Drop a pre-existing degree column so a re-run overwrites rather than producing a
+    # ``*_right`` join collision (mirrors the pandas keep-subset).
+    base = nodes.drop(col) if col in _colnames(nodes) else nodes
+    out = base.join(counts, on=node_col, how="left").with_columns(
+        pl.col(col).fill_null(0).cast(pl.Int32)
+    )
+    return g.nodes(out, node_col)
+
+
+def get_indegrees_polars(g: Plottable, col: str = "degree_in") -> Plottable:
```
lmeyerov: factor out the regular helpers

graphistry/compute/gfql/cypher/reentry/execution.py
```diff
@@ -215,8 +243,8 @@ def compiled_query_reentry_state(
         non_null_mask = cast(SeriesT, prefix_alias_values.notna())
         has_non_null_ids = bool(non_null_mask.any()) if hasattr(non_null_mask, "any") else True
         if not has_non_null_ids:
-            base_nodes = cast(Optional[DataFrameT], getattr(base_graph, "_nodes", None))
-            id_column = cast(Optional[str], getattr(base_graph, "_node", None))
+            base_nodes = cast(Optional[DataFrameT], base_graph._nodes)
+            id_column = cast(Optional[str], base_graph._node)
```
lmeyerov: avoid casts

graphistry/compute/gfql/lazy/engine/polars/dtypes.py
```diff
@@ -0,0 +1,55 @@
...
+are numeric, float, or string-like), used by the predicate lowering, the expression
+lowering, and the result projection. Keeping ONE definition avoids the guards
+silently diverging when a polars dtype is added or a classification is fixed at one
+site only. Polars is imported lazily (optional dependency), matching the engine's
+convention.
+"""
+from typing import Any
+
+
+def is_int(dt: Any) -> bool:
```
lmeyerov: Try to replace `dt: Any` with typed versions..

graphistry/compute/gfql/lazy/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+    polars error) instead of a cryptic ComputeError, pointing at the CPU engine."""
+    return NotImplementedError(
+        "GFQL engine='polars-gpu': this query plan is not fully GPU-executable on the "
+        "installed cudf-polars (NO-CHEATING: we raise rather than silently run it on CPU "
+        "and call it a GPU result). Rerun with engine='polars' for native CPU execution. "
+        f"Underlying GPU error: {type(exc).__name__}: {exc}"
+    )
+
+
+def collect(lf: Any) -> Any:
```
lmeyerov: avoid Any's here

.github/workflows/ci.yml
```diff
@@ -1073,7 +1079,8 @@ jobs:
         python -m coverage combine \
           --data-file=build/changed-line-coverage/combined.coverage \
           build/changed-line-coverage/artifacts/core/coverage.core-py3.14 \
-          build/changed-line-coverage/artifacts/gfql/coverage.gfql-py3.12
+          build/changed-line-coverage/artifacts/gfql/coverage.gfql-py3.12 \
+          build/changed-line-coverage/artifacts/polars/coverage.polars-py3.12
```
lmeyerov, line 1083: combine gfql + polars coverage

graphistry/compute/gfql/call/executor.py
```diff
@@ -147,6 +168,23 @@ def execute_call(g: Plottable, function: str, params: Dict[str, Any], engine: En
...
+            if engine in (Engine.POLARS, Engine.POLARS_GPU):
+                _res_nodes = getattr(result, '_nodes', None)
+                _res_edges = getattr(result, '_edges', None)
+                _probe = _res_nodes if _res_nodes is not None else _res_edges
+                if _probe is not None and 'polars' not in type(_probe).__module__:
+                    raise NotImplementedError(
+                        f"GFQL engine='{engine.value}' does not natively support call "
+                        f"'{function}'; it runs on pandas/cuDF and coercing the result back to "
+                        "polars would be a silent bridge. Use engine='pandas'."
+                    )
```
lmeyerov: revisit - hypergraph, umap, etc too key

graphistry/compute/gfql/lazy/engine/polars/predicates.py
```diff
@@ -0,0 +1,382 @@
+"""Vectorized polars filter_by_dict for the native polars GFQL engine.
+
+Predicates are lowered to native polars expressions (no pandas round-trip) for
```
lmeyerov: weak typing + suspicious uses of `is`

graphistry/compute/gfql/lazy/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+# NOTE: these overrides are PROCESS-GLOBAL (plain module state), unlike the per-collect
+# execution TARGET (_TARGET is a ContextVar because target_mode sets it transiently). They are
+# meant as a global default set once at startup, not a per-request/per-thread override — a
+# concurrent set_cpu_streaming() affects all in-flight GFQL queries in the process.
+_cpu_streaming_override: Optional[bool] = None
+_gpu_executor_override: "Optional[GpuExecutor]" = None
+
+#: Valid cudf-polars GPU collect executors (public; see :func:`set_gpu_executor`).
+GPU_EXECUTORS = ("in-memory", "streaming")
+GpuExecutor = Literal["in-memory", "streaming"]
```
lmeyerov, line 70: audit for polars models/types to factor out as part of cleaner code + fewer circ dep risks?

graphistry/compute/gfql/lazy/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+    The streaming executor benchmarks faster + more stable than the default collect on
+    big traversal joins (isolated 80M-edge 2-hop semijoin 1669→1040 ms, ~1.6×; end-to-end
+    chain dilutes to ~1.04–1.11× as forward/backward/combine overhead is unaffected),
+    parity-identical. Opt-in (default off) because small/interactive sizes REGRESS
+    (~0.86× at 100K) from streaming overhead.
+    Resolution: :func:`set_cpu_streaming` override > ``$GFQL_POLARS_CPU_STREAMING`` (``"1"``) > ``False``.
+    """
+    if _cpu_streaming_override is not None:
+        return _cpu_streaming_override
+    return _os.environ.get("GFQL_POLARS_CPU_STREAMING", "0") == "1"
```
lmeyerov, line 84: memoize? unsure how hot

graphistry/compute/gfql/row/pipeline.py
```diff
@@ -1409,7 +1426,20 @@ def _gfql_eval_expr_ast(self, table_df: Any, node: Any) -> Tuple[bool, Any]:
...
+                    is_cudf = type(inner).__module__.startswith("cudf")
+                    if is_float and is_cudf:
+                        # libcudf float->string does not match Python/pandas float repr
+                        # (scientific-notation thresholds, exponent formatting) -> a silent
+                        # cross-engine divergence (issue #1663 finding 2; polars declines this
+                        # too). Round-trip the float column through the pandas oracle on the host
+                        # (arrow -> pandas float -> str = the exact pandas-engine result) so cuDF
+                        # is string-IDENTICAL to pandas. Niche op; one host transfer of one column.
+                        host_str = pd.Series(inner.to_arrow().to_pylist()).astype(str)
+                        out = s_cons(Engine.CUDF)(host_str.to_numpy(), index=inner.index)
```
lmeyerov, line 1457: do we need a Fast / Engine-Native mode, and a perf hinter mode?

graphistry/Engine.py
```diff
@@ -7,24 +7,38 @@
...
+# the same lazy Polars engine — frames stay ``pl.DataFrame``, so it shares the path.
+POLARS_ENGINES = (Engine.POLARS, Engine.POLARS_GPU)
 
 class EngineAbstract(Enum):
     PANDAS = Engine.PANDAS.value
     CUDF = Engine.CUDF.value
     DASK = Engine.DASK.value
     DASK_CUDF = Engine.DASK_CUDF.value
+    POLARS = Engine.POLARS.value
+    POLARS_GPU = Engine.POLARS_GPU.value
```
lmeyerov, line 35: how do they pick streaming, in-memory, etc?

PR #1672, opened 2026-07-01
feat(gfql): enrich gfql_explain with planner cost diagnostics (LP1)

lmeyerov:
## What

Stacked on the #1658 CSR adjacency-index PR. Makes `gfql_explain` (and the `index_trace()` per-hop records) explain **why** the seeded-hop planner chose the index or the scan — not just a `used_index` yes/no.

Per-hop step now carries:
- `frontier_n` — seed count
- `n_keys` — distinct source keys in the adjacency index
- `seed_deg_sum` / `est_result_rows` — a **free Σ-seed-degree fanout estimate**, read straight from the CSR `group_offsets` already on the index (no extra scan)
- `threshold_frac` — the engine's cost-gate crossover fraction
- `decision_reason` — human-readable, recorded at **every** bail point (`policy=off`, `no resident index (policy=use)`, `query not index-coverable`, `graph missing node/edge columns`, `no index available (build declined)`, `frontier N >= frac*n_keys (…) -> scan cheaper`) plus the final `-> index` / `index path not applicable -> scan`

Top-level `gfql_explain` return also lifts `est_seed_cardinality`, `est_result_rows`, `chosen_direction`, `decision_reason`.

## Why

The indexing/optimization roadmap calls out planner explainability (estimated cardinality, chosen direction, why-index-vs-scan) as the #1 next lever after the index itself. This is the smallest, safest slice of that.

## Safety / performance

Purely **additive and diagnostic**. The fanout estimate is computed **only** inside an `index_trace()`/`gfql_explain` context (`_trace_active()` gate), so the hot traversal path builds no diagnostic dict and computes no degrees — **zero hot-path cost, traversal behavior unchanged**.

## Tests

- `+1` engine-parametrized test (`test_explain_exposes_planner_diagnostics`): force → per-step + top-level diagnostics populated and reasons present; off (index resident) → `decision_reason == "policy=off"`.
- `45` CPU index tests pass (pandas + polars); ruff + mypy clean.
- GPU (cudf/polars-gpu) lanes deferred to dgx validation (local GPU drivers unavailable).

## Finding surfaced by this PR (pre-existing, not introduced here — logged for follow-up)

With diagnostics visible, `[n({'id':0}), e_forward]` under `index_policy='use'` uses the index on **polars but not pandas** (pandas chain routes seeds as source-match → `idx0` None → scans). The **direct** `hop()` uses the index on both engines. This is exactly the silent-scan case these diagnostics are meant to expose; the pandas chain seed-selection consulting the resident index is a separate follow-up.


lmeyerov:
Folded into #1658 — the gfql_explain cost-diagnostics (LP1) enrich the index feature this PR ships, so they now travel with it (2 commits cherry-picked onto #1658). Part of shrinking the stack (11→9 PRs).

Issue #1673, opened 2026-07-02
GFQL Cypher dialect: host the streamgl-viz filter language (regex/case, floor/ceil/round/pow, prune-isolated, two-mask via GRAPH{})


lmeyerov:
## Goal
Extend the **GFQL Cypher dialect** so it can be the single target for a (net-new) streamgl-viz filter/exclusion → Cypher translation layer, running on all four engines (pandas, cudf, polars, polars-gpu). Strategy: extend the Cypher flavor (we own it — `GRAPH { … }` constructor, `CALL … YIELD` procedures incl. graph-level, growable WHERE grammar + scalar-fn allow-list), rather than route around it to raw GFQL-AST.

Grounded in a per-engine conformance analysis of the viz filter AST (`~/Work/graphistry2/apps/core/viz` + forge shaping) vs the GFQL engine.

## Frontend (dialect) gaps
- **G1 — regex operator `=~`** (openCypher-standard; + optional LIKE/ILIKE) → lower to `match`/`fullmatch`/`contains(regex=True)`; case-insensitivity via inline `(?i)` and/or a case flag. *Flagship viz gap — Cypher WHERE has no regex/LIKE/ILIKE/case-insensitivity today.*
- **G2 — scalar functions `floor` / `ceil` / `round` / `pow`** → add to `GFQL_SCALAR_FUNCTIONS` + evaluator (all engines). Currently absent from the GFQL language entirely (rejected on every engine).
- **G3 — graph-level `CALL graphistry.pruneIsolated()` + `graphistry.pruneEdgelist()`** (`graph_only` procedures, like the existing `networkx.*` graph algos). Match viz semantics: `pruneEdgelist` = unconditional both-endpoint induced subgraph; `pruneIsolated` = opt-in, **undirected** degree, drops self-loop-only nodes, re-prunes. (No "prune isolated" op exists today; a `[n,e,n]` chain drops orphans but that's the opposite of viz's keep-by-default default and isn't attribute-preserving.)
- **G4 — two-mask composition via `GRAPH { … }`** — viz produces two *independent* row masks (node ⟂ edge) + prune; express as a GRAPH pipeline. May require closing the "GRAPH bindings + UNION" gap (parser) — scope carefully; the prune CALLs may make a single-pipeline form sufficient.

## Backend (per-engine execution) prerequisites — so the extended Cypher actually runs everywhere
[...]

PR #1674, opened 2026-07-02
fix(gfql/polars): Contains honors regex=/flags= (B1, viz-filter #1673)

lmeyerov:
First increment of the viz-filter → GFQL Cypher conformance work (#1673). Targets the in-flight polars-engine branch (#1660) since the fix is in its `filter_by_dict` lowering — **fold into #1660 or take as a stacked fix, your call.**

## Bug
The native polars `Contains` lowering (`lazy/engine/polars/predicates.py`) always used `str.contains(..., literal=False)`, so a **literal** request (`contains(pat, regex=False)`) was still regex-interpreted: a pattern with a metacharacter over-matched. Proven: `contains('a.c', regex=False)` matches `'abc'` on polars but only `'a.c'`/`'a.c.d'` on pandas/cuDF. `flags=` was also silently dropped.

## Fix
- `regex=False` → literal match (`literal=True`); case-insensitive literal folds both sides (matches pandas' result — polars has no literal+case flag).
- regex mode → maps `case=`/`flags=` (IGNORECASE/MULTILINE/DOTALL/VERBOSE) to a Rust-regex inline flag prefix `(?ims…)`.

## Why it matters
The viz filter language emits literal `CONTAINS`/`LIKE` as well as regex; without this, a literal filter with any regex metacharacter over-matches on polars. Prereq (B1) for the Cypher `=~` regex operator landing cross-engine (#1673).

## Tests
- +1 engine-parametrized differential-parity test (`test_polars_contains_regex_and_case_parity`): literal/regex × case-sensitive/insensitive, polars == pandas. Fails before the fix (literal `.` matched `'abc'`), passes after.
- polars chain + row-pipeline parity suites: **150 passed**; ruff + mypy clean.
- GPU (cudf/polars-gpu) parity deferred to a small dgx pass (#1673 B4) — local GPU unavailable.


lmeyerov:
Folded into #1660 — the Contains regex/flags fix is a fix to #1660's own polars engine code, so it now lives there (cherry-picked as a18c594b, same author/content/message, on top of #1660's team-polish commit). #1675's base retargeted to dev/gfql-polars-engine. Part of shrinking the stack (11→9 PRs).

PR #1675, opened 2026-07-02
feat(gfql/cypher): standard =~ regex + numeric fns + toLower/toUpper (#1673)

### LINKED ISSUE #1673: GFQL Cypher dialect: host the streamgl-viz filter language (regex/case, floor/ceil/round/pow, prune-isolated, two-mask via GRAPH{})

lmeyerov:
Implements the streamgl-viz filter-language conformance gaps (#1673) by adopting **standard openCypher/neo4j Cypher syntax** — not inventing dialect. Based directly on master (the polars `contains` regex fix it originally stacked on merged via #1660).

## Research first
Verified each construct against the neo4j Cypher manual, openCypher 9, and ISO GQL (sources in the issue). Finding: most "gaps" are **standard Cypher**, so we adopt the standard forms. `LIKE`/`ILIKE`/`BETWEEN` are SQL-only (not in Cypher or GQL) → intentionally not implemented; chained comparison (`1 < n.age < 65`) already worked.

## What's added (all standard openCypher/neo4j)
- **`=~` regex-match operator** — Java-regex, **full/anchored** match (`n.name =~ 'AB'` matches only `'AB'`), inline `(?i)`/`(?m)`/`(?s)` flags; lowers to `fullmatch`. Simple `WHERE …=~…` on all engines (`filter_by_dict`); composes through `AND`/`OR`/`NOT`/`RETURN` on pandas/cuDF (polars declines the complex `OR`/`NOT` row-filter form with an honest `NotImplementedError` — pre-existing `where_rows` gap). Adds native polars `Match`/`Fullmatch` lowering.
- **Numeric functions** `floor`, `ceil` (≡`ceiling`), `round(x)`/`round(x, precision)` (complementing existing `abs`/`sqrt`/`sign`).
- **`toLower`/`toUpper`** — the idiomatic case-insensitive compare.

Wired end-to-end: cypher grammar + shared `expr_parser` (REGEX_MATCH terminal) + allow-lists + evaluators (pandas/cuDF `row/pipeline.py`+`row/dispatch.py`; polars `predicates.py`+`row_pipeline.py`).

## Tests / quality
Differential-parity tested pandas↔polars for `=~`, Match/Fullmatch, numeric fns, `sign`, `toLower`/`toUpper`. Full polars lane passes; ruff+mypy clean. A focused adversarial review found no BLOCKERs; its two findings are addressed (polars parity now guarded by parametrized tests; polars `sign()` cast to Int64 to match pandas/neo4j Integer spec).

**GPU parity verified on dgx (GB10, `--gpus all`, 2026-07-05):** full polars lane + conformance matrix + `test_engine_polars_gpu.py` = **1803 passed, 0 skipped/failed** — cuDF + polars-gpu (cudf_polars) matrix cases (incl. Match/Fullmatch + the scalar fns) all parity-green vs the pandas oracle.

## Review hardening (4-wave converged adversarial review, 2026-07-05)
A multi-wave findings-mode review (artifacts: `plans/viz-filter-pipeline/waves/pr1675/`) found and fixed, each dgx-repro'd before and re-verified after: composed `=~` broken on `engine='cudf'` (no `.str.fullmatch`; now delegates to the hardened `Fullmatch` predicate); cuDF fullmatch alternation mis-anchoring (`'ab|cd'` silently matched `'abXXX'`); the cuDF `(?i)` lowercase-fold silently inverting escape classes (now a precise decline gate: uppercase escapes / case-crossing ranges / hex escapes / non-ASCII decline, lowercase escapes keep folding); polars `Match`/`Fullmatch` missing the Rust-regex guard (lookarounds crashed non-NIE); `toLower(numeric col)` broadcasting the Series repr on the pandas oracle (now declines like neo4j's type error); polars `floor`/`ceil`/`round` returning Int64 on int columns (now Float64 like neo4j/pandas); NotImplementedError being re-labeled by broad excepts (honest NIEs now survive to the caller); plus matrix coverage for the `=~` family, null cells, dtype pins, and discriminating regression fixtures. Waves 3–4 converged at zero BLOCKER/IMPORTANT.

## Docs
`docs/source/gfql/cypher.rst`: `=~` in WHERE Forms + a new "Scalar Functions and Operators" section (fns/toLower + why LIKE/BETWEEN aren't provided). CHANGELOG entries for both increments.

## Deferred (documented)
- `round(x, precision, mode)` 3-arg form; at this layer `round` uses numpy half-to-even — the stacked #1677 replaces it with neo4j's documented tie-breaking.
- `^` power operator: initially implemented right-associative, then **removed during standards vetting** — the openCypher TCK pins `^` LEFT-associative and the current conformance corpus marks it reject-expected, so re-adding is a coordinated corpus change (spec record in #1677/#1673).


graphistry/compute/gfql/lazy/engine/polars/predicates.py
```diff
@@ -233,6 +233,21 @@ def predicate_to_expr(col: str, pred: ASTPredicate, dtype: "Optional[pl.DataType
         prefix = _inline_regex_flag_prefix(case, flags)
         return c.str.contains(f"{prefix}{pred.pat}", literal=False)
 
+    if name in ("Match", "Fullmatch") and hasattr(pred, "pat") and isinstance(pred.pat, str):
```
lmeyerov: avoid dynamic typing patterns..

graphistry/compute/gfql/lazy/engine/polars/row_pipeline.py
```diff
@@ -103,12 +103,35 @@ def _lower_function(node: FunctionCall, columns: Sequence[str]) -> Optional[pl.E
         return pl.coalesce(args)
     if name == "abs" and len(args) == 1:
         return args[0].abs()
+    # neo4j/openCypher numeric fns (parity-verified vs the pandas engine).
     if name == "sqrt" and len(args) == 1:
-        # sqrt of a negative -> NaN on both pandas and polars; parity-verified.
-        return args[0].sqrt()
+        # sqrt of a negative -> NaN on both pandas and polars; Float64 cast so sqrt(int)
+        # returns float like neo4j/pandas; parity-verified.
+        return args[0].cast(pl.Float64).sqrt()
```
lmeyerov, line 110: avoid dangerous dynamic typing patterns..

graphistry/compute/gfql/row/pipeline.py
```diff
@@ -1358,6 +1360,51 @@ def _gfql_eval_expr_ast(self, table_df: Any, node: Any) -> Tuple[bool, Any]:
...
 
+            # neo4j/openCypher numeric fns: floor/ceil (alias ceiling), round(x[, precision]).
+            # Return FLOAT like neo4j; null in -> null out.
+            if fn in {"floor", "ceil", "ceiling"} and len(values) == 1:
+                inner = values[0]
+                use_ceil = fn in {"ceil", "ceiling"}
+                if hasattr(inner, "astype"):
+                    null_mask = self._gfql_null_mask(table_df, inner)
+                    f = inner.astype(float)
+                    if hasattr(f, "ceil" if use_ceil else "floor"):  # cuDF native
```
lmeyerov, line 1371: avoid dynamic typing patterns

graphistry/compute/predicates/str.py
```diff
@@ -424,21 +424,79 @@ def _validate_fields(self) -> None:
             )
 
 
+_CUDF_REGEX_UNSUPPORTED = re.compile(r'\(\?=|\(\?!|\(\?<|\\[1-9]|\(\?P[=<]')
```
lmeyerov, line 435: are we sure, eg, https://docs.rapids.ai/api/libcudf/legacy/md_regex ? maybe track down

PR #1679, opened 2026-07-05
feat(gfql/cypher): EXISTS { pattern } subqueries — declarative prune-isolated, native on 4 engines

lmeyerov:
Implements `EXISTS { <pattern> }` pattern-existence subqueries (openCypher-standard) in WHERE position, native on all four engines (pandas/cuDF/polars/polars-gpu) — the declarative prune-isolated building block for the streamgl-viz filter pipeline (frozen semantics: the panel-algebra doc, viz-filter L1). Stacked on #1677.

## What's added
- **Parser**: `EXISTS { <pattern> }` / `NOT EXISTS { ... }` in WHERE. The body parses into the SAME `BooleanExpr(op="pattern")` leaf as a bare pattern predicate — zero new lowering machinery; `NOT` composes through the existing lift into `anti_semi_apply`. Inline block comments and property maps (`{w: 1}`) inside the pattern work. RETURN/WITH-position `EXISTS` and the `not((a)--(b))` spellings keep their clean fail-fast.
- **Existential locals**: aliases introduced inside the braces are allowed (`EXISTS { (n)--(m) }` with `m` unbound outside) — the correlated bindings project them away. Bare pattern predicates keep the conservative no-new-aliases guard.
- **Drop-self flavor**: the one supported inner `WHERE` form is endpoint inequality — `EXISTS { (n)--(m) WHERE m <> n }` ("has a neighbor other than itself"). Carried as a safelisted `neq=` param on the semi-apply ops: pandas/cuDF filter the correlated bindings table; polars excludes self-loop edges (exactly the `m <> n` witness for a single-edge pattern).
- **Polars goes native** (was honest-NIE): `semi_apply_mark` / `anti_semi_apply` (correlated key sets via the polars chain executor's named-flag columns — the backward prune makes the flag exactly "a full pattern match exists through this node"; order-preserving `is_in` instead of joins, null keys matching pandas merge semantics) and `rows(binding_ops=...)` for the single-entity row table (mirrors the pandas alias-lookup-frame layout). v1 declines honestly: multi-alias correlation, non-single-hop shapes, `query=` nodes.

## Prune-isolated, declaratively (the point)
```cypher
MATCH (n) WHERE EXISTS { (n)--() } RETURN n              // keep-self flavor
MATCH (n) WHERE EXISTS { (n)--(m) WHERE m <> n } RETURN n // viz drop-self flavor
```
Both flavors are oracle-pinned in the conformance matrix on a discriminator graph (an isolated node + a self-loop-only node), 4-engine parity-or-NIE.

## Tests / verification (dgx GB10, `--gpus all`)
Full polars lane + conformance matrix + GPU suite + row-pipeline ops = **2082 passed, 0 failed** (final tip); cypher lowering suite 1193+ passed (EXISTS marker-shape parity with bare patterns, NOT→anti, neq threading, boundary declines). End-to-end sanity: `NOT EXISTS {(n)--()}` returns exactly the isolated node on pandas == cuDF == polars. Conformance-ledger waiver text updated to native status (`join_apply` remains the honest-NIE waiver). ruff + mypy clean.

## Honest boundaries (clear errors, no wrong answers)
`EXISTS` in RETURN/WITH projections; general inner `WHERE` clauses; multi-pattern bodies; full `MATCH..RETURN` subquery bodies; multi-alias correlation on polars (pandas/cuDF handle it via the pre-existing bindings machinery).

## Review hardening (3-wave converged adversarial review, 2026-07-05)
Findings-mode review (artifacts: `plans/viz-filter-pipeline/waves/pr1679/`) found and fixed, dgx-verified: `GRAPH { }` + `EXISTS` was **silently dropping the filter** (all engines — now fails fast, anchored to the grammar's constructor-at-start rule with zero false positives proven); the working graph-state prune patterns are pinned 4-engine per the graph-shape acceptance requirement (`GRAPH { MATCH (a)-[e]-(b) }` keep-self with ALL edges; `+ WHERE a.id <> b.id` drop-self); EXISTS-body keyword scans run on masked text (property strings like `'WHERE it hurts'` no longer falsely decline); `neq` validates as an exact string pair (was a late IndexError / silent first-pair filter); self-alias `EXISTS { (n)--(n) }` declines instead of a non-NIE crash; null-key and alias-collision corners decline honestly; the pandas name-flag column leak is mirrored exactly; native-must/multi-alias-NIE/decline-shape pins added (changed-line coverage gate green). TCK: the openCypher existential-subquery scenario `expr-existentialsubquery1-3` is formally **promoted to supported** in the tck-gfql corpus (mirror branch `feat/gfql-cypher-exists`); `1-1` tracked as a whole-entity-rendering wrong-rows debt. Waves 2–3 converged at zero BLOCKER/IMPORTANT. Known follow-up (pre-existing at base): bare pattern predicates inside `GRAPH { }` bodies also silently drop (`row_pre_filters` unwired in the graph compiler) — L3 item.

## Post-convergence tip movements (structure/CI only, re-verified)
`pattern_apply.py` module split (module-per-family, per review steer): the polars semi-apply/pattern-existence lowerings move out of `row_pipeline.py` into `graphistry/compute/gfql/lazy/engine/polars/pattern_apply.py` — dispatch imports updated, no behavior change (dgx 1703-pass; post-rebase import-shape greps). The polars coverage-audit baseline gains the required `pattern_apply.py` entry (floor 67.0; dgx-measured 75.36% on the CI-verbatim CPU lane, identical on this and the L2 tree) — the audit fails un-baselined target files, which was the only CI red. CI green at the final tip.


docs/source/gfql/cypher.rst
```diff
@@ -271,6 +271,21 @@ WHERE Forms
...
+  (``EXISTS { (n)-[:R {w: 1}]->() }``), and the one supported inner ``WHERE``
+  form is endpoint inequality: ``EXISTS { (n)--(m) WHERE m <> n }`` — "has a
+  neighbor other than itself", i.e. the drop-self-loop prune-isolated flavor.
+  Runs natively on all four engines. Not yet supported (clear errors):
+  ``EXISTS`` in RETURN/WITH projections, general inner ``WHERE`` clauses,
+  multi-pattern bodies, full ``MATCH .. RETURN`` subquery bodies, and ``EXISTS``
+  inside ``GRAPH { }`` pipelines. For prune-isolated in GRAPH STATE (nodes AND
+  edges back), use edge patterns instead: ``GRAPH { MATCH (a)-[e]-(b) }`` keeps
+  every edge-touching node with ALL its edges (self-loops included); add
+  ``WHERE a.id <> b.id`` for the drop-self-loop variant.
```
lmeyerov, line 288: there is probably fuzzing & tck repo work we should do across the pr's

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -5763,7 +5763,10 @@ def _where_expr_tree_pattern_predicates(expr: BooleanExpr) -> List[WherePatternP
                     line=cur.span.line,
                     column=cur.span.column,
                 )
-            out.append(WherePatternPredicate(pattern=cur.pattern, span=cur.span, negated=False))
+            out.append(WherePatternPredicate(
+                pattern=cur.pattern, span=cur.span, negated=False,
+                pattern_origin=getattr(cur, "pattern_origin", "bare"),
+                pattern_neq=getattr(cur, "pattern_neq", None)))
```
lmeyerov: avoid dynamic typing patterns

graphistry/compute/gfql/cypher/parser.py
```diff
@@ -430,10 +432,16 @@ def _build_where_with_pattern_lift(
     pattern_preds: List[WherePatternPredicate] = []
     for leaf in pattern_leaves:
         assert leaf.pattern is not None, "pattern_atom invariant: pattern payload always set"
-        pattern_preds.append(WherePatternPredicate(pattern=leaf.pattern, span=leaf.span, negated=False))
+        pattern_preds.append(WherePatternPredicate(
+            pattern=leaf.pattern, span=leaf.span, negated=False,
+            pattern_origin=getattr(leaf, "pattern_origin", "bare"),
+            pattern_neq=getattr(leaf, "pattern_neq", None)))
```
lmeyerov: avoid dynami typin gpatterns here too

graphistry/compute/gfql/lazy/engine/polars/pattern_apply.py
```diff
@@ -0,0 +1,231 @@
...
+def _semi_apply_join_col(left: Any, alias: str, node_id: str) -> Optional[str]:
+    """Mirror the pandas join-column choice: prefer ``alias.node_id``, else bare ``alias``."""
+    for cand in (f"{alias}.{node_id}", alias):
+        if cand in left.columns:
+            return cand
+    return None
+
+
+def semi_apply_mark_polars(
+    g: Plottable, binding_ops: Sequence[dict], join_aliases: Sequence[str], out_col: str,
```
lmeyerov: avoid unqualified `dict` etc

graphistry/compute/gfql/lazy/engine/polars/pattern_apply.py
```diff
@@ -0,0 +1,231 @@
...
+def _binding_ast_ops(binding_ops: Sequence[dict]) -> Optional[List[Any]]:
+    """Deserialize the semi-apply family's serialized binding ops; None on failure."""
+    from graphistry.compute.ast import from_json as ast_from_json
+    try:
+        return [ast_from_json(op_json, validate=False) for op_json in binding_ops]
+    except Exception:
+        return None
+
+
+def rows_binding_ops_polars(g: Plottable, binding_ops: Sequence[dict]) -> Optional[Plottable]:
```
lmeyerov: qualify types

PR #1680, opened 2026-07-05
feat(gfql/cypher): searchAny cross-column search predicate — inspector semantics, native on 4 engines

lmeyerov:
Implements `searchAny(entity, term[, opts])` — the streamgl-viz inspector's cross-column table search as a composable openCypher-style WHERE predicate, native on all four engines (pandas/cuDF/polars/polars-gpu), plus Python twins `g.search_nodes()` / `g.search_edges()`. Viz-filter L2 (frozen semantics: panel-algebra D2; naming/options via a 4-persona design pass — artifacts under `plans/viz-filter-pipeline/research/searchany-persona-pass.md`). Stacked on #1679.

## Semantics (dtype gate is SEMANTICS, not optimization)
- OR across the entity's columns; **case-insensitive substring by default** (case-folded, never regex — the common viz call avoids every engine regex limit); `regex: true` opt-in obeying the same per-engine decline rules as `=~` (now enforced on cuDF too: prep + casefold-or-decline before delegating, instead of libcudf crashes or `\D`→`\d` inversion).
- Dtype gate: string columns always; **integer columns iff the term is a numeric literal** (the inspector's `/^[0-9.-]+$/` gate); floats/dates/booleans only via the explicit `columns:` list. Explicit non-string columns are served where stringification provably matches the pandas oracle: bool everywhere (polars canonicalizes `'True'/'False'`), float/date on pandas only — **polars and cuDF decline them honestly** (dgx-probed: pandas `str(1e16)='1e+16'` vs polars `'1e16'` vs cuDF `'1.0e+16'`, and cuDF truncates mantissas — `0.1+0.2 → '0.3'`).
- Options map `{caseSensitive, regex, columns}` strict-validated: unknown keys, non-boolean flags, malformed column lists, and unbound aliases all error clearly; missing explicit columns raise the same E108 on pandas AND polars (pinned to beat the dtype gate in precedence); `$param` terms decline with a clear message; null cells never match (incl. `astype(str)`'s `'nan'/'<NA>'` artifacts, masked back out).
- `searchAny(...)` spelled **inside a string literal is data, not a call** — the lowering lift matches against a literal/comment-masked copy of the WHERE text (with rescan-after-skip so quote-spanning pseudo-matches can't swallow real calls).
- Nodes and edges independently searchable; composes through AND/OR/NOT (lowered like the pattern-predicate markers); two `searchAny` in one WHERE compose (pinned).

```cypher
MATCH (n) WHERE searchAny(n, 'acme') RETURN n
MATCH (n) WHERE searchAny(n, 'alpha') AND searchAny(n, '7') RETURN n
```

## Engineering
Per-column matching **reuses the parity-hardened `Contains` predicate** on pandas/cuDF (every cuDF workaround + honest decline from the #1675 review waves inherited) and a fold/`any_horizontal` lowering on polars (Rust-regex guard on the regex path; module-per-family `polars/search.py`). Safelisted row op with schema effects; conformance-ledger entries; polars coverage baseline gains a `search.py` floor (79.0; dgx-measured 86.96 on the CI-verbatim lane). One deliberate base change: the call executor's NIE pass-through extends from polars-only to **row-pipeline calls on any engine** (`is_row_pipeline_call`), so kernel-level per-engine declines surface as honest `NotImplementedError` instead of a wrapped E303 — non-row-pipeline NIEs (e.g. `fa2_layout` without GPU) still wrap, pinned.

## Review + verification
5 adversarial findings-mode waves to convergence (waves 4+5 consecutive clean; artifacts `plans/viz-filter-pipeline/waves/pr1680/`). Waves fixed 2 BLOCKERs (in-literal lift rewrite; cuDF regex guard bypass) and 5+1 IMPORTANTs (null-cell stringify match, polars E108, `$param`, polars bool `'true'` silent divergence, float repr divergence — each now pinned). dgx GB10 (`--gpus all`, rapids 26.02): targeted 5-file lane **1703 passed** at the converged tip; CI green incl. the polars coverage audit. ruff + mypy clean.

## Boundaries (clear errors, all pinned)
`searchAny` in RETURN/WITH projections (WHERE only); edge-alias `searchAny(r, …)` declines on polars (pending multi-entity binding-row support); explicit columns beyond string/int/bool decline on polars + cuDF; polars-frame Python twins decline (use the cypher op); regex declines mirror `=~`.


graphistry/compute/gfql/lazy/engine/polars/search.py
```diff
@@ -0,0 +1,109 @@
...
+from __future__ import annotations
+
+from typing import Optional, Sequence
+
+from graphistry.Plottable import Plottable
+
+from .row_pipeline import _active_table, _rewrap
+
+
+def search_any_polars(
```
lmeyerov, line 13: ensure cross-platform not just polars, and cross-test

graphistry/compute/gfql/lazy/engine/polars/search.py
```diff
@@ -0,0 +1,109 @@
+"""Native polars lowering for the ``search_any`` cross-column search row op
```
lmeyerov, line 1: as part of index support, we can revisit this later when we add text indexes

graphistry/compute/gfql/lazy/engine/polars/search.py
```diff
@@ -0,0 +1,109 @@
...
+            base = pl.col(real).cast(pl.String)
+        if regex:
+            pat = term if case_sensitive else f"(?i){term}"
+            exprs.append(base.str.contains(pat, literal=False))
+        elif case_sensitive:
+            exprs.append(base.str.contains(term, literal=True))
+        else:
+            exprs.append(base.str.to_lowercase().str.contains(term.lower(), literal=True))
+    marked = left.with_columns(pl.any_horizontal(exprs).fill_null(False).alias(out_col))
+    return _rewrap(g, marked)
```
lmeyerov, line 109: we should audit our tests:

- what happens when nested structures
- are we appropriatley short circuiting, eg, skipping bool cols and numeric cols for non-numeric str
- ... what happens when pretty printer adds "." and "," to ui? should we search those too?
- true/false variants (True, False, 0, 1, ...) can probably be converted to bool vals for when searching bool cols, what's normal here?

graphistry/compute/ComputeMixin.py
```diff
@@ -487,6 +487,55 @@ def get_topological_levels(
...
+                "search_any op or engine='pandas'")
+        mask = search_any_mask(
+            df, term, case_sensitive=case_sensitive, regex=regex, columns=columns)
+        if mask is None:
+            raise GFQLValidationError(
+                ErrorCode.E108,
+                "search_edges columns= includes a column absent from the edges table",
+                field="columns", value=columns,
+                suggestion="List only columns present on the edges table.")
+        return self.edges(df[mask])
```
lmeyerov, line 537: DRY

PR #1681, opened 2026-07-05
test(gfql): viz-filter-pipeline acceptance suite + regression benchmark (L3)

lmeyerov:
The viz-filter-pipeline ACCEPTANCE SUITE + REGRESSION BENCHMARK (viz-filter L3) — the proveout layer over the whole stack (#1675 → #1677 → #1679 → #1680). Frozen semantics: `plans/viz-filter-pipeline/research/panel-algebra.md` (v1-FINAL). Stacked on #1680.

## `graphistry/tests/compute/gfql/test_viz_pipeline_conformance.py`
- **Curated full-panel pipelines** (9, each: hand-derived oracle pin + an independent plain-pandas second oracle + `assert_parity_or_nie` across pandas/cuDF/polars/polars-gpu): node+edge filters, **exclusion-dominates** composition (`AND NOT (...)` with three-valued null semantics pinned), `(pred OR IS NULL)` keep-null leaves, **both EXISTS prune-isolated flavors**, searchAny composed with filters, deterministic `ORDER BY`/`SKIP`/`LIMIT` paging **ascending AND descending** — the fixture's row order is deterministically shuffled so an engine that silently drops the sort fails on membership, and the ASC+DESC pins are jointly seed-independent (review wave-3 proof).
- **Graph-state prune shapes**: exact node **and edge** pins for `GRAPH { MATCH (a)-[e]-(b) }` (keep-self: self-loop-only node kept with its edge, mutual pair keeps both directed edges, isolated dropped) + the drop-self variant, on a formula-derived fixture (self-loops, parallel edges, isolated + self-loop-only discriminator nodes, unicode/metachar/null name cells, a Categorical column).
- **40-seed panel-state fuzzer**: random panel states → generated single cypher query → pandas result vs an independently computed plain-pandas expectation (zero pandas declines allowed), then 4-engine parity-or-NIE; out-of-dialect combos must decline with clear errors (crash classes fail).
- **Case/regex/unicode trick matrix** (17 rows): `(?i)` vs case-sensitive `=~`, ß/İ full-case-mapping pins, literal-vs-regex metachars through searchAny, anchored `=~` vs CONTAINS, null cells, Categorical (decline ok, silent-wrong not).

## `benchmarks/gfql/viz_filter_pipeline.py`
Six streamgl-viz panel scenarios (filters+excludes+ORDER/LIMIT, keep-self GRAPH prune, EXISTS prune, node search, edge search, combined) × `--scale {100k,1m,10m}` × engines with **native-frame-per-engine fairness** (one-time conversion excluded from timing; both result frames touched inside the clock), warmup+timed medians, an NIE-tolerant never-crash matrix, `--est` for safe-run budgeting, upfront output-path validation, and JSON receipts. The ~350ms@800K interactive reference is printed as a comparison column, not asserted.

## Review + verification
4 adversarial findings-mode waves to convergence (waves 3+4 consecutive clean; artifacts `plans/viz-filter-pipeline/waves/pr1681/`) — every curated pin and all 17 trick rows independently recomputed by hand in review; oracle-independence and bench fairness adversarially confirmed. dgx GB10 (`--gpus all`): suite **31 passed** at the converged tip on the final stack; CI green.

## Receipts (dgx GB10 — `plans/viz-filter-pipeline/receipts/`)
- **100k ×4 engines**: polars 2–5× pandas across panel scenarios; everything within the interactive reference except pandas `combined` (1.55×).
- **1m ×4 engines**: prune/exists/combined at 1.15–2.3× the 350ms reference on polars/cudf/polars-gpu (pandas 3.4–15.7×); search sub-budget on all engines (cudf `search_nodes` 36ms).
- **10m ×4 engines**: receipt in `receipts/` (25M edges; run under the memory-safety harness).

## Findings surfaced by this layer (pinned + tracked, exactly its job)
- The same-path WHERE route **dedupes parallel edges** (one per unordered endpoint pair — a mutual pair may collapse to one direction) — diverges from the panel algebra's E₂ edge multiplicity; pinned at exactly that tolerance with the prune-correctness facts still asserted; repo-debt filed.
- Edge-alias `searchAny(e, ...)` **declines on polars** (honest NIE) — gap tracked; explicit non-string/int/bool searchAny columns decline on polars + cuDF (float repr divergence, dgx-probed in #1680).
- Pre-existing (inherited notes): cuDF same-path TypeError on the drop-self GRAPH form; bare pattern predicates inside GRAPH bodies silently drop (`row_pre_filters` unwired in the graph compiler).

### DIFF
--- a/graphistry/compute/gfql/search_any.py
+++ b/graphistry/compute/gfql/search_any.py
@@ -8,7 +8,9 @@ delegates to the parity-hardened ``Contains`` predicate, so every pandas/cuDF
 quirk and honest decline gate carries over; cuDF regex obeys the same decline
 rules as ``=~``."""
 import re
-from typing import Any, List, Optional
+from typing import List, Optional
+
+from graphistry.compute.typing import DataFrameT, SeriesT
 
 # inspector's numeric-term gate (streamgl-viz sortAndFilterRowsByQuery.js)
 _NUMERIC_TERM_RE = re.compile(r"^[0-9.\-]+$")
@@ -18,18 +20,18 @@ def is_numeric_term(term: str) -> bool:
     return bool(_NUMERIC_TERM_RE.match(term))
 
 
-def _is_searchable_string_dtype(dtype: Any) -> bool:
+def _is_searchable_string_dtype(dtype: object) -> bool:
     import pandas.api.types as pat  # cuDF mirrors the pandas dtype API
     return bool(pat.is_string_dtype(dtype)) or bool(pat.is_object_dtype(dtype))
 
 
-def _is_int_dtype(dtype: Any) -> bool:
+def _is_int_dtype(dtype: object) -> bool:
     import pandas.api.types as pat
     return bool(pat.is_integer_dtype(dtype))
 
 
 def search_candidate_columns(
-    df: Any, term: str, columns: Optional[List[str]]
+    df: DataFrameT, term: str, columns: Optional[List[str]]
 ) -> Optional[List[str]]:
     """Columns to search: the explicit list (None if any is missing — caller declines
     loudly) or the auto dtype gate."""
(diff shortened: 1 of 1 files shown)


lmeyerov:
Superseded by #1696 (same branch, rebased onto master, base=master). This PR was auto-closed when its base branch #1680 merged+deleted, and GitHub won't reopen a closed PR with a deleted base — so the L3 acceptance content lives on in #1696. Ignore this one.

PR #1693, opened 2026-07-05
refactor(gfql/cypher): consume WHERE directly from MatchClause (remove flatten seam)

### LINKED ISSUE #1687: GFQL Cypher parser: remove the WHERE flatten seam (assembler should consume MatchClause.where directly)
Follow-up from the Earley→LALR parser PR (#1682).

## Context
The LALR grammar bundles a trailing `WHERE` into its `match_clause` (so WHERE binds to its preceding clause declaratively, which eliminated the WITH..WHERE attachment ambiguity). To keep that change **AST-byte-identical**, the `query_body` transformer currently *re-flattens* a bundled `MatchClause(where=…)` back into a `[match, where]` item sequence so the pre-existing clause-sequence assembler (a state machine that predates the bundling) runs untouched.

## Ask
Have the assembler consume `MatchClause.where` **directly** and delete the positional WHERE-attachment inference. Re-prove equivalence with the existing differentials (`test_grammar_invariants.py` + `test_differential_grammar_parity.py` — byte-identical ASTs, rejection parity). With the seam gone, the AST's WHERE-related fields may also be simplified (the AST is an internal contract, safe to redesign with consumers refactored in lockstep + differential re-proven).

Scoped, low-risk, no new machinery. Parser-only.

🤖 tracked via Claude Code

lmeyerov:
Closes #1687. Follow-up cleanup to the LALR parser (#1682).

## Context

The grammar bundles a trailing `WHERE` onto its `MATCH` clause (that's what made the WITH..WHERE attachment unambiguous). To keep the Earley→LALR switch **byte-identical**, the transformer took a shortcut: it split `MatchClause.where` back out into a synthetic standalone `WhereClause` item and let the pre-existing clause-assembler re-attach it — so the gnarliest ~230-line method ran untouched. #1682 flagged that seam as a scoped follow-up.

## Change

The assembler now consumes `MatchClause.where` **directly**:
- **primary MATCH** keeps its `WHERE` on the clause (and it becomes the query's top-level `where`, last-clause-wins so `MATCH … WHERE … OPTIONAL MATCH … WHERE …` scopes each predicate) — the grammar already set it, so this is a pure deletion of the split-then-reattach round-trip;
- **post-WITH re-entry MATCH**'s `WHERE` moves into `reentry_wheres` with the clause carrying none (the one case that genuinely needs a split).

The now-unreachable standalone-`WhereClause` branches in both `query_body` and `graph_constructor` are removed (the grammar can't emit a standalone WHERE item). **Net −23 lines.**

## Pure refactor — no behavior change

- **Old-vs-new production AST differential** over a **1,989-query** repo corpus (every WHERE / reentry-MATCH / graph-constructor shape): **1989/1989 byte-identical, 0 AST diffs, 0 language diffs.**
- Grammar invariants (zero conflicts / `strict=True` / semantic-ambiguity-zero / rule-coverage) + full-repo LALR≡Earley differential: pass.
- **Full cypher suite on dgx-spark: 1,737 passed / 0 failed.**
- mypy clean.

## Base

Stacked on `test/gfql-viz-pipeline-acceptance` (the tip of the #1675→#1681 grammar stack) rather than master, so it works against the final grammar (post-EXISTS/searchAny) and rebases along with the stack. Confirmed the zero-conflict invariants still hold on the tip after those grammar additions.

https://claude.ai/code/session_01W5vkD2ZCyv3bmecBYoYYQy


lmeyerov:
red b/c unrelated EXISTS work

Issue #1695, opened 2026-07-05
GFQL searchAny: WYSIWYG search coverage — match user-visible formatting (float, datetime) + format options


lmeyerov:
Follow-up from the searchAny work (#1680). searchAny currently searches **string** and **integer** columns (numeric-term-gated), matching the streamgl-viz inspector's `shouldSearch` for those dtypes. It does **not** search **float** or **date** columns, which the inspector *does* (numeric-term-gated). This issue tracks closing that gap.

## Reference (ground truth)
`~/Work/graphistry/apps/core/viz/src/worker/services/dataframe/sortAndFilterRowsByQuery.js` + `src/formatters/defaultFormat.js`. The inspector searches number/integer/date columns (iff the term is numeric) via `defaultFormat`: fractional floats → `sprintf('%.4f')`; whole numbers → raw `String()`; dates → moment `MMM D YYYY, h:mm:ss a z`. Full analysis: `plans/viz-filter-pipeline/research/searchany-inspector-parity.md`.

## Float — proven native cross-engine path (dgx-probed, NO host-bridge, NO broad-NIE)
The naive `astype(str)` diverges in the exponent regime (`1e16` → `'1e16'` vs `'1e+16'`) — the original reason floats were excluded. The fix:
- **`round(4)` then fixed-point decimal render**: `cudf .astype(Decimal128(scale=4)).astype(str)` ≡ `polars .cast(Decimal(scale=4)).cast(str)` **exactly** (vectorized, no exponent, GPU-native).
- Apply the **same `round(4)+decimal` convention on pandas too** (not its `f\"{v:.4f}\"`, which uses true float bits and diverges at half-boundaries like `0.12345`→`0.1235` vs `0.1234`) so all four engines agree.

## Implementation sketch (on `graphistry/compute/gfql/search_any.py` + `lazy/engine/polars/search.py`)
1. `_is_float_dtype` + include float in the auto gate for numeric terms.
2. `_canonical_float_str(series)` = `round(4)` → fixed-4-decimal string, engine-aware (pandas/cudf/polars Decimal).
3. Wire into `search_any_mask` (kernel) + `search_any_polars`.
4. **Remove** the now-unneeded exclusions: the cuDF explicit-float NIE (`search_any.py`) and the polars `_stringify_ok` float-exclusion (`search.py`).
[...]

Issue #531, opened 2023-12-24
[FEA] GFQL indexing support


lmeyerov:
**Is your feature request related to a problem? Please describe.**

One of the easiest ways to speedup pandas code is to support indexes

**Describe the solution you'd like**

GFQL has a few interesting scenarios here:

* Indexing on node/edge IDs wrt global lookups
* Indexing on additional columns, especially text
* Indexes being passed in
* Indexing being requested
* Indexes happening on-the-fly at start & mid-traversal

It's unclear what's most important, I'm guessing:

* node/edge ID indexing
* str indexing

PR #1700, opened 2026-07-06
feat(gfql): searchAny WYSIWYG float + datetime search (#1695)

lmeyerov:
## What

Makes GFQL `searchAny` / `search_nodes` / `search_edges` match **what the streamgl-viz inspector RENDERS** for **float** and **datetime** columns — you search the *displayed* string, not the raw value (#1695). Adds WYSIWYG format call-params `{floatPrecision, temporalFormat, tz}`.

Floats were previously excluded from search because `astype(str)` diverges across engines (exponent regime + half-tie double-rounding). This makes the render **byte-identical to the inspector** and, per the parity-or-honest-NIE convention, keeps every engine either exact or an honest decline.

## The decisive experiment (dgx cross-engine parity fuzzer)

Ran a float-render parity fuzzer on **dgx-spark** (all engines) with the **real viz `sprintf-js`** as the oracle. It settled the render and **falsified the original "cross-engine decimal cast = parity" plan**:

| engine | verdict |
|---|---|
| **pandas `"%.4f" % v` direct** | ✅ **exact** — byte-identical to `sprintf-js` (`3.28045`→`"3.2805"`, `0.12345`→`"0.1235"`) |
| pandas w/ `np.round` prepass (old draft) | ❌ double-rounds 73/210 half-ties (`-3.74825`→`"-3.7482"`) |
| cuDF float→Decimal128 | ❌ **truncates** (wrong on generic values, not just ties) |
| polars Decimal cast | ❌ diverges at 5th-decimal ties + drops sign on `-0.0000` |

There is **no native GPU printf**, and a per-cell UDF would be a forbidden host-bridge. Datetime render verified **byte-identical to real `moment.utc`** for the UTC path.

## Design

- **Float**: `"%.4f" % v` direct for fractional (single correctly-rounded conversion), `str(int(v))` for whole. **pandas-exact**; cuDF/polars **honest-NIE**. `inf`→`"Infinity"` (matches JS).
- **Datetime**: reproduces the inspector's moment `'MMM D YYYY, h:mm:ss a z'` via strftime, **localized to a caller `tz`** (per-timestamp DST-correct). pandas-only; cuDF/polars honest-NIE.
- **Options API** `{floatPrecision, temporalFormat, tz}`: threaded through the cypher option-map (strict typed parse), the ast op, call-validation, and the python twins; defaults to the inspector (4 decimals; UTC).
- **Auto-gate (decision A, symmetric)**: float/datetime stay **out of the auto-gate on every engine**, so bare `searchAny('...')` behaves identically cross-engine; float/datetime are reached via explicit `columns=`. (`@leo` chose this over the "GPU raises" alternative.)

## ⚠️ Open decision for @lmeyerov

Datetime default `tz='UTC'` (chosen for determinism — a server-local default varies by deployment and would break the parity oracle / tests). You suggested "default to server time" — trivially switchable if you'd prefer that; the `tz=` call-param already delivers the localization knob either way.

## Adversarial review (`.agents/skills/review`)

Multi-angle finders + verify. Three confirmed findings fixed: (1) `inf`/`-inf` render + RuntimeWarning; (2) am/pm lowercasing corrupting alpha tz abbrevs (`AMT`→`amT`) — now word-bounded; (3) negative/bool `floatPrecision` + empty `tz` silently misrendering via the twins (which bypass the safelist) — added kernel-level `validate_format_opts` + tightened the safelist. Cross-engine-safety and parity-invariant hypotheses were refuted (clean).

## Testing

All on **dgx-spark** (`test-rapids-official:26.02-gfql`, pandas + cuDF + polars + polars-gpu):
- New `test_search_any_float_wysiwyg`, `test_search_any_datetime_wysiwyg`, `test_search_any_format_options` — pandas exact + cuDF/polars honest-NIE, half-tie/inf/tz/validation pinned.
- Full conformance + cypher suites: **2041 passed / 7 skipped / 15 xfailed**, no regressions.


lmeyerov:
## Perf: interactive latency + the base-graph index question

Benchmarked the "typing into a search window" scenario on dgx-spark (pandas, 100K/1M rows, 8 mixed-typed columns), since each keystroke dispatches a fresh query but the columns+data are fixed.

### Render optimizations landed (1ce40e60)
The per-keystroke render was the hot spot:
- **Float**: the `np.char.mod`/numpy-masked-scatter render was *slower* than a plain list-comp (multiple full-array passes + object boxing) — reverted to a list-comp, byte-identical, ~15% faster.
- **Datetime**: pandas `dt.strftime` is a ~20×-slow per-element path (~2.2s/1M). For the default format in UTC (common, deterministic) I now assemble the render vectorized from `.dt` components + a month lookup — **byte-identical** to strftime (verified incl. midnight/noon/single-digit-day/NaT). Custom `temporalFormat` / non-UTC `tz` keep strftime.

Per-keystroke @ 1M: **datetime 3111ms → 1512ms (2.05×)**, float 674ms → 557ms.

### The index question (base vs view) — quantified

The render is **term-independent**, so it should be computed once and reused. Measured @ 1M / 8 cols:

| approach | per-keystroke | vs today |
|---|---|---|
| (A) today — render+match every keystroke | 2903 ms | 1× |
| (B) per-column render cache — render once, N `Contains` | 439 ms | 6.6× |
| **(C) concatenated search-blob index** — render+join once, **1** `Contains` | **68 ms** | **43×** |

- One-time build (term-independent): blob **567 ms**; per-column cache 2480 ms.
- **Base vs view**: slicing the base blob to a 50% filtered view = **8 ms slice + 36 ms keystroke** — so the index lives at the **base-graph level** and **slices to the filtered/rendered view**; no per-keystroke re-render.

**Recommendation**: a base-graph **search-text index** — a single concatenated, lowercased WYSIWYG search-text column built once per (data, format-options), sliced to the active view, so each keystroke is one `Contains`. That's the 43× win and it's what turns 1M-row interactive search from ~3s to ~68ms/keystroke. It's an architectural addition (where the index lives on the `Plottable`, invalidation on data/option change, integration with the chain/viz dispatch) — proposing to track it as a follow-up unless we want it in this PR.

(The correctness feature in this PR is unchanged and CI-green; the render opts are pure speedups with byte-identical output.)

lmeyerov:
Search-text index follow-up filed: #1703 (base-graph blob index, 43× interactive win, cross-referenced to the #531/#1658 CREATE INDEX stack per @lmeyerov). The render speedups in this PR are the standalone, byte-identical part.

lmeyerov:
Both halves of this have now landed, by different routes.

**Float** landed in #2089 (`22bfde5cd`) with a different design. This PR's decisive experiment stands — the cross-engine decimal cast really was the wrong path — but the implementation that shipped renders through `graphistry/compute/gfql/wysiwyg.py` rather than inline in `search_any.py`, with `float_precision` as a parameter.

**Datetime** landed just now in #2105 (`e43d5fdbd`), ported onto that same module rather than rebased. This branch is 1316 commits behind and its conflicts sit almost entirely in the float code that #2089 superseded, so a rebase would have meant resolving them in favour of deleting most of what was being rebased.

Your analysis carried over directly, and two parts of it were worth keeping:

- the moment format `'MMM D YYYY, h:mm:ss a z'` and its strftime equivalent, including the lowercase `%p` that moment's `a` needs
- `tz` as a caller parameter, with the reasoning that the inspector renders in the viewer's zone and a server cannot know it. That turned out to matter more than cosmetics: `2024-01-01T02:00Z` is 2023 in New York, so the zone decides which rows match.

Two things came out differently once measured, and are worth recording here rather than leaving as folklore:

**The `~3x faster` render claim does not hold.** Component extraction with a per-row `%Z` measures **0.7x — slower** than plain `dt.strftime` on 1M rows, because 60% of the cost is the per-element `strftime('%Z')`. Your `_assemble_default_datetime_utc` was presumably fast because it hardcodes the literal `'UTC'`, which is consistent with its own note about falling back to strftime "where a zone's abbrev varies". Resolving the abbreviation once per distinct UTC offset instead gives **1.8x**, and that is what shipped.

**The parity evidence needed widening.** This PR verified against `moment.utc`. The shipped render is byte-identical to real `moment-timezone` across 3000 timestamps in 12 zones, then checked adversarially against all 597 zones — where the only differences turned out to be timezone-database vintage disagreements, not arithmetic.

Engine support ended up broader than this PR proposed: pandas and polars render every zone and answer identical rows, and cuDF matches at UTC and declines other zones rather than mislabel them, because cudf's `strftime` ignores `tz_convert`.

Closing as superseded, not rejected — the work is in `master` by both routes. Thank you for the analysis; the format and the `tz` parameter are yours.

lmeyerov:
Superseded: float in #2089, datetime in #2105. See the comment above for the disposition of each part.

Issue #1662, opened 2026-06-29
[FEA] GFQL/Cypher GRAPH subgraph projection for independent node/edge filters


lmeyerov:
**Is your feature request related to a problem? Please describe.**

StreamGL/viz filter compilation needs to express a common subgraph projection in one PyGraphistry GFQL/Cypher dispatch:

- apply node filters/exclusions to the node table
- apply edge filters/exclusions to the edge table
- drop edges whose endpoints are not in the filtered node table
- optionally prune isolated nodes
- preserve isolated nodes when prune-isolated is disabled

This is not a normal path query. A connected pattern like:

```cypher
MATCH (a)-[e]->(b)
WHERE f(a) AND f(b) AND g(e)
```

drops valid filtered nodes that have no matching edge. Today downstream code may work around this by running separate node and edge queries and merging results, but that creates more round trips, pushes graph semantics into the UI layer, and prevents PyGraphistry from optimizing the whole projection as one graph operation.

We need a first-class GRAPH/GFQL form and optimizer path for independent node relation filter + independent edge relation filter + endpoint reconciliation.

**Describe the solution you'd like**

Add a GRAPH/GFQL subgraph projection primitive, or support a canonical rewrite form, with semantics equivalent to:

```python
V0 = { n | node_pred(n) }
E0 = { e=(src,dst) | edge_pred(e) }

E1 = { e in E0 | e.src in V0 and e.dst in V0 }

if prune_isolated:
    V1 = { n in V0 | degree_E1(n) > 0 }
else:
    V1 = V0

return Graph(nodes=V1, edges=E1)
```

Important: when `prune_isolated=false`, nodes in `V0` must be retained even if no edge in `E1` touches them.

Candidate syntax, to refine:

```cypher
GRAPH {
  NODES (n)
  WHERE f(n)

  EDGES (a)-[e]->(b)
  WHERE g(e)

  EDGES BETWEEN SELECTED NODES
  PRUNE ISOLATED false
}
```

Equivalent UI compiler intent:

```text
node_pred = node_filters AND NOT node_exclusions
edge_pred = edge_filters AND NOT edge_exclusions
endpoint_policy = between_selected_nodes
prune_isolated = user_toggle
limit = user_toggle / existing limit behavior
```

The exact syntax is open, but the engine should expose one canonical target form for streamgl-viz and similar clients to generate.

**Implementation notes / proposed tasks**

1. Define GRAPH projection semantics
- Decide canonical syntax or API form.
[...]

PR #1706, opened 2026-07-07
fix(gfql): apply GRAPH residual predicates as masks

lmeyerov:
## What

Fixes `GRAPH { MATCH ... WHERE ... }` residual predicate handling for the safe graph-mask cases instead of silently dropping the predicate or merely raising.

This PR compiles single-alias residual predicates into explicit graph residual-mask metadata, then applies those masks to the target node or edge frame before the graph-state `Chain` runs. That covers residual forms such as OR/NULL expressions and `searchAny(...)` for a single node or edge alias.

Pattern-predicate and multi-alias residuals still fail loudly because they need first-class projection semantics rather than a local entity mask. Polars also declines these GRAPH residual masks for now instead of silently bridging or mis-evaluating.

## Why

Issue #1662 exposed a silent wrong-results bug: GRAPH constructors could accept row residual predicates from Cypher lowering, but `Chain(lowered.query, where=lowered.where)` had nowhere to apply them.

The root fix is to preserve the residual and apply it at graph state construction time when it is semantically a node or edge mask.

## Handoff / backport notes

Use the net PR diff as the source of truth; the first commit (`b1903892`, fail-fast) was intentionally superseded by the later mask implementation. If cherry-picking into a branch that has moved ahead, prefer either cherry-picking the full stack or reapplying the net diff for these files:

- `graphistry/compute/gfql/cypher/lowering.py`: adds `CompiledGraphResidualFilter`, records it on compiled query/binding structs, and compiles only safe single-alias node/edge residuals.
- `graphistry/compute/gfql_unified.py`: applies residual masks before `_chain_dispatch`; pandas/cuDF path is implemented, Polars raises a clear validation error.
- `graphistry/tests/compute/gfql/cypher/test_lowering.py`: semantic contract for node residuals, alias/property-column collision regressions, cuDF smoke, Polars honest decline, `searchAny`, edge residuals, named `GRAPH ... USE`, and pattern-predicate rejection.
- `docs/source/gfql/cypher.rst`: user-facing supported/unsupported behavior.
- `bin/ci_cypher_surface_guard_baseline.json`: intentional internal compiled-struct surface growth.

Backport gotchas:

- Do not port only the fail-fast guard; that was the rejected interim state.
- Keep `graph_residual_filters` on both standalone `CompiledCypherGraphQuery` and named `CompiledGraphBinding`; otherwise `GRAPH g = GRAPH { ... } USE g ...` regresses.
- Keep endpoint reconciliation for node masks: after filtering nodes, filter edges to endpoints in the surviving node IDs before running the Chain.
- Keep the residual eval-frame alias marker unconditional; otherwise a real node/edge property column named like the alias (`a`, `r`, etc.) breaks alias-property residuals such as `a.score` / `r.weight`.
- Keep unsupported pattern/multi-alias residuals loud. They are not equivalent to a simple entity mask.
- If the target branch already has a native Polars row-expression graph-mask path, replace the honest Polars decline with parity tests; otherwise keep the decline test.

## Surface change

The compiled Cypher structs gain one internal metadata field for GRAPH residual masks. The Cypher surface guard baseline was regenerated intentionally for:

- `CompiledCypherQuery`: 7 -> 8 fields
- `CompiledGraphBinding`: 6 -> 7 fields
- `CompiledCypherGraphQuery`: 6 -> 7 fields

## User impact

- Native GRAPH predicates continue to work.
- Single node/edge residual predicates inside `GRAPH { }` now filter the returned subgraph on pandas/cuDF.
- `searchAny(...)` works in the supported single-alias GRAPH mask form.
- Unsupported pattern/multi-alias residuals still fail clearly instead of returning wrong results.
- Polars GRAPH residual masks fail clearly pending native support.

## Validation

- `python bin/ci_cypher_surface_guard.py` -> passed
- `python -m pytest -q graphistry/tests/compute/gfql/cypher/test_lowering.py -k 'graph_constructor or graph_binding_applies_residual_mask_before_use'` -> 17 passed
- `python -m pytest -q graphistry/tests/compute/gfql/test_engine_polars_conformance_matrix.py -k 'prune_isolated_graph_shape'` -> 1 passed
- `python -m pytest -q graphistry/tests/compute/gfql/test_viz_pipeline_conformance.py -k 'graph_state_prune_pipeline'` -> 1 passed
- `python -m ruff check graphistry/compute/gfql/cypher/lowering.py graphistry/compute/gfql_unified.py graphistry/tests/compute/gfql/cypher/test_lowering.py` -> passed

Refs #1662


docs/source/gfql/cypher.rst
```diff
@@ -92,6 +92,13 @@ state instead of a row table:
     subgraph._nodes
     subgraph._edges
 
+``GRAPH { MATCH ... WHERE ... }`` accepts native graph-state predicates
+that lower to chain node/edge filters. Single node- or edge-alias residual
+predicates such as disjunctions and ``searchAny(...)`` are applied as GRAPH
+subgraph masks before the match chain runs. Pattern-predicate and multi-alias
+residuals remain unsupported inside ``GRAPH { }`` until first-class independent
+node/edge subgraph projection covers their semantics.
+
```
lmeyerov, line 106: rewrite & reexamine semantics

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -176,6 +176,14 @@ class CompiledCypherExecutionExtras:
     logical_plan_defer_code: Optional[str] = None
 
 
+@dataclass(frozen=True)
+class CompiledGraphResidualFilter:
```
lmeyerov, line 180: comment

PR #1704, opened 2026-07-06
bench(gfql): fair same-HW cross-DB competitor matrix runners (Kuzu/Memgraph/Neo4j/Pokec)

lmeyerov:
## What
Consolidates the competitor benchmark harnesses onto the index/bench stack (they were misplaced on the cypher-conformance branch). Each competitor runs via **its own** `prrao87/graph-benchmark` repo query — the fairness fix (our hand-written queries undersold Kuzu ~90×).

## New runners
`graph_benchmark_{kuzu,memgraph_q1_q9,memgraph_neighbors,neo4j_q1_q9,pokec}.py`, `graph_benchmark_{compare,polars_q5_q7}.py`, `graph_benchmark_{expansion,ldbc,neighbors,path,pattern,planner}_probe.py`, `run_graph_benchmark_memgraph_dgx.sh`, `docs/source/gfql/benchmark_memgraph.rst`. `graph_benchmark_q1_q9.py` gains the **q1 columnar in-degree shortcut** + per-query policy metadata. README/RESULTS/graph_benchmark.md union-merged with the stack's index-harness docs.

## Scope
**Bench-only** — no `graphistry/` source (the shortestPath-cache perf commit is split to a separate PR). Runnable on dgx via `dgx-guard/safe_run.sh`.

## Fair-matrix findings (dgx, q1–q9, 100k/2.42M FOLLOWS)
- Real repo-query medians: **Kuzu q8 10.4ms / Ladybug q8 19.5ms are genuine WINs** (factorized WCOJ path-count).
- GFQL wins q1 (cuDF **19.7ms** via the shortcut, beats Kuzu 253ms) / q2 / q4–q7 / q9 (GPU).
- Pokec seeded (Memgraph's benchmark): **GFQL wins 8/8** via native indexed hop (#1658 payoff).
- Pending in follow-ups: Neo4j+Memgraph repo-query runs, Lance-graph, GFQL polars/polars-gpu, seeded across DBs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1725, opened 2026-07-11
bench(gfql): add typed Pokec Polars lanes

lmeyerov:
## Summary

- add explicit pandas, cuDF, Polars, Polars-GPU, legacy `both`, and four-engine `all` Pokec selections
- convert resident graph frames once outside measured queries and keep Polars-GPU frames as ordinary Polars
- retain measured samples, per-seed distributions, cardinality, units, planner paths, execution target, CUDA synchronization, and unsupported/error/OOM/fallback states
- classify indexed Polars-GPU traversal honestly: NumPy index paths are CPU fallback and cost-gated index/scan tasks are mixed, never GPU-only

## Why

The Pokec runner only exposed pandas/cuDF, which left two GFQL engine lanes missing from the per-task report. A naive Polars-GPU addition would also mislabel the physical adjacency index path as GPU even though that index uses host NumPy arrays. This change adds the lanes while making execution provenance and failure states publication-safe.

## Validation

- review skill: 8 waves after runtime amplification, converged with 0 unresolved BLOCKER/IMPORTANT findings
- local static: Python 3.8 parse, `py_compile`, Ruff, focused mypy, `git diff --check`, credential scans
- guarded DGX exact head `52a22dd93888208eed0dcd22d2cb693ee09f3a55`: focused tests `11 passed, 6 subtests passed`
- guarded Pokec medium pilots and full R1: Polars indexed and genuine Polars-GPU Cypher each pass all 8 supported seeded tasks with exact prior cardinality parity; `pattern_cycle` remains explicitly unsupported
- full R1 configuration: 30 fixed seeds, 1 warmup, 5 measured samples; 80 GB RMM cap and 20 GB host floor

## Evidence caveat

Polars-GPU with native indexed traversal is retained as an explicit CPU/mixed fallback receipt. Genuine GPU seeded results use the separate `index=off`, `traversal=cypher` policy and must not be compared as though it used the same physical index policy.

Stacked on #1704. Do not merge out of order.


lmeyerov:
Closing: this is benchmark harness, and under the repo SPLIT rule perf evidence and harness
live in **pyg-bench** while pygraphistry takes correctness and generalizable engine fixes only.

**No content is lost — the migration has already happened.** Verified file by file rather than
assumed:

| this PR | pyg-bench | status |
| --- | --- | --- |
| `benchmarks/gfql/graph_benchmark_pokec.py` (665 lines) | `ported/pygraphistry/benchmarks/gfql/graph_benchmark_pokec.py` | **byte-identical** |
| same | `benchmarks/graphbench/graph_benchmark_pokec.py` (the active copy) | identical + a portability fix (guards `_HERE.parents[2]` against `IndexError`) |
| `graphistry/tests/benchmarks/gfql/test_graph_benchmark_pokec.py` (235 lines) | `ported/pygraphistry/graphistry/tests/benchmarks/gfql/test_graph_benchmark_pokec.py` | **byte-identical** |

So the typed Polars / Polars-GPU lanes, the once-outside-the-measured-region frame conversion,
and the honest execution-provenance classification (NumPy index paths reported as CPU fallback
rather than mislabelled GPU) all exist in pyg-bench today.

**Closing the PR does not drop the surface.** Pokec remains on the benchmark portfolio. One
correction worth recording while closing, since it changes what that surface is worth: Pokec is
**not** currently a scoreable competitor surface. The only Pokec artifact on the perf host is a
2026-07-19 run whose top-level keys are `dataset, edges, gfql, gfql_index, gfql_traversal,
nodes, runs, seeds, warmup` — `gfql` holds pandas/cuDF/Polars/Polars-GPU and **there is no
competitor section at all**. It is a GFQL-vs-GFQL engine comparison. The dataset it referenced,
`/data/pokec_small_import.cypher`, is also gone from the host.

The surface itself is legitimate — the runner targets Memgraph's own advertised mgBench Pokec
workload, the same dataset and query text Memgraph markets itself on — but it has never been
run against the competitor. Scoring a Pokec cell means re-fetching the mgBench dataset,
standing up Memgraph, and running matched queries with row validation. That is real setup, not
a cheap win, and it belongs in pyg-bench.

Issue #1726, opened 2026-07-12
GFQL: #1658 index does not engage on polars/polars-gpu direct hops (frame-identity churn)


lmeyerov:
## Summary
The #1658 CSR adjacency index accelerates seeded `hop()` on pandas and cudf but **never engages on polars / polars-gpu** for direct hops: every hop falls back to the O(E) scan even when a resident, valid index was just built via `gfql_index_all(engine="polars")`. Correctness/parity is unaffected (the scan result is identical) — this is purely a lost fast path. cudf is affected only for a narrow cost-gate case (below); polars/polars-gpu miss on *every* covered hop.

## Evidence (index_trace, qualitative)
Wrapping identical single-seed single-hop queries in `index_trace()`:
- **pandas:** every hop records `path="index"` (the #1658 CSR engages).
- **cudf:** mostly `path="index"`; one low-cardinality single-hop query records `path="scan"` with reason `"frontier N >= 0.02*n_keys (...) -> scan cheaper"`.
- **polars / polars-gpu:** **all** hops record `path="scan"`, `used_index=False`, and the trace shows ~2 steps per hop (vs 1 on pandas).

The doubling is diagnostic: `maybe_index_hop` runs at two sites — the shared hook in `compute/hop.py` and again in `lazy/engine/polars/hop.py` (so `chain_polars` gets the hook too). On a hit the first call returns early (1 step); on a bail both fire (2 steps). So 2 steps == a confirmed bail.

## Root cause (polars/polars-gpu): edge-frame identity churn
`registry.get_valid()` validates a resident index primarily by frame **identity** (`idx.source_ref is df`). `hop()` runs `_coerce_input_formats(self, engine)` before the index hook. For native-polars frames that helper unconditionally re-wraps the edges/nodes through `_pl_nan_to_null`, and `_pl_nan_to_null` returns a **fresh frame object** whenever the frame has any float column (it keys on column presence, not on whether a NaN is actually present).

Because `gfql_index_all` builds three indexes (each coercing) and `.hop()` coerces once more, the live edges frame is a *different object* on every step. So each adjacency index's `source_ref` no longer matches the live edges frame → `get_valid` returns `None` → `_indices_for_direction` returns `None` → `index_seeded_hop` returns `None` → scan. pandas and cudf don't hit this: their coercion returns the *same* frame object (pandas identity / cudf→cudf identity); only the polars tail rebuilds.

[...]

PR #1733, opened 2026-07-19
fix(gfql): restore identity-stable _pl_nan_to_null (regression from #1731)

lmeyerov:
## Summary
Restores the identity-stable `Engine._pl_nan_to_null` from **#1726** (`3809ecd9`), which **#1731**'s "final hardening" (`c2138b0d`) accidentally reverted. The revert is **live in current master** and silently re-broke the **#1658 seeded-index** engagement on Polars / Polars-GPU for any graph carrying a float column.

## Root cause
`_pl_nan_to_null` returned a *fresh* polars frame whenever any `Float32/Float64` column existed (keyed on column *presence*, not actual NaN). `fill_nan(None)` on a NaN-free column is a value-level no-op but still yields a new object, and that identity churn defeats the #1658 CSR index's `source_ref is df` residency check — so a resident index is invalidated across a `df_to_engine` re-conversion and the query silently falls back to the O(E) scan.

#1726 fixed this by probing `is_nan().any()` per float column and returning the **same** object when no NaN is present. `c2138b0d` restored the unconditional `df.with_columns([fill_nan(None) ...])`, reintroducing the churn. This PR re-lands the identity-stable probe (rebuild only — and only the columns where — a NaN is genuinely present; the rare LazyFrame path keeps the unconditional rewrite).

## Impact
- **Regressed:** any graph with a float column on `engine='polars'`/`'polars-gpu'` (SNB/LDBC, graph-bench) — resident #1658 index invalidated across re-conversion → O(E) scan.
- **Not affected:** all-int/str graphs like Pokec (hit the `if not float_cols: return df` fast return).

## Validation (dgx-spark, `26.02-gfql-polars` image, via safe_run)
- **113 passed** — restored `TestPlNanToNull` (4 unit tests) + the full 4-engine `test_index.py` parity suite (pandas/cuDF/polars/polars-gpu).
- **Mechanism proof at the `df_to_engine` boundary:** regressed → `df_to_engine(polars→polars)` identity preserved **False**; fixed → **True**, with NaN-present input still converting to null (`null_count=1`). Value-identical to prior behavior; only object identity changes.
- Note: a `gfql_index_all` + `gfql_explain` path did not surface a difference by itself (it reuses the resident polars frame without re-conversion); the impact is confined to paths that re-cross `df_to_engine` (the SNB native-hop lane, per #1726's original measurement of 120/120-scan → 54-index).

## Change
- `graphistry/Engine.py` — restore identity-stable `_pl_nan_to_null` (+ comment noting the #1726→#1731→re-land history).
- `graphistry/tests/compute/test_engine_coercion.py` — restore `TestPlNanToNull` as a regression fence.
- `CHANGELOG.md` — Fixed entry.

Regression fix only; no behavior change for NaN-carrying frames or non-float frames.


lmeyerov:
Closing as subsumed by #1735 (perf/gfql-pl-nan-clean-cache): #1735 contains this PR's identity-stable `_pl_nan_to_null` restoration (the #1726 regression fix from #1731) verbatim in behavior, plus the recycle-safe clean-cache that removes the per-call O(E) NaN re-scan. Diff-verified: both touch only `graphistry/Engine.py` + `test_engine_coercion.py`, and #1735's Engine.py change is a strict superset (identity guard + weakref.finalize id-cache). Review/land #1735 instead — it is the bottom of the GFQL perf stack (#1735 → #1734 → Cypher-Search fast path).

PR #1735, opened 2026-07-19
perf(gfql): identity-stable _pl_nan_to_null + clean-cache (kills O(E) per-hop NaN re-scan on polars)

lmeyerov:
## Summary

Every `hop()` runs `_coerce_input_formats` → `_pl_nan_to_null`, which scanned `is_nan().any()` over **every float column** of the (unchanged) resident edge frame **on every call**. Repeated seeded hops on a resident graph — the seeded-Search / native-hop pattern — therefore grew **O(E) with edge count on polars/polars-gpu**, while pandas stayed flat. This is the dominant per-call cost that made polars/polars-gpu seeded LDBC Search grow with scale and lose to Kuzu.

## Change (`graphistry/Engine.py`, `_pl_nan_to_null`)

- Probe an eager polars frame for real NaN **once**; return a clean frame **UNCHANGED** (restores the #1726 identity guard that #1731 reverted, so frame-identity caches like the #1658 CSR index keep engaging).
- Rewrite only the columns that genuinely carry NaN (values identical to the old unconditional `fill_nan`).
- **Cache the id of clean frames** so later calls skip the O(E) probe — recycle-safe via `weakref.finalize` (a rebound/new frame re-probes → NaN→null semantics unchanged).

## Measured (dgx-spark)

Indexed polars `g.hop` on 8M edges with one float column: **33.8 ms (growing O(E)) → 0.20 ms FLAT (~140×, O(degree))**, matching pandas. pandas unchanged (flat). polars *no-float* hop still grows mildly (0.17→0.91 ms) — a smaller, separate per-call cost, out of scope here.

## Validation

- Regression: **184 passed** — full 4-engine `test_index.py` parity + `test_engine_coercion.py`. The one failure (`test_chain_dask_edges`) is **pre-existing on clean master** (RAPIDS-image dask issue; verified on the unpatched checkout).
- New `TestPlNanCleanCache`: NaN-present cleaned to null; clean float frame returned same-object + cached; distinct frames don't cross-contaminate (all 4 pass).

## Relationship to #1733

Subsumes PR #1733 (the #1726 identity-stable reland) — this does the identity-stable behavior **and** adds the clean-cache. #1733 can be closed in favor of this, or this rebased on it.

## Impact

Makes polars/polars-gpu seeded Search O(degree) on float-column (i.e. real) graphs — the contained lever toward a 4-engine LDBC Search win. Broadly speeds all polars Search that re-hops a resident graph.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1734, opened 2026-07-19
perf(gfql): engage #1658 index for seeded gfql()/Cypher chains on ALL 4 engines (incl. typed edges)

lmeyerov:
## Summary

A seeded hop expressed as a `gfql([...])`/Cypher **chain** (rather than a direct `g.hop(...)`) previously **always fell back to the O(E) scan**, even with a resident #1658 adjacency index. Root cause: both chain executors attach a synthetic per-edge id column to the edge frame (pandas/cuDF eager `_chain_impl`: `copy(deep=False)`+assign; polars native lazy executor: `with_row_index`) → a fresh edge-frame object → the index's `source_ref is df` identity guard missed → `index_seeded_hop` returned `None` → scan. This is the material Search lever for real Cypher/LDBC workloads (seeded traversals expressed as chains with typed edges).

## Changes

- **`registry.rebind_edges(new_edges)`** — re-points the edge adjacency indexes' identity guard at the chain's augmented frame. Safe: the shallow copy / `with_row_index` preserves the indexed `src`/`dst` columns by value, and the structural fingerprint (rows + bound columns + engine) is unchanged by an added column. NODE_ID left untouched. Applied at **both** augmentation sites: `compute/chain.py` (pandas/cuDF eager) and `lazy/engine/polars/chain.py` (polars/polars-gpu native lazy).
- **Typed edges** — `_hop_is_index_coverable` now accepts a **simple scalar-equality `edge_match`** (`-[:KNOWS]->` / `e_forward({"type": X})`) on the wavefront path; `index_seeded_hop` applies the edge predicate to the CSR-matched rows each hop, **parity-exact** with the scan's `filter_edges_by_dict`.
- **Deliberately NOT covered (stay on scan, no over-reach):** predicate/membership-list `edge_match`, `edge_query`, and the direct-hop (non-wavefront) path.

## Validation (dgx-spark, 500k nodes / 4M edges / 4 edge-types, warm median-of-7; every cell verified index result == scan result)

| shape | pandas | cuDF | polars | polars-gpu |
|---|---|---|---|---|
| typed 1-hop | **2.1×** | **1.9×** | **1.3×** | engaged (1.0×) |
| typed 2-hop | **1.4×** | **2.0×** | **1.4×** | **1.3×** |
| untyped 1-hop | **2.7×** | **2.0×** | **5.4×** | **12.4×** |
| membership-list (control) | scan ✓ | scan ✓ | scan ✓ | scan ✓ |

[...]

PR #1736, opened 2026-07-19
feat(gfql): polars-native connected OPTIONAL MATCH (fixes pandas-ism crash + unseeded-arm dispatch + polars join twin)

lmeyerov:
## What

Stack PR #3 (base: #1734 `feat/gfql-typed-edge-index-coverage` -> #1735 -> master). Fixes the first of the two polars blockers found by the 2026-07-19 official-Cypher LDBC baseline (`plans/gfql-benchmark-numbers/cypherbase-sf01/`):

`MATCH ... OPTIONAL MATCH ...` (LDBC IS7 / message-replies shape) on `engine='polars'` crashed with `AttributeError: 'DataFrame' object has no attribute 'dropna'` — `_optional_arm_start_nodes` in `gfql_unified.py` ran pandas-only frame ops (`.dropna()/.drop_duplicates()/.rename(columns=)`/boolean-mask indexing) on the joined binding rows, which are polars frames on the polars engine.

## Fix

Branch on frame type in the seed extraction: polars uses `drop_nulls().unique().rename({...})` + `filter(col.is_in(...))`; the pandas/cuDF path is unchanged. After this the polars row pipeline can still honestly decline the full IS7 query (parity-or-error `rows` NIE — the second blocker, next stack PR), but never crashes with a pandas-ism.

## Validation (dgx, 26.02-gfql-polars container)

- Minimal IS7-shaped repro: AttributeError before -> clean run to the row pipeline after (traceback-verified at `gfql_unified.py:380`).
- `test_runtime_physical_cutover.py` + `test_engine_polars_cypher_conformance.py`: **167 passed**.
- New `test_optional_match_polars_frames.py`: pandas oracle (row values + knows-flag) + polars no-pandas-ism contract. **2 passed**.

## Why it matters

LDBC cypher-surface coverage on polars is 6/20 with 5 more queries blocked only by the polars `rows` NIE; this crash was the one *dishonest* failure (a bug, not a decline). Fixing it is a prerequisite for the `rows` native-lowering PR that unlocks the polars Cypher lane.


lmeyerov:
Extended beyond the dropna fix — **connected OPTIONAL MATCH is now polars-NATIVE end-to-end**, not just crash-free:

1. **Engine-polymorphic seed extraction** (the original AttributeError fix at `gfql_unified.py:380`).
2. **Unseeded-arm dispatch on polars**: `start_nodes` is a pruning hint (arm rows are left-outer-joined on shared aliases; `where`-arms already ran unseeded), but it forced the polars bindings-row path into its bounded-reentry pandas-only decline → NIE on queries polars can run natively. Polars engines now dispatch the arm unseeded.
3. **Polars twin of the arm join/alias-synthesis block**: `is_in` semi-join prune, `join(how='left')`, null-preserving alias synthesis.

Probe-verified on dgx: IS7-shaped `MATCH ...OPTIONAL MATCH ... RETURN ... ORDER BY` returns oracle-exact ordered rows on `engine='polars'` (previously: AttributeError crash; after fix 1 alone: honest NIE). Full IS7's `CASE r WHEN null` projection still declines honestly (`select` NIE — polars CASE lowering is a separate feature, next stack PR). **184 tests pass** (cutover + polars conformance + logical planner + 3 new regression tests). Perf follow-up noted in-code: optional seeded pruning for polars via first-alias membership injection if arm fan-out ever dominates.

lmeyerov:
Third commit: **full LDBC IS7 now runs natively on polars, oracle-exact** — the last blocker was the simple-CASE marker `__cypher_case_eq__(x, null)` (from `CASE r WHEN null THEN false ELSE true END`) declining in the polars function lowerer. Lowered the null-literal forms as `is_null()` (exact pandas null-mask parity); the general `CASE x WHEN v` form still declines honestly (it carries pandas' bool/numeric cross-dtype rules — separate follow-up). Regression test upgraded to a discriminating graph (creator KNOWS one replier, not the other → flags [False, True] pinned on pandas AND polars). dgx validation: 184 cutover/conformance/planner tests + 2 CASE mixed-dtype tests pass.

lmeyerov:
Fourth commit — the noted perf follow-up, measured: **polars OPTIONAL MATCH arm pruning via first-alias id-membership injection**. The polars arm previously ran unseeded (correct but computed the full pattern before the outer join). The same restriction `start_nodes` provides on pandas is now expressed engine-natively: the shared first alias's bound ids become a membership filter on the arm's first node op (guarded: only when that op has no `query` and no existing id constraint; unseeded fallback otherwise). **LDBC SF0.1 harness: IS7 message-replies cypher on polars 710.8 → 77.2 ms (9.2×)** — vs ast-lane 39.6 ms and matched Kuzu 24.2 ms. 184 cutover/conformance/planner tests pass. PR now = 4 commits: crash fix → polars-native arm+join → CASE-null lowering → arm pruning.

PR #1737, opened 2026-07-19
feat(gfql): polars whole-entity aggregation Cypher — HAS_label disambiguation + identity-key + key_prefixes (LDBC IC4 native)

lmeyerov:
## What

Stack PR #5 (base: #1736 → #1734 → #1735 → master). Unlocks the polars `rows`/aggregation NIE class found by the 2026-07-19 LDBC cypher baseline: the official **IC4 (new-topics)** query now runs natively on polars, parity-exact — comma-MATCH, `WITH DISTINCT`, CASE projections, whole-entity `sum` aggregation, HAVING-style WHERE.

## Three pandas-parity ports (all previously honest NIEs)

1. **HAS_<Label> destination disambiguation** (`binding_rows_polars`): instead of declining every labeled-destination pattern, apply pandas' `_gfql_disambiguate_has_edge_destination_nodes` rule — an unlabeled next-node op after a `HAS_<Label>`-typed edge narrows to that label ONLY when candidate node ids collide across labels (unique ids keep the unnarrowed domain, exactly like pandas).
2. **Whole-entity identity key**: `alias.__gfql_node_id__` (#1650 grouping key) resolves to the bare `alias` id column — the polars bindings table intentionally omits pandas' join-residue columns; the bare alias column IS the identity key.
3. **`group_by(key_prefixes=...)`**: expand every `<prefix>*` row-table column into the key set exactly like pandas (functionally dependent on the identity key → group sizes unchanged).

## Validation (dgx, 26.02-gfql-polars)

- Harness (pyg-bench snb-interactive, SF0.1, official query texts): **new-topics cypher on polars = ok, 127.8 ms, expected-rows parity** (was NIE). Polars cypher coverage 7/20 → **8/20**.
- `test_runtime_physical_cutover` + polars conformance + logical planner: **184 passed**.
- New `test_polars_rows_entity_groupby.py`: pandas oracle + polars parity on a discriminating fixture (tag invalidated by a pre-window post). **2 passed**.

## Not in scope (named follow-ups)

Undirected bounded var-length `-[:KNOWS*1..2]-` (IC6/IC11 — immediate-backtrack avoidance port), zero-hop unbounded varlen (IS6, `feat/issue-983` branch), shortestPath scalar backend (IC13).


lmeyerov:
Self-review checkpoint (per .agents/skills/review discipline) caught **two parity divergences** in the initial HAS_label port vs pandas' actual call site (`row/pipeline.py:3670`): (1) pandas probes id duplicates on the **reached candidate domain** (candidate_source ∩ wavefront) — the initial port probed the full node table, which would over-narrow when ids collide globally but not among reached nodes; (2) pandas gates the narrowing to **forward single-hop** edges — the initial port applied it to all directions. Fixed: the polars twin now runs after the hop's state semi-join with identical gates, and re-semi-joins the state when it narrows. Two new engine-parametrized regression tests pin both branches against the pandas oracle (skip-narrowing on unique-reached; narrow on reached-collision). 186 suite + 6 file tests pass on dgx; IC4 harness parity re-confirmed.

lmeyerov:
## Review-skill pass (3 dimensions + adversarial seam re-review) — ready for human review

**BLOCKER found & fixed (`2cdc9c33`):** the HAS_`<Label>` disambiguation gate probed the chain-combine node table, which is already id-deduplicated — the duplicate probe could never fire (coverage-verified dead code), and with the old NIE decline removed, duplicate-id graphs silently returned **row-order-dependent** first-occurrence rows where pandas narrows to the label row (reproduced end-to-end both row orders). Since chain-combine has already dropped the label rows pandas narrows to, a faithful native answer is impossible from the deduped table. Fix: probe the **pre-chain base graph** (`_gfql_rows_base_graph`) — unique-id graphs stay native (pandas would not narrow there either, structurally + 90-comparison differential fuzz; **IC4 keeps its native win**), duplicate-id/unknown-base graphs decline to the honest NIE. New tests reach the actual gate (coverage-confirmed), incl. dup-id NIE both row orders + unique-id parity.

**Also from review:**
- nan_clean soundness comment corrected — polars *does* have in-place mutation (`extend`, `replace_column`, `insert_column`, `hstack(in_place=True)`); the real contract is "don't mutate frames after handing them to GFQL" (same as the pandas-input paths).
- `AggSpec` import made runtime (TypedDict now `get_type_hints`-introspectable; safe without polars — verified under an import blocker).
- `lower_select_items`/`lower_order_by_keys` return `List[pl.Expr]` (consistency with the PR's own typing sweep).
- key_prefixes genuine-expansion + key-overlap test (previous unit test's prefix matched zero columns); malformed-AggSpec decline tests.
- Old "narrowing" tests kept but relabeled honestly: their single-MATCH shape routes through the plain chain executor, not `binding_rows_polars`.

**Filed #1739** (pre-existing on master, A/B-verified): the single-MATCH *aggregating* HAS_`<Label>` shape on duplicate-id graphs silently diverges on polars via a different path — out of this PR's scope.

**Seam re-review verdict: APPROVE** (6/6 probes pass; full local gfql lane 4443 passed, failures = known environmental set only). CI green at `2cdc9c33`; #1738 re-rebased on top (`551cc63e`), also green.

Non-blocking follow-ups noted: exercise the two defensive decline lines (base-graph-absent / id-col-absent), and a pre-existing LazyFrame-input crash at `chain.py:720` (before this gate) deserves its own issue.

graphistry/compute/gfql/lazy/engine/polars/row_pipeline.py
```diff
@@ -27,6 +27,19 @@
...
+# ``List``; Python-side lowering may pass tuples. Shapes are safelist-validated
+# (gfql/call/validation.py) before reaching these helpers, so the runtime isinstance/len
+# checks below are defense-in-depth, not the contract.
+SelectItem = Union[str, Tuple[str, JSONVal], List[JSONVal]]
+"""Projection item: ``'col'`` | ``(alias, expr)`` where a str expr is expression text and any
+other JSON scalar is a constant literal (e.g. synthetic ``__cypher_group__=1``)."""
+OrderKey = Union[Tuple[str, str], List[str]]
+"""Sort key: ``(expr, 'asc'|'desc')``."""
+AggSpec = Union[Tuple[str, str], Tuple[str, str, Optional[str]], List[Optional[str]]]
+"""Aggregation spec: ``(alias, func[, expr])``; ``expr`` None/omitted only for count(*)."""
```
lmeyerov: factor out?

PR #1738, opened 2026-07-19
[EXPERIMENTAL] perf(gfql): seeded-chain fast path — slices 1-2 shipped, slice 3 (lean combine) in design

lmeyerov:
## What

Stack PR #6 (base: #1737 → #1736 → #1734 → #1735 → master). First slice of the evidence-driven Cypher-Search fast path (plan.md 'FAST-PATH EVIDENCE'): eliminates the two per-call **O(E)** costs that made seeded Cypher latency GROW with edge count (LDBC SF0.1→SF1 grew 4-5× on 11× data; Kuzu stays flat).

1. **Typed keep-mask cache** (`gfql/index/traverse.py`): the #1658 index path rebuilt its simple-equality edge keep-mask via a full column compare (`eq_str`, ~2.8ms/1M rows) on EVERY seeded hop. Now cached per (resident frame, engine, scalar edge_match), recycle-safe via `weakref.finalize` (the #1735 clean-cache pattern).
2. **Augmented-frame cache** (`lazy/engine/polars/chain.py`): the polars chain rebuilt `with_row_index` per call — an O(E) copy whose fresh object identity ALSO defeated every downstream id-keyed cache (including #1). One augmentation per resident frame; stable identity restored.
3. (Also: a typed-edge slice cache in `binding_rows_polars` for the bindings-row path, same pattern.)

## Evidence

- Before: cProfile of official IS1 on 200k/1M polars graph showed `eq_str` 2.8ms/call + `with_row_index` per call; warm median 18.7-23ms.
- After: **both eliminated from the warm profile**; remaining costs are ~25 lazy collects (~7ms — slice 2/2: plan fusion) + ~4.2ms Cypher re-compile (slice 3: plan cache).
- Parity: **274 tests pass** (138 index 4-engine + 136 polars cypher conformance), zero changes.

## Safety of the caches

Pure functions of the resident frame; eviction on frame GC via `weakref.finalize` so a recycled id can never serve stale data while the source frame is alive; rebinding `g.edges(...)` yields a new frame object → new key → recompute (semantics unchanged).


lmeyerov:
Closing this as a **quarry, not a landing branch**, after auditing it against #1757.

## They are the same work, diverged

#1757's own description says it was extracted from this PR (`ee510f78`). Verified against the current branches: they overlap on **5 files including all three cache sites** — `gfql/index/traverse.py`, `lazy/engine/polars/chain.py`, `lazy/engine/polars/row_pipeline.py`, plus the chain test and CHANGELOG. This branch carries 10 further commits and 2 more files (`filter_by_dict.py`, `polars/predicates.py`).

Leaving both open means neither gets measured and any benchmark number is ambiguous about which tree produced it. #1757 is the reviewed, tested slice, so it is the only one eligible to proceed.

## What that costs, stated plainly

This branch's later slices are **not** being written off — 4a (stop re-resolving lazy schemas per predicate) and 4b (fuse the per-column temporal-constructor probe) target real per-call overhead. They are re-extracted as clean, individually **measured** PRs if and when the lane justifies them, rather than riding along inside an experimental branch nobody can attribute a number to.

## Why not simply land the whole thing

Independent of the duplication, the shared cache design does not currently meet the bar for landing anywhere: three **module-level global dicts** holding **data-plane** objects (an O(E) boolean mask, a full `with_row_index` edge frame, a full eager typed slice), with **no locking** and **no bounded size**. Eviction relies solely on `weakref.finalize` on the source frame, so N live graphs imply N x O(E) retained copies plus one slice per distinct edge type.

The correctness argument is the sharper problem: "a recycled `id()` can never serve a stale entry" holds under prompt refcount collection, but a frame in a reference cycle is collected by gc at an arbitrary later time, leaving a window where a new frame reuses a freed `id()` and is served **another frame's cached data**. That is a silent-wrong-answer path, not just a leak.

Preconditions for any version of this work: bounded size, explicit locking, and a key stronger than `id()` — or, better, per-graph state reusing the existing `_gfql_index_registry` (which already carries fingerprint + object-identity validation) instead of a second, weaker mechanism.

Tracking continues on #1757 and #1755.

Issue #1741, opened 2026-07-20
GFQL polars chain: undirected bounded varlen (-[:T*1..k]-) silently diverges from pandas — missing backtrack avoidance includes the seed


lmeyerov:
## Summary

On master (c355660a), the native polars chain executes undirected bounded variable-length hops (\`-[:KNOWS*1..2]-\`) **natively but with wrong results**: it includes nodes reachable only by immediately re-traversing the same undirected edge (seed bounce-back), which pandas (the parity oracle) correctly excludes via relationship-uniqueness (trail) semantics. This violates the parity-or-honest-NIE contract — the shape should either match pandas or decline.

## Repro (master c355660a, polars 1.42)

```python
import polars as pl, graphistry
nodes = pl.DataFrame({"id": [f"p{i}" for i in range(6)], "label__Person": [True]*6})
edges = pl.DataFrame({"s": ["p0","p1","p2","p0","p3"], "d": ["p1","p2","p3","p4","p5"], "type": ["KNOWS"]*5})
g  = graphistry.nodes(nodes, "id").edges(edges, "s", "d")
gp = graphistry.nodes(nodes.to_pandas(), "id").edges(edges.to_pandas(), "s", "d")
q = "MATCH (p:Person {id: 'p0'})-[:KNOWS*1..2]-(f:Person) RETURN f.id AS fid ORDER BY fid"
print(gp.gfql(q, engine="pandas")._nodes.to_dict("records"))
# [{'fid': 'p1'}, {'fid': 'p2'}, {'fid': 'p4'}]
print(g.gfql(q, engine="polars")._nodes.to_pandas().to_dict("records"))
# [{'fid': 'p0'}, {'fid': 'p1'}, {'fid': 'p2'}, {'fid': 'p4'}]   <-- p0 = seed via edge backtrack
```

The path p0-[e]-p1-[same e]-p0 reuses one undirected edge, so p0 must not appear at *1..2 (pandas agrees; Cypher trail semantics).

## Impact

- LDBC IC6/IC11 shapes (\`-[:KNOWS*1..2]-\`) — the official texts currently decline later at the rows op, so benchmark numbers are honest NIE, but any user query of the bare shape returns silently wrong rows on engine="polars"/"polars-gpu".
- The undirected multihop path is marked native in \`_is_native_multihop\` (lazy/engine/polars/chain.py); the forward re-exec recompute does not exclude same-edge backtracking for undirected traversal.

## Fix directions

[...]

PR #1742, opened 2026-07-20
fix(gfql): decline undirected varlen + node alias on polars chain (honest NIE, #1741)

lmeyerov:
## Summary

Fixes the silent-wrong-results half of #1741: on `engine='polars'`/`'polars-gpu'`, `MATCH (a {..})-[:T*1..2]-(f) RETURN f.id` returned rows pandas (the parity oracle) correctly excludes.

- Root cause: undirected multi-hop on the polars chain lacks pandas' backtrack avoidance. The node/edge **sets** still agree (reachability), but the **named-alias flag** on nodes after the undirected varlen edge tags nodes whose only path bounces back over a single edge — e.g. the seed at `*1..2`. Cypher `RETURN` projects those flags → silent wrong rows.
- Fix (interim, integrity-first): decline the affected shape — undirected non-single-hop edge + any node alias in the chain — with a `NotImplementedError` referencing #1741, per the parity-or-honest-NIE contract. Unnamed undirected varlen chains keep native execution (their parity is pinned by a new test).
- The real backtrack-avoidance port (which also unlocks LDBC IC6/IC11 on polars) stays tracked in #1741.

## Tests

- `TestUndirectedVarlenAliasNIE`: NIE on the named shape; the exact #1741 cypher repro now declines instead of returning a wrong row set (pandas oracle result pinned); unnamed undirected varlen keeps native parity.
- Full polars chain suite: 994 passed locally.

### DIFF
--- a/graphistry/compute/gfql/lazy/engine/polars/chain.py
+++ b/graphistry/compute/gfql/lazy/engine/polars/chain.py
@@ -569,6 +569,21 @@ def _chain_traversal_polars(self: Plottable, ops, start_nodes: Optional[Any] = N
             "include_zero_hop_seed, prune_to_endpoints) require engine='pandas'."
         )
 
+    # issue #1741: undirected multi-hop lacks pandas' backtrack avoidance — the node/edge SETS
+    # still agree (reachability), but the named-alias flag on nodes after the undirected varlen
+    # edge can tag a node whose only path bounces back over one edge (seed at *1..2), which
+    # pandas correctly leaves untagged (trail semantics). Cypher RETURN projects those flags, so
+    # this silently returned wrong rows. Decline the affected shape (undirected non-single-hop +
+    # any node alias) until the backtrack-avoidance port lands; unnamed chains keep running.
+    if any(isinstance(op, ASTEdge) and not op.is_simple_single_hop()
+           and op.direction == "undirected" for op in ops) and any(
+               isinstance(op, ASTNode) and op._name is not None for op in ops):
+        raise NotImplementedError(
+            "polars chain engine: undirected variable-length hops with node aliases are not "
+            "yet supported (missing backtrack avoidance would mis-tag aliases — issue #1741); "
+            "use engine='pandas' for this query"
+        )
+
     edge_ops = [op for op in ops if isinstance(op, ASTEdge)]
     # Undirected edges in multi-edge chains: NATIVE for single-hop (backward pass threads BOTH
     # endpoints — see override below) and fixed-length multi-hop (generic backward hop +
--- a/graphistry/tests/compute/gfql/test_engine_polars_chain.py
+++ b/graphistry/tests/compute/gfql/test_engine_polars_chain.py
@@ -704,3 +704,34 @@ def test_engine_polars_predicate_correctness_fixes():
     # temporal val -> _cmp_expr declines (None) so upstream raises honest NIE; numeric still lowers.
     assert _cmp_expr(None, operator.gt, datetime.date(2020, 1, 1)) is None
     assert _cmp_expr(pl.col("x"), operator.gt, 5) is not None
+
+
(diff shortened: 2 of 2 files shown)


lmeyerov:
## Reviewer pushback accepted — my first diagnosis was wrong, and the real fix is smaller than I claimed

@lmeyerov asked why we raise instead of implementing it. Re-investigated, and the framing in this PR's description ("missing backtrack avoidance") is **incorrect**. Direct A/B:

- `hop(undirected, hops=2)` and `hop(undirected, min_hops=1, max_hops=2)` from seed `p0` return **identical node sets on pandas and polars**: `[p0, p1, p2, p4]`. Both engines include the seed reached by bouncing back over one edge. **Neither implements trail semantics in the traversal.**
- The final chain `_nodes` also match on both engines.
- The **only** divergence is the node-alias flag: pandas flags `f` on `[p1, p2, p4]`, polars on `[p0, p1, p2, p4]`. `RETURN f.id` projects that flag → the one bogus row.

**What pandas actually does** (`compute/chain.py` ~452-503): for a named `ASTNode` whose previous op is an `ASTEdge`, it narrows the alias with a **hop-distance gate** — `label_mask &= hop_col.notna() & (hop_col >= min_hop) & (hop_col <= max_hop)`, with min/max from the edge's `output_min_hops`/`min_hops`/`hops`. The seed carries hop 0, fails `>= 1`, and loses the alias. It is not backtrack avoidance at all. `_apply_node_names` in the polars chain has no hop gating whatsoever — that is the entire bug.

**Why it isn't a one-liner today**: the gate needs per-node hop distances (`label_node_hops`), which the polars chain currently NIEs for undirected varlen. So the real fix is:

1. Emit per-node hop distances for undirected bounded varlen in the polars hop — the layered BFS already knows each layer's distance, it just doesn't emit the column. This also removes an existing NIE.
2. Port pandas' gate into `_apply_node_names` (~20 lines).
3. Drop this PR's decline; its differential test flips from "raises" to "matches the pandas oracle" and becomes the parity gate.

**Proposed disposition**: keep this open only as stop-the-bleeding (master currently returns silently wrong rows, which is worse than an error). If (1)+(2) land promptly, **close this unmerged** and ship the real fix instead. Happy to go straight to the real fix if you'd rather not carry the interim decline at all — say the word.

Full diagnosis with the A/B evidence is recorded in `plans/gfql-benchmark-numbers/plan.md` so it survives context loss.

lmeyerov:
Superseded by the real fix: **#1746** (native `label_node_hops` in the polars hop) + **#1747** (the hop-distance alias gate, stacked on it). The repro now matches pandas exactly through the Cypher surface — `{p1, p2}` on both engines — instead of declining.

Leaving this open only until #1747 merges, since master currently returns *silently wrong* rows for this shape and a decline is strictly better than that. Once #1747 lands I'll close this unmerged and flip its test from "raises" to "matches the pandas oracle".

lmeyerov:
## Superseded by #1747 — will close once that lands

This PR declines (`NotImplementedError`) the undirected varlen + node-alias shape. #1747 now **implements** the pandas hop-distance alias gate instead of declining it, so that exact shape returns the correct rows (`MATCH (a)-[*1..2]-(b)` no longer over-includes the backtracked-to seed) — verified 4-engine against the pandas oracle. Implementing beats raising here.

Keeping this open only as a safety net until #1747 merges; it will be closed at that point. No further work planned on it.

lmeyerov:
Closing now (not waiting for #1747 to merge): this PR touches the same two files (chain.py + test_engine_polars_chain.py) and the same undirected-varlen-alias shape as #1747, with the opposite approach — declining (NIE) what #1747 correctly implements. Two open PRs conflicting on the same lines is confusing and offers no benefit: #1747 is reviewed, cross-engine-verified (0 regressions), and CI-green. Reopen if #1747 is ever rejected. Superseded by #1747 (refs #1741).

PR #1746, opened 2026-07-20
feat(gfql): native polars label_node_hops on the plain BFS (#1741 groundwork)

### LINKED ISSUE #1741: GFQL polars chain: undirected bounded varlen (-[:T*1..k]-) silently diverges from pandas — missing backtrack avoidance includes the seed

lmeyerov:
## What

Implements native `label_node_hops` / `label_seeds` in the polars hop's plain (shortest-path) BFS. Previously **every** hop-label request raised `NotImplementedError` on polars.

## Why

This is step 1 of the real fix for #1741, replacing the interim decline in #1742.

#1741 is: on the polars engine, `MATCH (a)-[*1..2]-(b)` over-flags the alias — a seed re-reached by backtracking gets aliased where pandas leaves it unaliased. My first diagnosis ("polars lacks pandas' backtrack avoidance") was **wrong**: A/B shows both engines return identical node sets and identical final `_nodes`; only the alias flag diverges.

The actual pandas rule is a **hop-distance gate** (`compute/chain.py` ~452-503): for a named node preceded by an edge op, `label_mask &= hop >= min_hop & hop <= max_hop`. The seed's hop label is null, so it fails the gate and loses the alias. polars cannot run that gate because it has no hop labels at all — hence this PR.

Step 2 (separate PR) ports the gate itself into `_apply_node_names` and closes #1741; #1742 then closes unmerged.

## The rule, and why it is direction-dependent

pandas labels differently per direction, and that asymmetry *is* the #1741 behavior:

| direction | labeled set per hop | consequence |
|---|---|---|
| forward / reverse | every destination, first-wins (`hop.py:540`) | a seed re-entered at hop 1 **is** labeled 1 |
| undirected | destinations **minus everything already visited** | a seed re-reached by backtracking stays **null** |

`label_seeds` writes hop 0 for seeds and marks them seen (`hop.py:361-363`).

## Verification

Differential fuzz against the pandas oracle: **4800/4800 exact matches** over random graphs × {forward, reverse, undirected} × hops 1–4 × `label_seeds` {on, off}. Both asymmetric shapes are pinned as explicit named tests, not just fuzz coverage.

Full local run: 1277 passed across `test_engine_polars_hop.py`, `test_compute_hops.py`, `test_engine_polars_chain.py`. mypy clean, ruff clean. (Three `test_engine_polars_conformance_matrix.py` temporal failures in this environment reproduce identically on `origin/master` — pre-existing, unrelated.)

## Still declined — honest NIE, no pandas bridge

- **`label_edge_hops` at every shape.** With labels on, the pandas edge output *duplicates* an undirected edge traversed in both directions (448/4800 differential cases). That looks like a pandas artifact rather than intended semantics; baking it into polars is not parity worth having, and #1741 needs only node labels. Worth a follow-up issue on the pandas side.
- **Labels under `min_hops > 1`**, whose labels come from the layered backward walk (`hop.py:676-778`) — a different rule, not yet ported.

## Note

`test_polars_hop_unsupported_raises` loses its `label_node_hops` and `label_seeds` cases — they are now supported. `label_edge_hops` and `min_hops>1` stay in the decline list.


lmeyerov:
Cross-engine validation posted on the stack tip #1747: 4-engine differential sweep on dgx-spark, A/B vs clean master `c355660a` — **0 regressions, 70 of 98 mismatches fixed**, polars and polars-gpu identical, pandas/cudf untouched at 144/144. Residual 28 (forward `-[*2..3]->`) is pre-existing on master, filed as #1748.

lmeyerov:
## Restructured after review — the hop-label correctness fix now lives HERE

The #1746 review correctly flagged that this PR, at its own head, shipped **silently-wrong undirected hop labels** whenever a seed was filtered out of the frontier (`source_node_match`) or the walk ran in wave-front mode (`return_as_wave_front=True`) — the node/edge *set* matched pandas but the *labels* diverged. Confirmed by direct A/B against the pandas oracle on the pre-fix head `4c981e2a`:

- direction × hops × `return_as_wave_front` × `source_node_match` × `destination_node_match` × 10 graphs = 480 cases
- **before: 456 agree / 24 disagree** — every failure undirected + wave-front, seed re-labeled
- **after this commit: 480 / 480**

The undirected seed pre-seed fix (previously stacked in #1747) has been **moved down into this PR** so #1746 is correct standalone, and a second parity fix was added: a `label_node_hops` name that collides with an existing column is now redirected to `<name>_1` like pandas' `resolve_label_col`, instead of the polars left-join auto-suffix `<name>_right` / `DuplicateError`.

`TestHopLabelsDifferential` is amplified with exactly the seed-filter / wave-front / multi-seed axes that were previously unvaried, plus a collision test. Full polars hop suite: 325 passed, ruff clean.

PR #1759, opened 2026-07-21
perf(gfql): seeded typed-hop fast path — seeded 1-hop RETURN in ≤1ms (#1755)

lmeyerov:
## What

Closes the **seeded-Cypher abstraction tax** (#1755): a seeded typed 1-hop
query — `MATCH (m {id})-[:T]->(p) RETURN p`, and the native chain equivalent
`[n({id}), e(edge_match), n({type})]` — spent ~20–38 ms in chain machinery
(two-pass combine + full-column `filter_by_dict` type scans + rows-pivot
projection) despite the hop itself being sub-ms. This adds a pandas-only fast
path that numpy-reduces the graph to the seed's 1-hop neighborhood, landing the
same query in **≤1 ms** (measured 0.847 ms on a 50k-Person / 200k-Message
graph; native hop 0.553 ms).

Two layers:
- **native** (`_seeded_typed_hop_numpy_pandas`, wired into `_try_chain_fast_path`):
  scalar-filtered seeded typed 1-hop → a handful of numpy passes.
- **cypher** (`_execute_seeded_typed_hop_fast_path` in gfql_unified): the lowered
  `MATCH (m {id})-[:T]->(p) RETURN p` whole-row RETURN surface.

Both are **byte-identical to the full path by construction** and fall through
(return `None`) for anything outside the covered shape.

## Correctness / parity

- **Differential parity** (fast-on vs fast-off) byte-identical across
  labelled/unlabelled RETURN, no-match, and native forward/reverse/typed shapes.
- **Independent oracle**: returned nodes are checked against the creator set
  **hand-computed from the raw edge/node frames** — not merely fast==full, since
  the full path could share a bug. Native returns `{seed} ∪ creators` (both bound
  endpoints); cypher `RETURN p` returns creators only.
- **Decline coverage**: multi-alias, field projection, return-source, reverse
  (seed on return node), predicate node/edge filters, undirected, multi-hop, and
  **variable-length hops** all fall through to the full path (asserted, and the
  full-path result asserted byte-identical).
- **Zero regressions**: full `compute/gfql` + `test_chain` suite = 77 failed /
  5090 passed; clean-master baseline = 77 failed / 5069 passed (identical failure
  set — all pre-existing local polars-drift / GPU-lib). +21 net new passing tests.

## Notes

- pandas-only (cuDF / polars fall through to their own paths — no cross-engine
  change; verified no regression).
- `mypy --config-file mypy.ini` clean on both changed source files (helpers are
  typed with `Plottable`/`ASTNode`/`ASTEdge` + `Optional` returns).
- A late-caught bug is included as its own commit: the cypher gate originally
  accepted variable-length edges (one `ASTEdge`, multiple hops) and silently
  truncated them — now gated on `e1.is_simple_single_hop()`.


graphistry/compute/chain.py
```diff
@@ -673,6 +673,169 @@ def _chain_otel_attrs(
     return attrs
 
 
+def _seeded_typed_hop_numpy_pandas(
+    g: Plottable, n0: ASTNode, n2: ASTNode, e1: ASTEdge,
+    src: str, dst: str, node: str, direction: str,
```
lmeyerov: literal type

graphistry/compute/chain.py
```diff
@@ -673,6 +673,169 @@ def _chain_otel_attrs(
...
+        n2_ok = valid
+    keep = np.isin(esrc, valid) & np.isin(edst, valid) & np.isin(eto, n2_ok)
+    if not keep.all():
+        edges = edges[keep]
+        fep = np.unique(np.concatenate([edges[src].to_numpy(), edges[dst].to_numpy()]))
+        cand = cand[np.isin(cand_ids, fep)]
+    return g.nodes(cand).edges(edges)
+
+
+def _seeded_typed_return_dst_pandas(
```
lmeyerov: DataFrameT, not pandas

graphistry/compute/gfql_unified.py
```diff
@@ -976,6 +976,89 @@ def _run_logical_pass_pipeline(logical_plan: LogicalPlan, ctx: PlanContext) -> L
...
+    # would silently truncate it. Reuse the same canonical gate the native fast
+    # path uses (also rejects hop labels / output slicing / fixed-point). See
+    # test_engine_polars_chain::TestVarlenAliasHopGate.
+    if not e1.is_simple_single_hop():
+        return None
+    if call.function != "rows":
+        return None
+    node = getattr(base_graph, "_node", None)
+    src = getattr(base_graph, "_source", None)
+    dst = getattr(base_graph, "_destination", None)
```
lmeyerov: getattr

lmeyerov:
Review response: the chain.py/gfql_unified.py specialization move-out is delivered by the stacked #1762 (as noted). Added the missing CHANGELOG entry; rebased over the merged #1758.

PR #1760, opened 2026-07-21
perf(gfql): generalize seeded typed-hop fast path to polars/polars-gpu (#1755)

lmeyerov:
**Stacked on #1759** (base = `perf/gfql-seeded-typed-hop-fastpath-1755`). Review/merge #1759 first.

## What

Generalizes the #1755 seeded typed-hop cypher fast path (pandas-only in #1759)
to **polars and polars-gpu** via `_seeded_typed_return_dst_polars` — the same
seed-first reduction (seed out-edges → typed-edge filter → destination nodes)
expressed with polars filters. It hooks upstream at
`_execute_seeded_typed_hop_fast_path`, *before* the lazy pipeline, so none of the
woven `binding_rows_polars` / `hop_polars` / combine machinery changes.

## dgx measurement (26.02-gfql-polars, 50k Person / 200k Message, warm median)

`MATCH (m:Message {id})-[:HAS_CREATOR]->(p:Person) RETURN p`:

| engine | full | fast | speedup | parity |
|---|---|---|---|---|
| pandas | 40.6 ms | 2.0 ms | 20.3× | ✅ |
| polars | 13.7 ms | 3.44 ms | 4.0× | ✅ |
| polars-gpu | 24.6 ms | 2.47 ms | 10.0× | ✅ |
| cudf | 27.8 ms | (falls through) | 1.0× | ✅ |

The fast path now engages on pandas/polars/polars-gpu; cuDF falls through (the
generic chain fast path already helps it).

## Correctness

- **Dispatch keys off the actual frame type** (`is_polars_df`), not the requested
  engine — the WITH..MATCH reentry path can request `engine=polars` while handing
  a pandas-materialized intermediate graph, which would otherwise run polars ops
  on a pandas frame. (Caught by `test_with_match_reentry_differential_fuzz`.)
- **Differential parity** (fast vs full) byte-identical on polars, plus an
  **independent oracle** (hand-computed creators) and varlen-decline coverage.
- **Zero non-seeded regressions**: full compute/gfql + test_chain suite = 77
  failed / 5096 passed vs #1759 baseline 77 / 5090 (identical failure set — all
  pre-existing local polars-drift/GPU-lib), +6 net new passing tests.
- mypy clean on both changed source files.


lmeyerov:
Review responses:
- `_seeded_typed_return_dst_polars` `direction` is now the `Direction` Literal (was `str`), matching the pandas/cuDF helpers.
- CHANGELOG entry added.
- On location (lazy/engine/polars/ was floated): keeping it beside the other seeded helpers (moved to `chain_fast_paths.py` by the stacked #1762) rather than under `gfql/lazy/engine/polars/` — it is a **chain-level fast path** dispatched from the cypher executor pre-pipeline, not part of the lazy row-pipeline engine; placing it in the lazy engine would create a back-edge from the engine into chain/AST internals and break the one-directional module-extraction convention (#1731).

PR #1763, opened 2026-07-21
perf(gfql): native polars translation for simple connected-join residuals (#1729/#1755)

lmeyerov:
## What

The connected-join grouped-count fast path applies scalar residuals — the shapes
#1729's lowering cannot push into `filter_dict` (`toLower(a.col) = toLower('lit')`,
`a.col <op> literal`) — by spinning a `where_rows` chain per alias (~1.7 ms each).
On graph-bench q5/q6/q7 that residual-apply was ~5 ms of a ~10 ms query.

On polars, translate exactly those simple shapes directly to native polars filter
expressions (`_residual_polars_expr`). **Any** expression outside the recognized
shapes falls back to the existing `where_rows` chain evaluator, so semantics can
never diverge.

## Measured (dgx, 20k graph-bench, polars, warm median)

| query | before | after |
|---|---|---|
| q5 | 10.2 ms | **6.6 ms** |
| q6 | 10.7 ms | **8.9 ms** |
| q7 | 13.5 ms | **10.5 ms** |

Values unchanged (q5=1111 etc.).

## Byte-parity verification

- Differential fast-vs-fallback identical on 8 adversarial cases: nulls, case
  variants, unicode casefold (`Straße`), numeric ranges, multi-expr conjunction.
- Full cypher lowering + row-pipeline suites: **1645 pass** — including the
  residual-specific connected-join tests that caught an earlier (rejected)
  implementation approach.


lmeyerov:
Review responses:
- `_residual_polars_expr` return type is now `Optional['pl.Expr']` (TYPE_CHECKING import), was `Optional[Any]`.
- `expr: str` is the honest input type — the #1729 lowering serializes residuals as canonical predicate *strings* in ASTCall params (a typed term needs a lowering-level refactor, out of scope); rationale now in the docstring.
- CHANGELOG entry added.
- New committed regression suite `test_residual_polars_native.py` (20 tests): positive shapes incl. nulls/casefold/negative literals, negative declines, the all-or-nothing group-fallback gate, pandas-frames-never-fast-lane.

PR #1765, opened 2026-07-21
refactor(gfql): fuse polars seeded RETURN-dst helper into one collect_all (#1755)

lmeyerov:
## What
`_seeded_typed_return_dst_polars` now composes its seed-first reduction as one lazy plan (semi-joins over `drop_nulls()`'d id sets) and collects once via `pl.collect_all([dstn, edges])`, replacing ~8 eager ops (19 internal collects/exec). The empty-seed early return is subsumed (empty flows through to schema-identical empty frames). Value-identical rows; row order may differ (documented contract). Also removes the whole `is_in(...implode())` deprecation surface.

## Honest perf note
Hypothesis was collect-overhead dominance (as in the two-star OLAP lane, where fusion won 30-40%). **Refuted here**: profiling shows the single fused collect takes ~3.0ms itself — this path is compute-dominated (columnar scans over the edge frame), so fusion is perf-neutral (dgx same-session: median 3.38→3.23ms, min ~2.2; within noise). Shipping as a simplification, not a speedup. The real ≤2ms lever for seeded polars is engaging the resident #1658 adjacency index from the cypher fast path — follow-up, tracked in plans/gfql-benchmark-numbers.

## Testing
Full seeded suite (51) green, incl. the polars gate tests (LazyFrame/mixed-engine decline, null-id parity, labels precedence) and the per-engine differential sweep; dgx 4-engine probe: engaged + parity on all engines.

### DIFF
--- a/graphistry/compute/chain_fast_paths.py
+++ b/graphistry/compute/chain_fast_paths.py
@@ -7,7 +7,7 @@ direction); this module imports only leaf modules (no back-edge into chain.py).
 """
 # ruff: noqa: E501
 
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple, cast
 
 from graphistry.Plottable import Plottable
 from .ast import ASTNode, ASTEdge, Direction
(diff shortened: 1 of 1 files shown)


lmeyerov:
Closing as obsolete rather than rebasing — three reasons, two from this PR's own body:

1. **Perf-neutral by its own measurement**: the fusion hypothesis was refuted (compute-dominated path; the single fused collect costs ~3.0ms itself). Pure style refactor.
2. **Architecturally superseded by #1768/#1770 (both shipped in 0.58.0)**: `_seeded_typed_return_dst_polars` on master now interleaves eager resident-index gathers (CSR + node-id positional lookups) with per-stage scan fallbacks — a whole-function lazy `collect_all` plan can't be rebased onto that without undoing the index decoupling or re-interleaving eager gathers into the lazy plan. Complexity with no measured gain.
3. **The deprecation payload is already handled on master**: the `is_in(Series)` deprecation (polars#22149) is mitigated via the documented `.implode()` pattern, cited in the code comments.

If a semi-join/lazy form is wanted later, it should be designed against the post-index helper shape (and only with a measured win). Reopen if that reading is wrong.

PR #1776, opened 2026-07-25
perf(gfql): dispatch indexed fixed-hop bindings before the canonical traversal

lmeyerov:
## What

The resident-index fixed-hop path (`rows(binding_ops=...)` — the Cypher multi-alias `MATCH … RETURN` shape) was only consulted **inside row materialization**, after the engine had already executed the full canonical traversal. It therefore built its compact path bag on top of the graph-sized work it exists to avoid.

Both boundaries now ask the same shared, structural gate **first**:

- pandas / cuDF — `_handle_boundary_calls` (`compute/chain.py`)
- native Polars — `chain_polars` (`gfql/lazy/engine/polars/chain.py`)

Only when the helper can serve the **whole** middle exactly do they skip the canonical traversal and hand the compact state to the unchanged row materializer via an internal graph copy. A decline is memoized so the rows path does not re-attempt (or re-record) the same decision.

Engagement stays **operator / index / dtype / cost** based — no query, schema, or hop-count recognition. Seeded, prefiltered, policy-bearing, shortest-path, unsupported, and cost-gated shapes still fall back to canonical execution with identical results.

Also included:

- **Indexed seed lookup covers `{id, …extra scalar constraints}`** — gather the one indexed row, then filter it, instead of scanning the whole node table. Missing ids and non-matching constraints return the same empty result; duplicate-id graphs cannot build the index and are unaffected.
- **Seeded typed-hop property projection keeps its fast path through trailing `DISTINCT` / `ORDER BY` / `SKIP` / `LIMIT`** by delegating those plain frame ops to the canonical chain.
- **Fix (pre-existing, cuDF)**: that projection applied the *pandas* rows-pivot artifact (int → `float64`, bool → `object`) on every engine, but cuDF's canonical pivot preserves source dtypes — so the fast path returned `float64`/`object` where cuDF's own canonical path returns `int64`/`bool`. The cast rule is now engine-aware; the dtype-class decline guard is unchanged.
- **Empty edge-frame alias marker columns** now follow each engine's own canonical column order (pandas first, polars appended).

## Not in this PR

`Engine.py` AUTO source-native routing is **unchanged** — the deferred `#1743` policy is deliberately out of scope, and the two tests that pinned it were dropped.

## Tests

Standard-derived (LDBC-shaped; benchmark names appear only in test ids/comments), all comparing against the **same-engine** canonical path:

- exact value / order / dtype / index / schema parity, one trace decision per serve/decline, missing / stale / `off` index lifecycle, named-edge schema, policy hooks, shortest-path exclusion, exception propagation;
- **no-canonical-traversal proofs** for both pandas (`ASTEdge.execute` forbidden) and polars (`_chain_traversal_polars` forbidden);
- an **unnamed-middle regression** pinning the bypass gate — reverting the gate fails it on both engines;
- mixed int / float / bool / string / id projection dtype parity across pandas / polars / cuDF;
- `test_out_of_shape_declines_with_parity` reconciled: `DISTINCT` and `ORDER BY … LIMIT` moved to a new engages-with-parity case, since they are now served *with* proven parity.

## Validation (dgx-spark, capped guard, read-only source)

- full CPU `graphistry/tests/compute`: **6232 passed**, 1 failed — master's pre-existing `test_chain_dask_edges`
- cuDF lane (indexed + seeded fast-path files): **30 passed**
- `ruff`: clean · `mypy`: **0 new errors vs master**
- standard LDBC SNB SF1 interactive-short profile: the redundant two-hop typed-mask and 16-merge buckets disappear; profiled call **1263 ms → 106 ms (~11.9×)**, exact 19-row oracle preserved (diagnostic measurement, not a published comparison number)

---

## Review follow-ups (2026-07-25)

Manual review flagged the dynamic typing this PR introduced. Addressed in `83212ab1`:

- The three smuggled `_gfql_indexed_bindings_*` attributes collapse into one typed **`IndexedBindingsHandoff`** dataclass (`gfql/index/handoff.py`) that owns the only attribute access for them and answers `serves(binding_ops, engine)` / `declined(binding_ops)`, instead of every call site re-deriving plan equality with `getattr`/`setattr`. `chain.py`, the polars chain, both row pipelines, and `frame_ops` now make typed calls.
- The review skill (`agents/skills/review/SKILL.md`) now documents the **batched-`.assign()`** rule (copy-then-mutate → one `assign`; `rename` and boolean-masking already copy).

Verified on this branch alone: full CPU `graphistry/tests/compute` **6232 passed** with only master's pre-existing `test_chain_dask_edges`; cuDF lane **60 passed**; ruff clean; mypy unchanged vs master.

The remaining review items land in the stacked #1777, because they touch files that PR also modifies (`index/bindings.py`, `index/lookup.py`, `index/api.py`): the `_concat`/join/degree-helper relocation and DRY, the `.assign()` conversions, the decline-reason fix, and the cross-engine test amplification. Reviewing the stack top-down shows the final state.

Engine-shape divergences found along the way are filed separately as #1779.

### Review follow-ups, round 2

- **`7777e191` declared fields.** The `_gfql_*` execution context is now DECLARED on `Plottable` with defaults on `PlotterBase` (`bind()` is `copy.copy`, so per-graph values still travel), mirrored on `RowPipelineMixin` (its adapter is not a `PlotterBase`) and in the `RowPipelineCtx` protocol. `index/handoff.py` drops to **zero** dynamic attribute calls; the policy/registry accessors, both chain boundaries, both row pipelines, and `frame_ops` stop using `getattr`/`setattr` for this family. Two real bugs fell out, both invisible under `getattr(..., default)`: `_SyntheticRowGraph` silently omitted two fields, and declaring a default **disables** any `getattr(x, name, fallback)` whose fallback differs — `get_registry` fell back to `EMPTY_REGISTRY` while the declared default is `None`. Every other reader was audited; the two that remain use `None`, which matches. Deliberately excluded: `_gfql_native_sp_cache` (different subsystem) and the repo-wide sweep.
- **`3015b736` purely functional boundary.** `set_handoff` mutated a graph in place, safe only via an invariant three branches away, and one decision was spread over three mutable locals. Now a single `_plan_indexed_middle(...)` returns `Optional[IndexedBindingsHandoff]` — `None` = doesn't apply, with state = serves, without state = tried and declined — computed once, with the gate as a named predicate mirroring the polars twin. Both boundaries only attach, never mutate.

Verified on this branch alone after each step: CPU **6232 passed** (only master's pre-existing `test_chain_dask_edges`), cuDF **60 passed**, ruff clean, mypy unchanged vs master.

### Review follow-ups, round 3

- **`54d6e77f`** — `row/frame_ops.py`'s context protocol was declared `Any`; it now types frames as `Optional[DataFrameT]`, ids as `Optional[str]`, the base graph and the adapter `_g` back-reference as `Optional[Plottable]`, and edge aliases as `Optional[Iterable[str]]`. That file now has **zero** `getattr` (including two pre-existing ones, pointless once the fields are declared). Three polars-on-`DataFrameT` calls the sharper types expose take localized ignores — the documented convention for engine-polymorphic frames.
- **`54d6e77f`** — `polars/row_pipeline.py` swallowed unexpected exceptions: extracting the finisher made the **indexed** path share the generic path's `except SchemaError: return None`. That rationale is the generic builder's (polars won't unify int/float join keys the way pandas implicitly does), while the indexed state comes from a helper that already verified those dtypes — so a `SchemaError` there is a bug being turned into a silent slow-path fallback. The finisher now takes `decline_on_schema_error`; the indexed caller passes `False`.
- **`6c41ee8e`** — completed the field conversion: four remaining non-test `getattr`/`setattr` sites converted, and **both** hand-rolled Plottable stand-ins were missing `_gfql_index_registry` (reachable, since the row pipeline's `base_graph` can be the adapter).

On `Any` in `polars/row_pipeline.py`: `ops` is now `Sequence[ASTObject]` with real `isinstance` narrowing, but `state`/`alias_frames` stay `Any` deliberately — the same parameter receives a polars eager frame, a polars `LazyFrame`, and the engine-polymorphic `DataFrameT` the indexed helper returns. A precise union there produced exactly 6 mypy errors, i.e. it buys only suppressions; the comment records that.

Verified on this branch after each step: CPU **6232 passed** (only master's pre-existing `test_chain_dask_edges`), cuDF **60 passed**, ruff clean, mypy unchanged vs master, no unnecessary suppressions.


graphistry/compute/gfql/index/bindings.py
```diff
@@ -0,0 +1,722 @@
...
+        return frame
+    if engine == Engine.POLARS:
+        # Native Polars intentionally omits pandas' alias-marker residue.
+        return frame
+    out = frame.copy()
+    out[name] = True
+    return cast(DataFrameT, out)
+
+
+def _concat(frames: Sequence[DataFrameT], engine: Engine) -> DataFrameT:
```
lmeyerov, line 145: DRY or at least put in proprer files

graphistry/compute/gfql/index/bindings.py
```diff
@@ -0,0 +1,722 @@
...
+    if engine == Engine.POLARS:
+        import numpy as np
+        import polars as pl
+
+        return cast(
+            DataFrameT,
+            frame.with_columns(pl.Series(_EDGE_ORD, np.asarray(positions))),  # type: ignore[operator]
+        )
+    out = frame.copy()
+    out[_EDGE_ORD] = positions
```
lmeyerov, line 171: shouldn't all these `.copy() -> mutate` be `.assign()` ?

graphistry/compute/gfql/index/bindings.py
```diff
@@ -0,0 +1,722 @@
...
+        reverse = one(dst, src, 1)
+        if engine == Engine.POLARS:
+            reverse = reverse.select(forward.columns)  # type: ignore[operator]
+        return _concat([forward, reverse], engine)
+    if direction == "reverse":
+        return one(dst, src, 0)
+    return one(src, dst, 0)
+
+
+def _estimate_join_rows(
```
lmeyerov, line 234: ins't there a file just for join/merges?

graphistry/compute/gfql/index/handoff.py
```diff
@@ -0,0 +1,72 @@
...
+    out._gfql_indexed_bindings_handoff = handoff
+    return out
+
+
+def set_handoff(g: Any, handoff: IndexedBindingsHandoff) -> None:
+    """Attach ``handoff`` to an object the caller is itself constructing."""
+    g._gfql_indexed_bindings_handoff = handoff
+
+
+def read_handoff(g: Any) -> Optional[IndexedBindingsHandoff]:
```
lmeyerov, line 65: Any

graphistry/compute/gfql/lazy/engine/polars/row_pipeline.py
```diff
@@ -636,6 +636,74 @@ def _rewrap(g: Plottable, table_df: Any) -> Plottable:
     return frame_ops.row_table(_RowPipelineAdapter(g), table_df)
 
 
+def _finish_binding_rows_polars(
+    g: Plottable,
+    ops: Sequence[Any],
+    state: Any,
+    alias_frames: Dict[str, Any],
+    node_id: str,
```
lmeyerov, line 650: A lot of Any

graphistry/compute/gfql/row/frame_ops.py
```diff
@@ -22,6 +22,9 @@ class RowPipelineCtx(Protocol):
         _nodes: Any
         _edges: Any
         _edge: Any
+        _gfql_rows_base_graph: Optional["Plottable"]
+        _gfql_start_nodes: Any
+        _gfql_rows_edge_aliases: Any
```
lmeyerov: More Any to refine

graphistry/compute/gfql/row/frame_ops.py
```diff
@@ -65,22 +72,48 @@ def row_table(ctx: RowPipelineCtx, table_df: Any) -> "Plottable":
...
     out._source = None
     out._destination = None
     out._edge = ctx._edge if ctx._edge is not None and ctx._edge in table_df.columns else None
     if out._node is not None and out._node not in table_df.columns:
         out._node = None
-    base_graph = getattr(ctx, "_gfql_rows_base_graph", None)
+    base_graph = ctx._gfql_rows_base_graph
     if base_graph is None:
-        base_graph = getattr(ctx, "_g", None)
+        base_graph = getattr(ctx, "_g", None)  # adapter-only back-reference
```
lmeyerov: Fix all these getattr

Issue #1709, opened 2026-07-07
GFQL/polars: native rows(binding_ops) — traversal queries NIE (blocks polars for edge/multi-hop Cypher)


lmeyerov:
## Summary (native-polars feature; effort L, multi-day)
On **polars**, any Cypher query that crosses an edge (a traversal) NIEs: `polars row pipeline does not yet support op 'rows'` when `rows` carries `binding_ops`. This blocks graph-benchmark **q1, q2, q8, q9** (and every multi-hop/edge query) on polars/polars-gpu. Single-MATCH-node queries already run natively on polars.

## Enforcement sites
- `graphistry/compute/gfql/lazy/engine/polars/chain.py:301-305` (`_run_calls_polars`) and `253-257` (`_call_native_on_polars` force-declines `rows` when it carries `binding_ops`/`alias_endpoints`).
- `graphistry/compute/gfql/row/pipeline.py:4712` whitelist `_POLARS_NATIVE_ROW_PIPELINE_CALLS = {"rows","skip","limit","distinct","drop_cols"}`; force-decline at `4735-4744`.

## Why it's pandas/cuDF-bound
The traversal itself already runs on polars (each `ASTNode/ASTEdge.execute(engine=POLARS)` uses the polars hop engine). What is pandas/cuDF-API-bound is the **connected-binding state-threading + materialization** (the joins/merges):
- `_gfql_connected_bindings_state` (`pipeline.py:3347`), `_gfql_connected_bindings_row_frame_from_state` (`3576`)
- `_gfql_cartesian_node_bindings_row_table` (`3788`), `_gfql_multihop_binding_rows` (`3251`, ~93 lines)
- `_gfql_disambiguate_has_edge_destination_nodes` (`3232`) + `_gfql_has_edge_destination_label_col` (`3221`)
- `_gfql_shortest_path_scalar_bindings_row_table` (`3721`) + `_gfql_shortest_path_scalar_native` (`3630`)
- `_gfql_bindings_row_table` alias_endpoints variant (`3849`); `EdgeSemantics.orient_edges` (`same_path/edge_semantics.py:36`)

Idiom counts in the 3180–3860 range: `.merge`×7, `.iloc`×8, `.drop(columns=)`×11, `.rename(columns=)`×7, `.drop_duplicates`×5, `.isin`×3, `.assign`×4, `.duplicated`×2, `.groupby().min().reset_index()`, null-handling `.astype(object).where(...)`.

[...]

Issue #1780, opened 2026-07-26
perf(gfql): secondary node property index is consulted only at the connected-bindings seam — single-alias seeded lookups still full-scan


lmeyerov:
## Summary

The secondary node property index landed in #1777 works, but it is reachable from **exactly one** query shape. Every other seeded shape builds the index, keeps it resident and valid, and then never consults it — falling back to a full scan of the node frame.

Measured on official LDBC SNB SF1 (pandas, master `84be35fb`, 4 runs/arm, position-balanced ABBA+BAAB), property index ON vs OFF:

| query | OFF ms | ON ms | speedup | verdict |
|---|---|---|---|---|
| IS1 seed-lookup | 725.6 | 759.6 | 0.96x | built, never engaged |
| IS2 expand-order-limit | 826.0 | 842.1 | 0.98x | built, never engaged |
| IS4 message-content | 227.9 | 231.8 | 0.98x | built, never engaged |
| IS5 message-creator | 549.3 | 580.1 | 0.95x | built, never engaged |
| IS6 message-forum | 4286.6 | 4633.1 | 0.93x | built, never engaged |
| **IS7 message-replies** | **71.7** | **19.5** | **3.68x** | **engaged** |

Non-IS7 deltas (0.5-7%) sit inside per-arm run-to-run spread (1.4-16.4%), i.e. no effect. The IS7 arms do not overlap (ON 18.6-20.0, OFF 69.8-73.0).

## Root cause

The property index is consulted at one seam only: `connected_bindings` — multi-alias `rows()` over a connected fixed-hop path (`_seed_rows_via_property_index`). IS7 is the only interactive-short query with that shape.

- IS1 / IS2 / IS4 / IS5 use single-alias `rows(source=...)` or a node-only lookup and never reach the seam at all (instrumented `node_prop_seam_probes: 0`).
- IS6 reaches the seam and is declined `unsupported_shape`.

The single-alias seeded path (`chain_fast_paths._resident_seed_indexes`, feeding the `destination_return` seam) requires the seed predicate to be on the **node-id binding**. When the seed is on any other property it full-scans the node frame, without ever asking whether a resident property index covers that column.

That is the same degeneracy #1777 was filed to fix — SNB binds nodes on `__pgbench_node_key__` while every seed predicate is on the `id` property — but it was only fixed for the connected-bindings shape.

## Isolated reproduction

Same one-hop query at SF1 node counts, differing only in binding shape:

[...]

PR #1781, opened 2026-07-26
feat(gfql): native polars unbounded directed var-length rows(binding_ops) — unblocks LDBC IS6 on polars (#1709)

lmeyerov:
Closes part of #1709.

## The gap

`rows(binding_ops=...)` — the Cypher multi-alias bindings table — already lowered natively on polars for fixed-length hops, bounded directed `-[*i..k]->`, bounded undirected `-[*1..k]-`, and node-cartesian patterns. An **unbounded** variable-length segment still declined:

```
NotImplementedError: polars engine does not yet natively support cypher row op 'rows';
use engine='pandas' for this query (no pandas fallback; parity-or-error by design)
```

That was the last shape blocking `engine='polars'` on **LDBC SNB interactive-short-6** (`MATCH (m:Message)-[:REPLY_OF*0..]->(p:Post)<-[:CONTAINER_OF]-(f:Forum)-[:HAS_MODERATOR]->(mod)`) — the one interactive-short query polars could not answer at all.

## Now native

**Unbounded DIRECTED fixed point**: `-[*]->`, `-[*0..]->`, `<-[*]-`, `-[:TYPE*]->`, with or without a fixed-hop tail, seeded or unseeded.

Pandas finds the traversal depth by expanding the **path** frontier until it is empty, which materializes every partial path at every hop. This lowering splits that in two:

1. a dedup-by-node frontier walk that computes the exhaustion depth `D` (each hop is O(N), not O(paths));
2. the **same lazy bounded pair-join loop** the `-[*1..k]->` arm runs, with `max_hops = D`.

Step 1 is exact, not an approximation: the path frontier at hop `h` is non-empty iff *some* walk of length `h` leaves a seed, and deduping by endpoint node changes no walk's **existence** — only its multiplicity, which step 2 then reproduces in full (`pairs` is never deduped, so parallel edges multiply per hop exactly as the pandas merge does). The emitted rows are pandas-identical while the exponential path expansion happens once, lazily. No pandas bridge, no `to_pandas()` round trip.

**Non-termination**: a cycle reachable from a seed means infinitely many paths. Pandas raises E108 *"Cypher multi-alias row bindings currently require terminating variable-length segments"*; so do we, via the same shared helper, with the same message. Detection is by pigeonhole on the distinct nodes touched by the matched edges (a walk of `N` edges visits `N+1` nodes, so a non-empty frontier at hop `N` **is** a reachable cycle) — i.e. before the blow-up rather than after it, which is the one place we deliberately beat pandas' cost while landing on the identical outcome.

The bounded `-[*1..k]->` arm and the new arm now literally share one loop (`_directed_varlen_reachable_polars`) rather than two copies kept in sync by a comment.

## Also fixed: a latent silent-wrong on `-[*0..k]->`

Found by the differential fuzz while adding the above, **pre-existing on master**:

The polars bindings builder rebuilt its table from the **chain output**, while the pandas oracle rebuilds from the **pre-chain base graph**. The traversal prunes to what *it* considers matched, and that is not the bindings builder's match set — a zero-hop variable-length segment binds a seed that has no matching outgoing edge, and the traversal drops that node entirely. So `MATCH (a)-[*0..2]->(b)` could silently return **fewer rows than pandas**: no error, just missing rows.

The builder now sources its node/edge tables from the same base graph pandas uses — which the *indexed* builder in the same function was already being handed. Perf impact measured below: none.

## Still declines (honest NIE, never a silent answer)

| shape | why |
|---|---|
| `-[*]-` (undirected unbounded) | needs both the `min_hops == 1` multiplicity reconstruction *and* backtrack-aware termination; pandas rejects it outright (`GFQLTypeError`) |
| `-[r*..]->` (aliased var-length) | pandas rejects variable-length relationship aliases outright |
| unbounded **without** `to_fixed_point` (`min_hops=2`, no max) | pandas silently truncates at `len(step_pairs) + 1` iterations rather than erroring, and its `step_pairs` row count is not reconstructible here — so the truncation depth, hence the answer, is not reproducible |
| undirected var-length with `min_hops != 1` | unchanged from master |

Separately unchanged: `min_hops > 1` **with** `to_fixed_point` still declines one level up, in `_chain_traversal_polars` (`_is_native_multihop`) — a polars *hop* gap, not a `rows` gap, and out of scope here.

AUTO engine routing is untouched (#1743 stays deferred).

## Verification

DGX Spark, RAPIDS image `sha256:74544b1b…`, candidate `7efc2083` vs master `84be35fb` run identically.

**Correctness**

| lane | master `84be35fb` | this branch |
|---|---|---|
| `tests/compute` CPU | 1 failed, 6260 passed | **1 failed, 6278 passed** |
| CPU failure | `test_engine_coercion.py::TestChainCoercion::test_chain_dask_edges` | same one, unchanged |
| `tests/compute -k cudf` | 8 failed, 468 passed | **8 failed, 468 passed** — identical set (pre-existing igraph-round-trip + cudf-coercion) |
| ruff | — | **All checks passed** |
| mypy (`--config-file mypy.ini graphistry`) | 206 errors | **166 errors** |

mypy delta computed by bucketing both runs on `(file, error-code)` and diffing (a raw count is not a delta): **no bucket is worse**; the reductions are `polars/row_pipeline.py` arg-type 4→0 and assignment 2→0, `polars/chain.py` arg-type 1→0, and `gfql_fast_paths.py` arg-type 47→32 / assignment 13→3 / attr-defined 5→0 / dict-item 2→0 / union-attr 1→0. The reduction comes from typing `filter_by_dict_polars` with a constrained `DataFrame | LazyFrame` TypeVar — it was declared eager and called with both flavours, which was poisoning inference downstream.

**Differential fuzz vs the pandas oracle** — 10 edge shapes (`*0..`/`*1..`/`*2..` fixed point, forward + reverse, typed + untyped, `to_fixed_point` with and without `max_hops`, bounded controls) × 2 pattern tails × random graphs (DAGs and cyclic, with self-loops, parallel and antiparallel edges, integer and string node ids), comparing the projected result table:

```
seed 1  (int ids):  cases=800   mismatch=0
seed 7  (int ids):  cases=2400  mismatch=0
seed 23 (int ids):  cases=1200  mismatch=0
seed 31 (str ids):  cases=1200  mismatch=0
```

Before the base-graph fix the same harness reported **171/800 mismatches**, all of them the dropped zero-hop rows.

**Perf** (dgx-spark, exclusive perf lock held for one run at a time, synthetic 440k-node / 420k-edge reply-forest + forum/moderator graph, warm median of 7, all rows verified non-empty):

| query | master | this branch |
|---|---|---|
| IS6-shaped `REPLY_OF*0..` walk, **polars** | `NotImplementedError` | **182.4 ms**, 1 row |
| IS6-shaped `REPLY_OF*0..` walk, pandas | 229.4 ms, 1 row | 220.7 ms, 1 row |
| q1-shaped 1-hop grouped count, polars | 116.9 ms | 113.6 ms |
| q8-shaped 2-hop count, polars | 63.6 ms | 64.2 ms |
| 1-hop filtered projection, polars | 171.0 ms | 177.8 ms |

The last three are the shapes the base-graph rebuild could have regressed; they land within run-to-run noise. This is a **synthetic** IS6-shaped probe, not an LDBC SF1 measurement — the SF1 harness number should be re-taken by the benchmark lane now that the query runs.

## Tests

- `test_engine_polars_binding_rows.py`: 12 new parity queries in `SUPPORTED` (unbounded forward/reverse/typed/filtered/with-tail, plus the `-[*0..2]->` zero-hop cases that were silently wrong); `-[*]-` moved into `DEFERRED`; 7 new focused tests — the IS6 pattern end-to-end at both a multi-hop seed and a **zero-hop** seed, the zero-hop-seed-with-no-outgoing-edge regression pin, path multiplicity over a diamond with parallel edges, cycle and self-loop both raising the E108 diagnosis on both engines, the three decline shapes, and the depth-0 boundary.
- `test_engine_polars_row_pipeline.py`: the unbounded-defers test now asserts the directed case is served and the undirected/aliased cases still raise.

---


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
...]-(f:Forum)-[:HAS_MODERATOR]->(mod)`), the one interactive-short query polars could not answer. It now runs natively: a dedup-by-node frontier walk finds the exhaustion depth, then the SAME lazy bounded pair-join loop the `-[*1..k]->` arm uses materializes one row per distinct edge SEQUENCE (Cypher path multiplicity, parallel edges included) — no pandas bridge, no `to_pandas()` round trip. A cycle reachable from the seed means infinitely many paths; that raises the same E108 "require terminating variable-length segments" error pandas raises, and is detected before the path expansion blows up rather than after. Still declining honestly (NIE, never a silent answer): UNDIRECTED unbounded (`-[*]-`, needs the min_hops == 1 multiplicity reconstruction plus backtrack-aware termination — pandas rejects it outright), aliased variable-length relationships (pandas rejects those too), and unbounded segments WITHOUT `to_fixed_point` (pandas silently truncates at a bound this lowering cannot reconstruct). Cross-engine parity is the gate: differential fuzz vs the pandas oracle over random DAGs and cyclic graphs with self-loops and parallel edges, plus pinned IS6/zero-hop/multiplicity/cycle tests.
```
lmeyerov: need pyg-bench perf expectation too now that unblocks some task(s)

once happy with postive/negative tests on either side of opt boundary, and ci still green, can land both sides

lmeyerov:
Assessed against the landing bar. **`[2]` and `[3]` are met; `[1]` is the only gap.**

- **`[2]` pos AND neg on either side of the boundary — YES.** For this PR the boundary is
  *native polars lowering* vs *the parity-or-error decline*, and both sides are covered:
  **13** `NotImplementedError` decline tests and **56** positive tests across
  `test_engine_polars_binding_rows.py` + `test_engine_polars_row_pipeline.py`. 260 pass locally.
- **`[3]` CI — green** (16 pass, 0 fail).
- **`[1]` pyg-bench — missing.** Since this PR's value is a *cell* (IS6 goes from
  un-answerable to answerable on polars) rather than a speedup, the expectation should be a
  **capability assertion** — IS6 must RUN on the polars lane and return the correct rows —
  not a latency threshold. That is what I'll add.

## Two things I found on the branch

It was carrying **three commits that were never pushed** (origin was at `7efc2083`, and since
origin was an ancestor they were additive and invisible on the PR):

**Kept — `1ab30dbc` "move the var-length row specializations out of row_pipeline.py".** This
is exactly the extraction the sequencing note calls for, so that later `row_pipeline.py` PRs
don't pay the conflict repeatedly. I verified it is **independent** of the other two (it
cherry-picks cleanly onto origin's head), lint clean, 260 tests pass — and pushed it here.
New module `lazy/engine/polars/varlen_rows.py` (+136 / −114). It deliberately keeps a
function-scope import: hoisting to module scope reintroduces a `row_pipeline → lazy engine →`
cycle, and the commit message says so.

**Moved off this branch — `bd559eae` + `18e4c953`, "seed the … fast paths from the resident
property index".** These are the **#1780 lever**, which is on our DEAD list at **1.00×**. They
do not belong on a var-length-rows PR. They are *not* the same implementation as the rejected
`perf/gfql-1780-node-seed-prop-index` (which wired `maybe_index_node_seed` into
`ASTNode.execute`) — these target the `_resident_seed_indexes` call sites instead — so they
may well behave differently. But #1780's entire history is a lever that engaged correctly and
bought nothing, so they need measuring, not assuming.

They are parked on **`perf/gfql-1780-propindex-seed-seams`** (pushed, no PR). Nothing is lost;
they just shouldn't ride along unmeasured on a PR about something else.

lmeyerov:
`[1]` is now addressed for this PR — as a **capability** expectation rather than a perf lane.

pyg-bench **#103** (merged): `configs/capability_expectations.yaml` +
`scripts/check_capability_expectations.py`. A latency lane is the wrong instrument here —
this PR's value is a *cell*, not a speedup, so there is no "before" number to ratio against
and a wall-clock threshold on a newly-working query is pure flake. The expectation instead
asserts IS6 **ran** (`status == "ok"`) and returned the **same rows as the pandas oracle on
the same lane**, so it cannot drift with the dataset.

### The baseline that makes this a clean pin

Measured 2026-07-27 at SF1 on **both** master `84be35fb` **and** the #1782–#1785 stack:

| engine | IS6 (`message-forum`) |
| --- | --- |
| polars | `status=failed`, `classification=not_yet_assessed`, `rows_returned=None` |
| pandas | `status=ok`, `rows_returned=1` |

Nothing already in the stack moves it — so this is your cell specifically, and the gate
currently **exits 1**. It exits 0 against a run where the query answers (verified both ways).

A missing probe result is a **failure**, not a skip: a query quietly dropped from a suite
would otherwise read as "expectation met", which is exactly how six of nine graph-bench cells
came to be scored without ever calling the query engine.

### What remains on this PR

Run the SNB polars lane against this branch and confirm the gate flips to exit 0. `[2]` is
already met (13 decline tests + 56 positive) and `[3]` is green.

Issue #101, opened 2018-12-05
for neo4j integration please also add node labels and relationship-type to graphistry data


jexp:
currently it only copies id and properties:

```
def bolt_graph_to_edges_dataframe(graph):
    import pandas as pd
    return pd.DataFrame([
        util.merge_two_dicts(
            { key: value for (key, value) in relationship.items() },
            {
                relationship_id_key:    relationship.id,
                start_node_id_key:          relationship.start_node.id,
                end_node_id_key:     relationship.end_node.id
            }
        )
        for relationship in graph.relationships
    ])


def bolt_graph_to_nodes_dataframe(graph):
    import pandas as pd
    return pd.DataFrame([
        util.merge_two_dicts(
            { key: value for (key, value) in node.items() },
            {
                node_id_key: node.id
            }
        )
        for node in graph.nodes
])
```

lmeyerov:
Agreed, both for `pygraphistry` and the direct connector. The type mapping fits naturally, such as with the legend. 

It's a little unclear how we should do labels for the table -- maybe column `label1=True`, and once we support array-types, `label=[label1, ...]`.

jexp:
Sounds good to me

jexp:
Sounds good to me

cwharris:
Deleted a duplicate comment. Didn't realize it would leave a note in it's place, but it makes sense! Backfire! :)

lmeyerov:
![image](https://user-images.githubusercontent.com/4249447/50053153-a5759200-00e4-11e9-8545-2e132b40c782.png)

Looks like the enterprise connector does the right thing out-of-the-box, we just need pygraphistry to do the same. Agreed, let's do it on the arrow branch for jan.

lmeyerov:
Already been in the enterprise studio, and now in notebooks as well with https://github.com/graphistry/pygraphistry/releases/tag/0.9.63 . Win!

PR #1782, opened 2026-07-26
perf(gfql): evaluate indexed edge_match on candidate rows, not the whole edge frame

lmeyerov:
# perf(gfql): evaluate indexed `edge_match` on candidate rows, not the whole edge frame

## The bug

`index_seeded_hop` built the simple-equality `edge_match` mask over **all E edges**:

```python
edge_keep = _build_edge_keep_mask(edges, edge_match, engine, xp)   # length E
...
rows = rows[edge_keep[rows]]                                        # read at ~degree positions
```

The mask is only ever read at `rows` — the handful of positions the CSR adjacency
lookup already returned. So a full-column predicate scan sat **inside** an O(degree)
traversal, and the indexed path scaled with the graph instead of with the answer.

This is the mechanism behind a long-standing puzzle in our LDBC work: the native
`g.hop()` is flat in graph size while the indexed *surface* query is not. It is the
same code either way — the difference is that the surface query carries a typed edge.

## Evidence it is the dominant cost

Profiling a seeded typed-edge query on a 3.18M-node / 14M-edge polars graph
(LDBC SNB SF1-shaped), master `84be35fb`:

```
comp_method_OBJECT_ARRAY   248 ms/call, 1 call/query   <-- 81% of a 308 ms query
```

Caller chain (spy on the pandas comparison op):

```
chain.py:_handle_boundary_calls -> _chain_impl -> ast.py:execute -> ComputeMixin.hop
  -> index/api.py:maybe_index_hop -> index/traverse.py:index_seeded_hop
  -> traverse.py:127 _build_edge_keep_mask   (series == val) over 14,000,000 rows
```

## The fix

`_build_edge_row_filter` does the schema-level validation up front — including the
dtype-mismatch decline that keeps error parity with the scan, which is O(1) and
unchanged — and `_EdgeMatchRowFilter.mask_for(rows)` applies `col == val` to the
gathered candidate rows only, using each frame's native `==` exactly as before (so
cuDF string columns stay on the cuDF layer).

Each edge row is returned by a given index at most once (frontiers are
set-differenced against `visited`), so a seeded hop examines O(edges traversed)
elements. The worst case is a fixed-point *undirected* walk reaching the whole graph,
where the out- and in-indices are filtered separately and the total approaches 2E
against the eager form's E — a constant factor on a query already O(E) in its
traversal alone.

A failure mid-traversal abandons the indexed path entirely (`return None`), so the
scan fallback stays parity-safe; nothing observable has been mutated at that point.

## Correctness

- **650 differential comparisons**, 0 divergences: pandas + polars × 13 traversal
  shapes × 25 random seeds, on a dense graph with **null-carrying** string and
  numeric columns. Compares columns, dtypes, row count and full content against the
  same query with the index disabled.
- Shapes covered: forward / reverse / undirected typed, 2-hop, 3-hop, fixed-point,
  two-column `edge_match`, numeric-only, untyped, `+rows(source=)`, and both DECLINE
  paths (membership `edge_match` stays on the scan; dtype mismatch still raises the
  same `GFQLSchemaError`).
- `graphistry/tests/compute/gfql/index`: **identical failure sets** before and after
  (38 failed / 169 passed on both, diff-clean — all pre-existing GPU-image failures
  on this host).
- **cuDF is not yet verified** — the 25.02 image hits the known GB10 driver fault.
  Flagged, not claimed.

## Tests

Two new regression tests (× 2 CPU engines) pin the **shape**, not a wall-clock number,
so they cannot go flaky on a loaded host:

- `test_typed_edge_predicate_only_reads_candidate_rows` — the predicate must see far
  fewer rows than the edge count.
- `test_typed_edge_predicate_cost_flat_in_graph_size` — growing the graph 8× at fixed
  degree must not grow the rows the predicate examines.

## Not a benchmark hack

Engagement keys on a structural precondition — an index-seeded hop carrying a
simple-equality edge predicate — and helps any typed traversal on any schema, engine,
or label. No query, constant, or dataset is recognized anywhere in the change.

## Measured on the real LDBC SNB SF1 lane

Position-balanced ABBA/BAAB (4 samples per build, each build in each slot position
exactly once — dgx-spark has a ~5% slot effect that makes naive A,B,A,B
uninterpretable), quiet host, host-wide perf lock held for the whole experiment,
`rows_returned` validated identical on every query. Pooled mean of per-slot medians:

| query | master `84be35fb` | this branch | delta |
| --- | --- | --- | --- |
| IS1 seed-lookup | 109.21 ms | **77.41 ms** | **−29.1%** |
| IS5 message-creator | 97.05 ms | **72.38 ms** | **−25.4%** |
| IS2 expand-order-limit | 114.74 ms | **92.89 ms** | **−19.0%** |
| IS4 message-content | 3.85 ms | 3.44 ms | −10.5% |
| IS7 message-replies | 13.46 ms | 13.15 ms | −2.3% |

**4 faster, 0 slower**, every result value-identical.

(IS6 fails and IS3 is a primitive gap on this lane on **both** builds — pre-existing,
not evidence either way.)

### What this does NOT do

It does not flip any competitor result. IS1 was ~77× behind Kuzu and is now ~55×.
The remaining dominant cost on IS1/IS5 is a different one, and the run artifacts
name it: `index_engagement` reports `node_prop: built_seam_never_reached`, i.e.
`MATCH (n:Person {id: $personId})` still full-scans 3.18M nodes because the property
index only reaches the connected-bindings seam (#1780). That is the next fix, not
this one.


lmeyerov:
## Correction to the mechanism claim in the description

Self-review found a method error in my own profiling. Flagging it rather than leaving the description overstated.

**What still holds — the load-bearing evidence is unaffected.** The SF1 numbers in the description came from the real `ab-is1-is7-sf1-polars` lane, master vs this branch, position-balanced ABBA/BAAB on a quiet host with rows validated identical: IS1 −29.1%, IS5 −25.4%, IS2 −19.0%, IS4 −10.5%, IS7 −2.3%; 4 faster, 0 slower. That is a genuine measured effect of this diff. The 650-comparison differential and the test-suite parity also stand.

**What does not hold — the stated mechanism.** The profile I cited (`comp_method_OBJECT_ARRAY`, 248 ms of a 308 ms query) was produced by a synthetic probe that called `g.gfql(...)` **without an engine argument**. AUTO resolved to pandas and converted the polars frames, so that ladder — and the "309.92 → 57.44 ms (5.40×)" figure — are **pandas-path** measurements. The lane runs `execution_engine: polars`, which takes a different route entirely (0 merges, native lazy path).

So: the change is a genuine O(E)→O(candidates) improvement, and it demonstrably helps the polars lane by ~29% on IS1. But I have **not** demonstrated that the object-array compare is the mechanism *on the polars path*, and the description reads as though I had. Treat the 5.40× as pandas-path only.

**Unrelated finding worth its own issue (#1767 territory):** on the same graph, `engine='polars'` measured **813 ms vs 17.6 ms for pandas** — 15–19× slower — because the indexes were built via `create_index(g, kind=...)` with no engine, i.e. pandas-engine indexes that don't validate on the polars path, so every seam silently declines to a full scan. That's a plausible user configuration with a large cliff and no diagnostic.

I'd rather this PR be judged on the lane A/B than on a mechanism story I got to via a misconfigured probe.

lmeyerov:
## Review response — one confirmed regression fixed, two false claims corrected

Ran `.agents/skills/review/SKILL.md` against this PR. It found a **real regression that this
PR's own CHANGELOG claimed was impossible**, and I've fixed it rather than documented it.

### The defect

I wrote: *"This never examines more elements in total than the eager mask did — it is strictly
less work, not a tradeoff."* That is false, and the class's own docstring already contradicted
it. Candidate-row evaluation beats one whole-column compare only while candidates stay a small
fraction of the frame. A fixed-point walk that reaches most of the graph inverts it — out- and
in-indices are filtered separately, so it gathers up to 2E elements by *random access* against
the eager form's single sequential pass over E.

Measured, pandas, `to_fixed_point=True, direction="undirected"` with `edge_match`:

| shape | master | this PR (before the fix) |
| --- | --- | --- |
| int etype, E=100k | 22.0 ms | 26.5 ms (**+20%**) |
| str etype, E=400k | 108 ms | 169 ms (**1.56× slower**) |

Instrumented: **1.94×E** elements gathered. And it is publicly reachable, not a synthetic
`g.hop()` shape — `_hop_is_index_coverable` allows `to_fixed_point` with `edge_match`, so
Cypher `-[:KNOWS*]->` lands here (`gfql_explain` confirms `used_index: True`). This is exactly
the failure the anti-hack rule exists to catch: we benchmarked the 1-hop seeded shape and
neither tested nor measured the one where the cost model inverts.

### The fix (`fd0d91ce`)

A cumulative-cost guard: the hop counts gathered rows and, once the **next** batch would push
the total past E/8, builds the whole-column mask once and reuses it for the rest of the walk.
The check happens *before* gathering — a post-hoc check overshoots by up to a whole batch,
which the first version of the new test caught (9184 gathered vs a 6024 limit). Total predicate
work is bounded at ~1.125×E; a seeded hop gathers ~degree and never approaches the threshold,
so it never pays for the mask. `full_mask()` reuses `mask_for`'s null-fill rule verbatim, which
is what makes the two forms agree cell-for-cell across the switch.

After the guard, vs master `84be35fb`, **edge counts identical**:

| shape | master | guarded | was |
| --- | --- | --- | --- |
| fixedpoint int, E=100k | 13.12 ms | 14.11 ms (+7.5%) | +20% |
| fixedpoint str, E=400k | 77.47 ms | 80.97 ms (+4.5%) | 1.56× |

The residual few percent is the designed cost of gathering up to E/8 before switching.

### Also corrected

- The docstring claimed each row is returned at most once "because frontiers are
  set-differenced against `visited`". `edge_match` is only reachable with
  `return_as_wave_front=True`, which **skips** the first-hop `visited` seeding — so seed ids
  can re-enter a later frontier and be gathered twice.
- **Error parity is narrower than I claimed.** Only the O(1) dtype gates are parity-exact. A
  *data-dependent* comparison failure (e.g. a list-valued cell in an object column outside the
  seed's neighborhood) is now observed only if it lands in the gathered candidates, so the
  indexed path can succeed where the scan raises — reproduced. Now stated in the CHANGELOG.
  The converse is safe: forcing a mid-traversal failure abandons the indexed path and returns
  byte-identical results to the scan.

### Tests added

- The mid-traversal abandonment branch — the one new branch, which carried all the parity risk
  and was previously untested and `# pragma: no cover` — now has a negative test asserting the
  fallback reproduces the scan result exactly.
- A fixed-point test pinning that total gathered stays within the guard's bound while the
  answer is unchanged.

### Typing (`51219e28`)

Typed slots; `Any` → `ScalarMatchValue` (the existing `index/types.py` alias, genuinely checked
via the `is_simple_equality_edge_match` TypeGuard); every `cast(ArrayLike/Any, …)` call-site
cast removed. That required one real fix: the `ArrayLike` protocol had no bitwise surface, so
`mask & col_mask` needed the cast sandwich — `__and__`/`__rand__` are now declared next to the
existing `__invert__`/`__add__` pairs. Verified load-bearing (stash it and mypy errors).
`mypy`: 323 files clean. Full compute suite A/B: **identical failure sets**.

### Still open

- **DRY**: `index/bindings.py` already has a canonical candidate-row filter that delegates to
  `filter_by_dict` (parity-exact by construction); `traverse.py` still hand-rolls a second one.
  Not addressed here.
- **cuDF remains unverified** — no GPU on this box, and the GB10 25.02 image faults. The tests
  that would close it already exist and just need a GPU:
  `test_chain_index_parity_vs_scan[chain0..3-cudf]` / `[…-polars-gpu]`. A third, GPU-specific
  concern is unmeasured: per-hop kernel-launch cost points the *opposite* way from the CPU win.
- pyg-bench regression lane for this fix — in progress, not yet run.

graphistry/compute/gfql/index/traverse.py
```diff
@@ -14,20 +14,32 @@
...
+
+# Cost guard for candidate-row edge_match evaluation. Gathering candidate rows beats one
+# whole-column compare only while the candidates stay a small fraction of the frame; a
+# fixed-point walk that reaches most of the graph inverts that and gathers up to 2E by
+# random access. Once cumulative gathered rows reach E/DIVISOR we build the whole-column
+# mask once and reuse it, bounding total predicate work at ~(1 + 1/DIVISOR)*E. The FLOOR
+# keeps small frames on the candidate-row path, where the whole-column compare is cheap
+# anyway and the switch would only add a branch.
+_EAGER_MASK_SWITCH_DIVISOR = 8
+_EAGER_MASK_SWITCH_FLOOR = 1024
```
lmeyerov, line 42: should these be extern'd?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
...M-edge polars graph (LDBC SNB SF1-shaped), a seeded typed-edge query with alias markers goes **309.9 → 57.4 ms (5.4×)**, value-identical, with the full-column compare (previously 248 ms, 81% of the query) gone from the profile. Parity held across 650 differential comparisons — pandas + polars × 13 traversal shapes × 25 random seeds on a dense graph with null-carrying string and numeric columns — comparing columns, dtypes, row counts and full content, including the decline paths (membership `edge_match` still routes to the scan; a dtype mismatch still raises the same `GFQLSchemaError`). One honest limit on error parity: only the O(1) dtype gates are parity-exact. A *data-dependent* comparison failure (e.g. a list-valued cell in an object column) is now observed only if it lands in the gathered candidates, so the indexed path can succeed where the scan raises. That needs pathological cell values to reach, but it is a real narrowing of the previous guarantee. Pinned by two structural regression tests that assert the *shape* rather than a wall-clock number: the predicate must see only candidate rows, and growing the graph 8× at fixed degree must not grow the number of rows it examines.
```
lmeyerov, line 15: once benefit proven & locked in pyg-bench, and happy with postive/negative tests on either side of opt boundary, and ci still green, can land both sides

lmeyerov:

a quiet host.

|---|---|---|---|---|---|---|

  (0.28 → 0.26 ms), i.e. flat.
  (6.67 → 6.66 ms), i.e. flat.

for.

real 5.3× growth.

warns about.

lmeyerov:

| bar | evidence |
| --- | --- |

### Beyond the bar

### Competitive result

   invisible. Now `.iloc[0:0]`.

own SF1 A/B.

PR #1783, opened 2026-07-26
perf(gfql): empty-left merge shrink + polars EORD/EID row-index dedup

lmeyerov:
Stacked on #1782 (base = `perf/gfql-indexed-edge-match-lazy`). Two independent, provably-safe reductions of graph-sized work.

## 1. Empty-left merge no longer scans the right side

`_lean_prefilter_right` shrinks `right` to the keys in `left` before a `how='left'` merge, but **declined when `left` was empty** — the one case where shrinking is both maximally profitable and trivially correct (no left key can match, and a left merge keeps only matched right rows).

Instrumenting a single-node named query on a 3.18M-node / 14M-edge graph showed the dominant work was `safe_merge(left=0x1, right=14000000x4, how='left')` — a zero-row intermediate joined against every edge in the graph.

Returning `right[0:0]` keeps identical columns and dtypes, so the merge still yields the same schema. **112.64 → 17.29 ms (6.5×)** on the affected shape.

## 2. Polars: reuse the synthetic edge id as the stable edge order

The native polars chain executor attaches `EID` via `with_row_index`, then later materializes a **second** `with_row_index` column (`EORD`) over that same frame. Both are 0..n-1 over the same frame in the same order — the second runs on the frame the first produced — so it is pure duplication of a graph-sized materialization. `with_row_index` was ~34 ms/query across the eager calls.

**148.22 → 135.43 ms (−8.6%)** on a seeded typed pattern with polars-engine indexes.

## Parity

- **108 polars-path cases, 0 differ, 0 raised** — 9 traversal shapes × 12 random seeds under `engine='polars'` with polars-engine indexes, hashed over column order, dtypes, row count and full content.
- **This is a new sweep, and it closes a real blind spot:** the existing 650-case differential runs `engine=auto`, which resolves to **pandas** and never exercises the polars lazy executor. A regression in that executor would have shipped clean.
- Index + lean-combine suites: no new failures.

## Scope honesty

Neither change flips a competitor result. Fix 1 helps shapes with an empty intermediate; IS1/IS5 are full `n→e→n` patterns whose intermediates are non-empty. Fix 2 is −8.6% on the polars path. They are stacked here because they are correct and general, not because they close the gap.


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+- **Native polars chain reuses the synthetic edge id as its stable row order**: when the executor had already added a synthetic edge-index column it also added a second, separate row-index column purely to restore input order at the end — two full `with_row_index` passes over the edge frame, and a redundant column carried through every intermediate. The existing synthetic id is already a contiguous, order-preserving row index, so it is now reused as the sort key and dropped once at the end. Only the pre-existing column is reused; when no synthetic id was added the separate order column is still created, so graphs that bring their own edge id are unaffected. Value-identical including row order (148.22 → 135.43 ms on a 3.18M-node / 14M-edge polars graph).
```
lmeyerov, line 16: once benefit proven & locked in pyg-bench, and happy with postive/negative tests on either side of opt boundary, and ci still green, can land both sides

lmeyerov:

a quiet host.

|---|---|---|---|---|---|---|

  (0.28 → 0.26 ms), i.e. flat.
  (6.67 → 6.66 ms), i.e. flat.

for.

real 5.3× growth.

warns about.

lmeyerov:

| bar | evidence |
| --- | --- |

### Beyond the bar

### Competitive result

   invisible. Now `.iloc[0:0]`.

own SF1 A/B.

lmeyerov:
## x-platform: clean

Full `graphistry/tests/compute/gfql/` on the GPU box —
`graphistry/test-rapids-official:26.02-gfql-polars`, `--gpus all`, so the cuDF and
polars-GPU lanes actually execute rather than skip:

| build | passed | skipped | xfailed | **failed** |
| master `4061e7ce` | 5414 | 25 | 15 | **0** |
| #1783 | 5414 | 25 | 15 | **0** |
| #1784 | 5450 (+36 = its own new tests) | 25 | 15 | **0** |

Skip counts identical across all three, so nothing was silently skipped into looking green.

Two scoping notes so the numbers aren't over-read:

- **#1783's lean-combine is pandas-gated by construction.** `_lean_engine_ok` returns `right`
  unchanged for cuDF / polars / dask, so the GPU engines cannot exercise that branch at all —
  this is inapplicability, not missing coverage. What the GPU run *does* validate for #1783 is
  the other half, the polars EORD/EID row-index change, which does run under GPU collect.
- **#1783's own new tests live in `tests/compute/test_chain_lean_combine.py`**, outside the
  `gfql/` tree, which is why its passed-count matches master rather than rising. Those are
  pandas-path tests and run in CI.

PR #1784, opened 2026-07-26
perf(gfql/polars): stop deduplicating semi-join key sides

lmeyerov:
# perf(gfql/polars): stop deduplicating semi-join key sides

Stacked on #1783 (`perf/gfql-lean-empty-left`). Review the top commit only.

## The bug

The native polars chain applied `.unique()` to every frame it fed into a `how="semi"` join:

```python
def _semi(df, ids_df, df_col, id_col):
    return df.join(ids_df.select(id_col).unique(), left_on=df_col, right_on=id_col, how="semi")

def _idframe_lf(lf, col):
    return lf.select(pl.col(col).cast(node_dtype).alias(NID)).unique()
```

A semi-join emits a left row **iff at least one matching right row exists**. Duplicate keys
on the right cannot change which rows come back, and — unlike an inner join — cannot
multiply them either. So the deduplication is a full hash pass over the key column bought
for no observable effect.

That would be merely wasteful if the key side were small. It isn't: **on an unfiltered hop
the key side IS the node table**, so this put O(N) work inside a query whose answer is
O(degree). The seeded single-hop planner builds two such key frames per hop.

## How it was found

Bisected, not guessed — and the first two hypotheses died:

| step | result |
| --- | --- |
| profile the seeded polars query | `collect_all` = 108.8 ms of 140 ms, 2 calls |
| time the semi-joins those calls perform | 3.17 ms (edges) + 1.07 ms (nodes) — **not the cost** |
| typed vs untyped edge | **−0.58 ms** — the edge predicate is not the cost either |
| edges 1.75M → 14M | 108 → 137 ms — only ~30 ms scales with E |
| **nodes 400k → 3.18M** (E fixed) | **40 → 129 ms — linear in N** |
| node width 5 → 31 columns | 136 → 136 ms — not attribute materialization |
| `nodes.select(key).unique()` alone, N=3.18M | **52.99 ms** ≈ half of the 105 ms |

Dumping the optimized plans confirmed it: `UNIQUE[keep_strategy: Any] BY None` over
`DF["key",…] PROJECT 1/32 COLUMNS`, repeatedly, feeding SEMI JOINs.

Isolated on raw polars, the rewrite is worth 7× on this shape:

| | with `.unique()` | without |
| --- | --- | --- |
| 3.18M-key side, tiny probe | 60.67 ms | **8.64 ms** (7.02×) |
| same, with the cast `_idframe` emits | 65.12 ms | **8.98 ms** (7.25×) |

## The fix

`.unique()` dropped only where the frame is provably a semi key side **and nothing else**:

- `chain.py` `_semi` (the helper hardcodes `how="semi"`)
- `chain.py` alias hop-window gate, and the next-edge endpoint gate
- `chain.py` endpoint gate in the two-hop fast path
- `pattern_apply.py` `start_nodes` gate
- `hop_eager.py` `_idframe_lf` — all four consumers (`allowed_source_lf`, `allowed_dest_lf`,
  `target_final_lf`, `frontier_lf`) are semi key sides

**Deliberately kept**: `named` in the alias pass. It feeds a `how="left"` join, where
duplicates genuinely *would* multiply rows. The eager multi-hop loop's `_idframe` is also
untouched — its frames flow into concat/anti-join bookkeeping, which is a separate argument
and a separate change.

## Correctness

- **380 differential comparisons, 0 divergences.** Two suites: the 108-case polars-path
  sweep, plus a new adversarial one built specifically because the usual synthetic graphs
  have unique node keys and would make this rewrite trivially equivalent — so it exercises
  **duplicate node keys, null ids, dangling edges, duplicate `start_nodes`**, across 11
  traversal shapes and 4 graph variants.
- `graphistry/tests/compute/gfql` + `test_compute_chain.py`: **identical failure sets**
  before and after (1 pre-existing GPU-driver collection error on this host, both sides).
- Row counts identical on every measured query.

## Tests

`test_engine_polars_semi_key_dedup.py` pins the *boundary*, not the speed (pandas is the
oracle throughout, so a divergence fails as a parity break rather than a guess):

- duplicates reaching a semi key side must not change results — 8 shapes × 3 graph variants
- duplicate node keys must not **multiply** output rows — the regression a wrongly-removed
  `.unique()` on the `how="left"` side would cause
- a **dangling endpoint must still be excluded** — pinning that the gate is load-bearing,
  not vacuous, so a future "this gate accepts everything, drop it" shortcut fails here
- duplicate `start_nodes` must be inert — the one key side a *caller* controls

## Not a benchmark hack

The rewrite keys on a structural property of the operator — a frame used only as a
`how="semi"` key side — and helps every traversal on every schema, engine and label. No
query, constant, dataset or column name is recognized anywhere. It removes work rather than
adding a special case, so there is no path it can make slower.

## Measured

Synthetic LDBC SNB SF1-shaped graph (3.18M nodes / 14M edges), `engine='polars'`,
seeded typed hop, min of 5:

| | before (#1783) | after | |
| --- | --- | --- | --- |
| E = 14M | 127.44 ms | **57.12 ms** | **2.23×** |
| E = 1.75M | 102.83 ms | **31.35 ms** | **3.28×** |

The larger speedup at *fewer* edges is the signature of the fix: the removed cost scaled
with node count, so it dominates more when there is less edge work to hide it.

Real-lane numbers to follow — this is a draft until the position-balanced SF1 A/B lands.


lmeyerov:
## Measured on the real LDBC SNB SF1 polars lane

Position-balanced ABBA/BAAB — 4 samples per build, each build in each slot position exactly
once, because this host has a ~5% slot effect that makes naive A,B,A,B uninterpretable.
Quiet host, host-wide perf lock held for the whole experiment, `rows_returned` validated
identical on every query. Pooled mean of per-slot medians.

| query | A = #1783 | B = this branch | delta |
| --- | --- | --- | --- |
| IS1 seed-lookup | 77.05 ms | **38.16 ms** | **−50.5%** |
| IS5 message-creator | 69.52 ms | **33.19 ms** | **−52.3%** |
| IS2 expand-order-limit | 91.84 ms | **52.72 ms** | **−42.6%** |
| IS7 message-replies | 14.10 ms | **11.66 ms** | **−17.3%** |
| IS4 message-content | 3.70 ms | 3.45 ms | −6.7% |

**4 faster, 0 slower**, every result value-identical.

### A discarded run, disclosed

The first A/B of this branch is **thrown away**, not reported: I committed the
`engine_arrays.py` change to the B worktree *while slots 3–8 were still running*, and
`--graphistry-repo-path` copies the local tree at each slot start — so "B" was not one build.
The numbers happened to look fine (−51%/−54%/−44%), which is exactly why they are not being
used. The table above is a full re-run on frozen trees.

That discarded run also showed IS4 at **+14.0%**, which would have been a regression. It did
not reproduce (−6.7% here). IS4 is seed-only — `MATCH (message:Message {id: X}) RETURN …`,
no hop — so no code path this PR touches is even reachable from it; the +14% was host noise
at a 3.5 ms scale.

lmeyerov:
Status: keeping this **draft** for one reason only — it does not yet have a pyg-bench lock-in
lane of its own. The lane in pyg-bench #101 pins #1782 and #1783; extending it to cover the
semi-join key-dedup invariant is the remaining item. Applying the same bar here that was set
for the PRs below it.

Everything else is in place: CHANGELOG entry, SF1 lane evidence (4 faster / 0 slower, rows
identical), 380 differential parity cases with 0 divergences, and tests that were
mutation-checked rather than merely run — removing the `.unique()` that must **stay** (the
alias frame feeding a `how='left'` join, where duplicates genuinely multiply rows) fails 7 of
them.

One correction to my own test hygiene, since it is visible in the history: I committed
`test_engine_polars_semi_key_dedup.py` without running pytest on it — the container-side
parity probes had covered the same ground, so the gap only surfaced at rebase. Three cases
failed because `start_nodes` stands in for the node frame, so a key-only frame made the seed's
`id` predicate raise before reaching the semi-join under test. Fixed in `604e37d2`.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
...semi key side only (the `_semi` helper, the alias hop-window and next-edge endpoint gates, the two-hop fast path's endpoint gate, the `start_nodes` gate, and the single-hop planner's id frames). It is deliberately **kept** on the alias frame that feeds a `how="left"` join, where duplicates genuinely would multiply rows, and the eager multi-hop loop is untouched (its frames also flow into concat/anti-join bookkeeping, a separate argument). Measured on a 3.18M-node / 14M-edge polars graph (LDBC SNB SF1-shaped), a seeded typed hop goes **127.4 → 57.1 ms (2.23×)** with identical row counts; at 1.75M edges **102.8 → 31.4 ms (3.28×)**, confirming the removed cost scales with node count, not edge count. Parity held across 380 differential comparisons covering duplicate node keys, null ids, dangling edges, duplicate `start_nodes`, and 11 traversal shapes; the gfql and chain suites show identical failure sets before and after. Pinned by tests that assert the boundary rather than the speed: duplicates reaching a semi key side must not change results or multiply rows, a dangling endpoint must still be excluded (the gate is load-bearing, not vacuous), and duplicate `start_nodes` must be inert.
```
lmeyerov: once benefit proven & locked in pyg-bench, and happy with postive/negative tests on either side of opt boundary, and ci still green, can land both sides

lmeyerov:

a quiet host.

|---|---|---|---|---|---|---|

  (0.28 → 0.26 ms), i.e. flat.
  (6.67 → 6.66 ms), i.e. flat.

for.

real 5.3× growth.

warns about.

lmeyerov:

| bar | evidence |
| --- | --- |

### Beyond the bar

### Competitive result

   invisible. Now `.iloc[0:0]`.

own SF1 A/B.

lmeyerov:
## x-platform: clean

Full `graphistry/tests/compute/gfql/` on the GPU box —
`graphistry/test-rapids-official:26.02-gfql-polars`, `--gpus all`, so the cuDF and
polars-GPU lanes actually execute rather than skip:

| build | passed | skipped | xfailed | **failed** |
| master `4061e7ce` | 5414 | 25 | 15 | **0** |
| #1783 | 5414 | 25 | 15 | **0** |
| #1784 | 5450 (+36 = its own new tests) | 25 | 15 | **0** |

Skip counts identical across all three, so nothing was silently skipped into looking green.

Two scoping notes so the numbers aren't over-read:

- **#1783's lean-combine is pandas-gated by construction.** `_lean_engine_ok` returns `right`
  unchanged for cuDF / polars / dask, so the GPU engines cannot exercise that branch at all —
  this is inapplicability, not missing coverage. What the GPU run *does* validate for #1783 is
  the other half, the polars EORD/EID row-index change, which does run under GPU collect.
- **#1783's own new tests live in `tests/compute/test_chain_lean_combine.py`**, outside the
  `gfql/` tree, which is why its passed-count matches master rather than rising. Those are
  pandas-path tests and run in CI.

PR #1785, opened 2026-07-27
perf(gfql/polars): make the chain combine proportional to the traversal result

lmeyerov:
# perf(gfql/polars): make the chain combine proportional to the traversal result

Stacked on #1784. **Draft** — needs a pyg-bench lane and an SF1 lane A/B before it can land.

## The recorded mechanism was wrong — for the third time in this area

The plan recorded the IS5 residual as `UNIQUE BY [key]` over a `UNION` of the per-step
**full** frames. Instrumenting the step frames says otherwise: they are **1 row each**, and
`LazyFrame.profile()` puts that `unique()` at **0.008 ms**. The union inputs were never the
graph.

## The real one: an empty-probe semi-join against the whole node table

`_combine_edges` runs the prev/next endpoint gates for **every** step — including *node*
steps, whose edge frame is `g._edges.clear()`, i.e. **zero rows**. The eager combine had a
skip for exactly that case. The collect-once rewrite lazified the step frames, and `.lazy()`
erases the height, so `if not is_lazy(df) and df.height == 0` went permanently false and the
skip **died silently**.

For step 0 the gate's key side is `prev_nodes = g._nodes` — the whole node table — and polars
builds the hash table on the **right** before discovering the left probe has no rows.
Isolated in raw polars: **6.99 ms for one empty-probe join at N=2M**, versus 0.22 ms against
a one-row key side. A chain pays one per node step.

This is the same class as #1783 (*empty left ⇒ don't materialize the right*), the semi-join
case rather than the left-merge case.

Second term found while profiling: node rows were materialized in **two** passes over the
node table (step ids, then the endpoints the first pass missed), 0.90 + 0.86 ms at N=2M.

## The change

- `_LazyShim` records the pre-lazy row count (tri-state); `_combine_edges` drops
  **known**-empty steps before the gates. A frame that arrives already lazy reports `None`
  and is planned normally — this keys on *known* empty, never on a guess.
- `_combine_nodes` → `_combine_node_ids` (ids only, **undeduped** — it is a semi key side)
  plus `_materialize_node_rows`, which unions both id sides first and reads the node table
  **once**. The row-level `unique(subset=[node], maintain_order=True)` is kept verbatim:
  those rows feed `how="left"` alias joins where duplicate keys multiply.

Both preconditions are structural — frame cardinality and semi-join algebra. No query,
schema, label or constant is referenced anywhere.

## Scaling, one dimension at a time (min of 9, local CPU)

| | before | after |
|---|---|---|
| **NODES** 250k → 4M @ E=2M | 10.17 → 45.48 ms | 7.60 → **17.17 ms** (2.65× at 4M) |
| node slope over 16× N | +35.3 ms | **+9.6 ms** (3.7× flatter) |
| **EDGES** 500k → 8M @ N=1M | 13.47 → 26.02 ms | 7.08 → 17.57 ms |
| edge slope | +12.6 ms | +10.5 ms (**unchanged**) |

The edge slope being unchanged is the point: this is a node-side fix and should not move it.
Same 2–3× on `n-e-n`, `n-e-n-e-n` and undirected.

## Parity

- **280** shape × graph combos (clean / self-loop / dangling / dup-keys / multi-edge /
  back-edge / isolated) and **400** dup-node-id combos × 5 repeats, compared as **full frames
  including row order**: **0 diffs**.
- `graphistry/tests/compute/` failure set byte-identical — 119 pre-existing before and after,
  **0 new**.
- 1003-case polars chain differential suite + #1784's semi-dedup suite pass. lint + typecheck
  clean.

## Mutation results (reintroduce the bug → the new tests must fail)

| mutation | result |
|---|---|
| drop empty-step skip | 2 fail |
| restore two-pass materialization | 1 fail |
| drop row-level node dedup | 18 fail |
| drop endpoint fold | 2 fail |
| drop node order restore | 34 fail |
| skip steps of **unknown** height | 1 fail |
| drop edge order restore | **0 fail — not caught** |
| drop `maintain_order` on node dedup | **0 fail — not caught** |

## What is NOT verified — disclosed rather than papered over

- **The edge order restore (`sort(EORD)`) is unpinned.** Probed 5 shapes × {5k, 60k, 300k}
  with the sort removed: the frame comes back in EORD order every time, so it is
  unobservable. Pre-existing defensive code, untouched.
- **`maintain_order=True` on the node dedup is unobservable** — the semi-join feeding it does
  not preserve left order, so "the first row in node-frame order" is not a property this
  engine guarantees. A/B confirms *which* duplicate survives is unchanged (400 dup-id combos,
  identical frames), but the docstring claim was **removed** rather than asserted. An initial
  version of that unit test asserted the unguaranteed order and was intermittently failing —
  caught before commit.
- **The endpoint fold could not be shown load-bearing end to end.** 6300 generated chain
  executions found no case where the step ids miss an edge endpoint, so the first version of
  that test was **vacuous** (it passed with the endpoint side removed entirely).
  `_materialize_node_rows` was extracted so the case can be constructed at the helper, where
  the mutation does fail. The fold is kept as a safety net, not deleted.
- **Remaining O(N)/O(E):** one full-frame semi probe per output frame (~0.9 ms per 2M rows,
  nodes and edges). That is the floor for "select rows by id" without an index-backed gather;
  a `NODE_ID`-index gather would make it O(result) but needs an eager id collect and breaks
  the collect-once design. Separate, larger change.
- Measured on local CPU only. No SF1 lane A/B yet — that plus a pyg-bench lock-in lane are
  what this needs before landing.


lmeyerov:

a quiet host.

|---|---|---|---|---|---|---|

  (0.28 → 0.26 ms), i.e. flat.
  (6.67 → 6.66 ms), i.e. flat.

for.

real 5.3× growth.

warns about.

lmeyerov:
## SF1 A/B against master — 3 faster, 0 over the gate

Master now contains #1782 + #1783 + #1784, so this measures **#1785's marginal contribution**,
not the stack's combined effect. Pooled **8 samples per build across two position-balanced
rounds**, rows identical on every query.

| query | master | +#1785 | delta | per-slot ranges |
| IS5 message-creator | 33.32 ms | **22.45 ms** | **−32.6%** | separated |
| IS1 seed-lookup | 39.24 ms | **29.19 ms** | **−25.6%** | separated |
| IS2 expand-order-limit | 53.24 ms | **43.09 ms** | **−19.1%** | separated |
| IS4 message-content | 3.76 ms | 4.09 ms | +8.6% | overlap |
| IS7 message-replies | 12.86 ms | 13.76 ms | +7.0% | overlap |

### The two flags, and what I am *not* claiming

At 4 samples these read **+14.6%** and **+10.8%** — over the gate. At 8 they are +8.6% and
+7.0% with fully overlapping distributions, the same decay the OLAP sweep's three flags showed
(one of which flipped sign entirely).

**But I am not calling them noise.** Both tilt *consistently* positive and B's distribution
sits above A's on each — IS4 `A[3.1…4.9]` vs `B[3.0…5.0]`, IS7 `A[11.2…14.4]` vs
`B[11.5…15.4]`. That reads as a small **real** constant cost, roughly **+0.3 ms and +0.9 ms**,
showing up only where the query is already tiny. It is mechanistically plausible: this PR adds
a tri-state height check and restructures node materialization, i.e. slightly more per-query
plan work. Below the project's 10% gate, disclosed rather than laundered.

### What this unblocks

This PR is what moves **IS5 across the line vs Neo4j**. On master today IS5 is 33.32 ms
against Neo4j's 27.28 — a **LOSE 1.22×**. With #1785 it is 22.45 ms, a **WIN 1.22×**. I had
earlier reported that flip as landed; it was measured on the full-stack tree and belongs to
this PR, not to master. IS2's flip is on merged code and is unaffected.

### Bar status

- **[1]** lock-in lane: passes all five probes (it was the only build to do so alongside #1784).
- **[2]** pos+neg tests either side of the boundary, mutation-checked — including one that
  caught an **engine-blind** probe calling `sort(EORD)` dead code when it is load-bearing
  under streaming collect.
- **[3]** CI green; GPU suite clean on device.

Issue #1797, opened 2026-07-27
CI: the test-polars py3.12 coverage cell is at its timeout ceiling; every polars test addition re-triggers a TIMEOUT reported as `cancelled`


lmeyerov:
## Summary

The `test-polars` py3.12 cell is at its wall-clock ceiling. It is the **only** cell in the matrix that runs coverage, and it shares one `timeout-minutes` with five siblings that do half its work. Adding a single test file to `bin/test-polars.sh` is now enough to push it over, and when it goes over GitHub reports the job as **`cancelled`** — which reads as unrelated fail-fast collateral, not as a budget failure.

This is a structural capacity limit, not a one-off. It will re-fire on the next polars test addition.

## Evidence

On PR #1794 (which appends one file to `POLARS_TEST_FILES`), `test-polars (3.12)` failed twice with **zero test failures**:

| attempt | duration | outcome |
|---|---|---|
| 2 | 10m15s | `cancelled` |
| 3 | 10m11s | `cancelled` |

Job log, attempt 3:

```
2377 passed, 13 skipped ... in 572.43s (0:09:32)
The operation was canceled.
```

The pytest phase alone was **9m32s of a 10m job budget**. Siblings (3.9/3.10/3.11/3.13/3.14) finish in **4m09s–5m51s**. A rerun does not clear it — it is 0-for-2.

Two things made this expensive to diagnose:

1. **A timed-out job is reported as `cancelled`**, indistinguishable at a glance from fail-fast collateral. The distinguishing evidence is the *duration*: a fail-fast cancel lands seconds after a sibling failure; a timeout lands at exactly `timeout-minutes`. Attempt 2 was 1 cancelled / 72 success / 1 skipped, so collateral was ruled out by evidence.
2. **`changed-line-coverage` then reports SKIPPED**, because it `needs:` the job that produces `polars-coverage-py3.12`. So one budget overrun silently disables a second gate.

## Root cause

`.github/workflows/ci.yml`:

- `test-polars` declares `timeout-minutes: 10` for the whole matrix job (was line 1579).
- Only the py3.12 cell runs `Polars tests with coverage` (`POLARS_COV=1`, was line 1618) **plus** the per-file `Polars GFQL coverage audit` **plus** an artifact upload.

So one cell carries roughly double the work of its siblings against a budget sized for the siblings. The timeout is simultaneously too tight for the cell that matters and too loose to be a useful hang detector for the ones that don't.

## Fix applied in #1794

[...]

PR #1805, opened 2026-07-28
test(gfql): make the polars lane see the 280 tests it was skipping everywhere

lmeyerov:
## The finding

A full local suite on py3.11 + polars reported 3 failures that reproduce identically on
`origin/master` (`233b64c8`). "Pre-existing, not mine" was the tempting read. It is the wrong
one: **all three fail because CI cannot see them, and one of them is a real product defect.**

polars is installed in **exactly one** CI lane — `test-polars`, driven by `bin/test-polars.sh`,
which runs a hand-maintained file list. Every other lane collection-skips a module-level
`pytest.importorskip("polars")` and every `polars` / `polars-gpu` parameter. So a polars-gated
test the list omits runs **nowhere**, and both lanes report green about it.

Measured against every job log of run
[30311675717](https://github.com/graphistry/pygraphistry/actions/runs/30311675717) (PR #1799):
**280 tests executed locally with polars installed and appear in ZERO CI job logs.**

## The three failures, and the cause of each

| test | assertion | cause |
|---|---|---|
| `index/test_indexed_bindings.py::test_explicit_polars_gpu_declines_indexed_helper_and_falls_back` | `ImportError: GFQL engine='polars-gpu' requires the RAPIDS cudf_polars stack` (`chain.py:1005`) | **the test is wrong** — it executes `engine='polars-gpu'` but gates only on `importorskip("polars")` |
| `test_viz_pipeline_conformance.py::test_case_regex_unicode_trick_matrix[toupper-eq-ss-fold]` | `toupper-eq-ss-fold: hand pin disagrees with the pandas mirror; assert [9] == [8, 9]` | **master is wrong** — pandas/polars silently disagree under pandas>=3 (#1802) |
| `…::test_case_regex_unicode_trick_matrix[tolower-turkish-dotted-i]` | `tolower-turkish-dotted-i: hand pin disagrees with the pandas mirror; assert [] == [7]` | same |

**#1 — evidence.** `grep` over all job logs: the test appears **9 times, every one `SKIPPED`, zero
executions.** It is collection-visible in the pandas lanes (`test-gfql-core` collects
`graphistry/tests/compute/gfql`) but skips there for want of polars, and its file is not in
`bin/test-polars.sh`, so the one lane that *has* polars never sees it. It has never run.

**#2/#3 — evidence.** These are not environment noise. pandas>=3 makes the Arrow-backed `str`
dtype the default, and Arrow's `utf8_upper`/`utf8_lower` are **simple** per-codepoint mappings,
where Python's (and pandas<3's object dtype) are **full**:

```
pd.__version__ == '3.0.5'
pd.Series(["straße", "istanbul İ"]).str.upper()  ->  ['STRAẞE', 'ISTANBUL İ']
pd.Series(["straße", "istanbul İ"]).str.lower()  ->  ['straße',  'istanbul i']
```

`row/pipeline.py:1521` implements Cypher `toLower`/`toUpper` with that accessor, so:

```
MATCH (n) WHERE toUpper(n.name) = 'STRASSE' RETURN n.id AS id
  engine='pandas'  ->  [9]
  engine='polars'  ->  [8, 9]      # Rust full case mapping — and the Cypher-correct answer
```

A **silent** cross-engine divergence: no decline, no NIE, just different rows — precisely what
these conformance suites exist to forbid. Filed as **#1802** with the repro and a sizing of the
fix. **CI's own `test-py3.12.lock` pins `pandas==3.0.3`**, so this is not an exotic local
environment; it is the environment CI runs. The test that catches it simply never executes.

## Sibling inventory — the cause is structural

280 tests, in 10 files, ran in no CI job of that run:

| file | invisible tests | why |
|---|---|---|
| `test_engine_polars_narrow_combine.py` | 120 | module-level gate, absent from lane |
| `index/test_indexed_bindings.py` | 40 | polars/polars-gpu params, absent from lane |
| `test_engine_polars_semi_key_dedup.py` | 36 | module-level gate, absent from lane |
| `test_viz_pipeline_conformance.py` | 31 | module-level gate, absent from lane |
| `test_engine_coercion.py` | 13 | polars params, absent from lane |
| `test_engine_polars_call_modality.py` | 12 | module-level gate, absent from lane |
| `test_rows_table_named_middle.py` | 10 | module-level gate, absent from lane |
| `test_reentry_caller_graph_immutability.py` | 9 | polars params, absent from lane |
| `test_rewrite_param_discard.py` | 7 | polars params, absent from lane |
| `cypher/test_lowering.py` | 2 | in the lane's `-k polars` phase, but the filter cannot select them |

Six whole modules were invisible, **four of them named `test_engine_polars_*`**. Also confirmed
invisible for a *different* reason and left alone here: `test_engine_polars_gpu.py` and
`index/test_index_gpu_edge_match.py` are GPU-gated — a separate lane gap.

## What this changes

- **`bin/test-polars.sh`** — adds the six invisible modules and the four whose polars parameters
  were invisible. 277 of the 280 now execute in the lane; the other 3 are the two renamed
  `test_lowering.py` tests (now selected by `-k polars`) and the polars-gpu test (now skipping
  honestly, its CPU half split out and running).
- **`test_polars_lane_completeness.py`** (new) — the durable part. It parses `POLARS_TEST_FILES`
  and fails when: a polars-mentioning test module is neither in the lane nor in a documented
  exemption; a module-level polars gate sits outside the lane; a listed path no longer exists; a
  polars-gated test in the `-k polars` phase is unselectable by that filter; or an exemption goes
  stale. Exemptions carry written reasons and are length-checked, so "" cannot re-open the hole.
  **Mutation-checked**: removing one entry from the array turns two of its assertions red.
- **The polars-gpu test is split** — the CPU half (`try_indexed_connected_bindings_state` must
  decline for `Engine.POLARS_GPU`) now runs wherever polars does; the execution half is
  `skipif`-gated on `cudf_polars`, mirroring `chain.py`'s own check.
- **The two unicode rows are `xfail(strict=True)`**, keyed on a probe of the pandas accessor's own
  behaviour — so they must still PASS on pandas<3, and the #1802 fix turns them into an XPASS
  that forces the marks to be retired. The hand pins stay at the Cypher-correct values.
- Renamed the two `test_lowering.py` polars tests that `-k polars` could not select.

Test-only. No product code touched, hence no CHANGELOG entry.

## Verification — CI green, 77/77

- **`test-polars` green on all six python versions**, and the behaviour is version-discriminating
  rather than suppressed:
  - py3.9 / py3.10 (**pandas 2.3.3**): the two unicode rows **PASS** — full case mapping, hand
    pins correct, xfail condition correctly not applied.
  - py3.12 / py3.13 / py3.14 (**pandas 3.0.3**): the same two rows **XFAIL** on #1802.
  - `test_explicit_polars_gpu_declines_indexed_helper` **PASSED**; the execution half
    **SKIPPED** for want of cudf_polars — instead of the ImportError it used to raise wherever it
    was actually run.
  - py3.12 lane: **2726 passed, 142 skipped, 3 xfailed** (was 2355 passed, 13 skipped), and the
    `-k polars` phase 51 passed (was 49). The polars per-file coverage audit passes unchanged.
- Locally (py3.11 + polars 1.43.1 + pandas 3.0.5): identical outcome; the completeness guard was
  **mutation-tested** (dropping one array entry turns two of its assertions red), not just
  observed green.
- `ruff check` clean; `mypy` reports no error in the new file.

## Ordering dependency — #1797's workflow patch should land first

It passed, but **by three seconds**. Measured job durations for `test-polars (3.12)`, the cell
that runs the suite under coverage:

| | duration | budget |
|---|---|---|
| master (`233b64c8`) | **8m21s** | `timeout-minutes: 10` |
| this PR | **9m57s** | `timeout-minutes: 10` |

Master was already at 84% of that budget before this change. #1797 records that adding a *single*
file previously tipped this cell into a timeout — which GitHub reports as `cancelled`, with zero
test failures in the log. The pending `ci.yml` split in the comment on #1794 (a dedicated
`test-polars-coverage` job at `timeout-minutes: 20`) should land **before** this, or the next
polars test — or a slightly slower runner — cancels the cell rather than failing it.

This PR changes no workflow file; that patch is untouched here.

### DIFF
--- a/bin/test-polars.sh
+++ b/bin/test-polars.sh
@@ -12,6 +12,12 @@ set -ex
 python -m pytest --version
 
 # Single source of truth for the polars test file list (CI reuses this script).
+#
+# COMPLETENESS IS ENFORCED, NOT REMEMBERED: polars is installed in NO other CI lane, so a
+# module gated by `pytest.importorskip("polars")` that is missing from this array runs
+# NOWHERE — it is collection-skipped everywhere else and silently reports green.
+# `graphistry/tests/compute/gfql/test_polars_lane_completeness.py` parses this array and
+# fails if any module-level polars-gated test file is absent from it (or listed but gone).
 POLARS_TEST_FILES=(
     graphistry/tests/compute/test_polars.py
     graphistry/tests/compute/gfql/test_engine_polars_hop.py
@@ -29,6 +35,20 @@ POLARS_TEST_FILES=(
     graphistry/tests/compute/gfql/test_polars_rows_entity_groupby.py
     graphistry/tests/compute/gfql/test_seeded_typed_hop_fastpath.py
     graphistry/tests/compute/gfql/test_residual_polars_native.py
+    # module-level `importorskip("polars")` files that previously ran in no lane at all
+    graphistry/tests/compute/gfql/test_engine_polars_narrow_combine.py
+    graphistry/tests/compute/gfql/test_engine_polars_semi_key_dedup.py
+    graphistry/tests/compute/gfql/test_engine_polars_call_modality.py
+    graphistry/tests/compute/gfql/test_engine_polars_gpu.py
+    graphistry/tests/compute/gfql/test_rows_table_named_middle.py
+    graphistry/tests/compute/gfql/test_viz_pipeline_conformance.py
+    # polars-parametrized cases inside otherwise-pandas modules: these files DO run in the
+    # pandas lanes, but their polars/polars-gpu parameters are skipped there for want of the
+    # wheel, so the polars lane is the only place those parameters can execute
+    graphistry/tests/compute/gfql/index/test_indexed_bindings.py
+    graphistry/tests/compute/gfql/test_reentry_caller_graph_immutability.py
+    graphistry/tests/compute/gfql/test_rewrite_param_discard.py
+    graphistry/tests/compute/test_engine_coercion.py
     # index tests exercise the seeded-index hook in the polars hop entry (hop.py) — without
     # them the hook dominates the now-thin file and trips its per-file coverage floor
     graphistry/tests/compute/gfql/index/test_index.py
--- a/graphistry/tests/compute/gfql/cypher/test_lowering.py
+++ b/graphistry/tests/compute/gfql/cypher/test_lowering.py
@@ -15913,7 +15913,7 @@ def test_node_dtypes_for_pushdown_fails_closed_on_unreadable_nodes() -> None:
     assert dict(dtypes) == {}
 
 
-def test_node_dtypes_for_pushdown_defers_the_conversion() -> None:
+def test_node_dtypes_for_pushdown_on_polars_defers_the_conversion() -> None:
     # Reading dtypes costs a full engine conversion, and only connected-join pushdown asks --
     # a path most queries never reach. Computing eagerly charged every string query for a
     # frame it discarded. Nothing may convert until a lookup actually happens.
@@ -15944,7 +15944,7 @@ def test_node_dtypes_for_pushdown_defers_the_conversion() -> None:
         filter_by_dict._read_node_dtypes = original
 
 
-def test_node_dtypes_for_pushdown_matches_the_full_conversion() -> None:
+def test_node_dtypes_for_pushdown_on_polars_matches_the_full_conversion() -> None:
     # polars -> pandas is DATA-dependent: an empty probe reports `bool` for a nullable Boolean
     # while the real conversion yields `object`. The gate must report what the executor sees,
     # so compare against the full conversion rather than asserting probe dtypes in isolation.
--- a/graphistry/tests/compute/gfql/index/test_indexed_bindings.py
+++ b/graphistry/tests/compute/gfql/index/test_indexed_bindings.py
@@ -1110,9 +1110,17 @@ def test_use_policy_sparse_serves_dense_declines(
         assert decisions[0]["reason"] == expected_reason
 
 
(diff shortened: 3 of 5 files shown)


lmeyerov:
**The `ci.yml` job split is no longer a prerequisite for this PR — #1814 removes the timeout risk without touching a workflow file.**

Context: this PR's `test-polars (3.12)` cell measured **9m57s against a 10-minute timeout — green by 3 seconds**, which is why it was queued behind a `ci.yml` split that only someone with `workflow` scope can push.

#1814 parallelizes `bin/test-polars.sh` under pytest-xdist instead (no workflow edit, no dependency change). Measured on CI: the 3.12 cell goes **501s -> 322s**, script step **484s -> 297s (1.63x)**. Applying that to this PR's cell gives **~373s (~6m13s)**, i.e. **~3.8 minutes of margin instead of 3 seconds**.

Verified there before relying on it: identical pass/skip **node-id sets** serial vs parallel (2417 ids, 2404 passed / 13 skipped), coverage a strict superset with zero lines lost, the `--cov-append` phase still appending into the xdist-produced data file, and a byte-identical `gfql-polars` coverage audit — so `changed-line-coverage` sees no change.

**Diffs are mechanically separate and rebase cleanly either way.** #1814 touches only the block *after* the `POLARS_TEST_FILES` array's closing paren (plus CHANGELOG); this PR touches only inside the array and the comment above it. #1814 also confirms it does not break `test_polars_lane_completeness.py`'s `^POLARS_TEST_FILES=\((.*?)^\)` parser — that anchored non-greedy match ignores the new `XDIST_ARGS=( ... )` array declared after it.

Suggested order: **#1814, then this PR.**

Issue #1804, opened 2026-07-28
GFQL: native polars bindings builder ignores rows(alias_prefilters=...) — 5 rows on pandas/cuDF vs 12 on polars


lmeyerov:
## What

`rows(binding_ops=..., alias_prefilters=...)` narrows the bindings table on pandas and cuDF and is **silently ignored** on the native polars engines. Same query, same `binding_ops`, different row count.

The generic path threads the hint into `_gfql_binding_ops_row_table`. The native polars row op calls its bindings builder with `binding_ops` and `attach_prop_aliases` only, so `alias_prefilters` never reaches it.

## Repro

Fixture and assertions are already in the tree — `graphistry/tests/compute/gfql/test_rewrite_param_discard.py::test_alias_prefilters_are_honoured_by_the_bindings_builder`, a 12-node / 12-edge graph:

```python
binding_ops = serialize_binding_ops([n(name="a"), e_forward(name="r"), n(name="b")])
g.gfql([rows(binding_ops=binding_ops,
             alias_prefilters={"a": [{"kind": "expr", "text": "a.id < 3"}]})], engine=ENGINE)
```

* `pandas` / `cudf`: **5 rows**, `max(a.id) < 3`
* `polars` / `polars-gpu`: **12 rows** — the whole unfiltered bindings table

## Severity

Bounded but real. The Cypher lowering that emits `alias_prefilters` **always keeps the equivalent post-join filter**, so Cypher-generated plans agree across engines and only lose the pushdown. **Hand-written GFQL that passes `alias_prefilters` without a matching post-filter gets engine-dependent answers with no error.** That is the silent-divergence class, just reachable only from the GFQL surface.

## Why filed now

The gap was known when #1790 landed and was pinned as a `pytest.mark.xfail(strict=True)` with a prose reason and **no issue number**, so there was nothing to close and nothing to find. Recording it here so the xfail can reference it and so widening the polars builder has a home.

Fixing it means porting the prefilter evaluator to the native polars builder — a feature, not a patch, which is why #1790 correctly declined to bundle it.

## Pinned

`test_alias_prefilters_are_honoured_by_the_bindings_builder` — strict xfail on `polars` and (as of the CB3 engine-parametrization) on `polars-gpu`, which is the same builder. When the builder learns the param, both markers must be deleted.

Issue #1803, opened 2026-07-28
GFQL: indexed bindings bypass never engages on engine='polars-gpu' (silent loss of the index path)


lmeyerov:
## What

`graphistry/compute/gfql/lazy/engine/polars/chain.py::_try_indexed_middle_polars` deliberately passes `Engine.POLARS_GPU` when the GPU execution target is active:

```python
engine = Engine.POLARS_GPU if active_target() == ExecutionTarget.GPU else Engine.POLARS
return try_indexed_connected_bindings_state(g, middle, engine=engine), True
```

but `graphistry/compute/gfql/index/bindings.py:307` — the first gate in `try_indexed_connected_bindings_state` — admits only three engines:

```python
if (
    engine not in (Engine.PANDAS, Engine.CUDF, Engine.POLARS)
    ...
):
    return None
```

So on `engine='polars-gpu'` the indexed fixed-hop bindings bypass **always declines**, and the query silently runs the canonical scan instead.

## Evidence

Measured on the dgx GB10 box, `graphistry/test-rapids-official:26.02-gfql-polars`, `docker run --gpus all`, cuDF 26.02 / cudf_polars present. `index_trace()` for `[n({id:1}, name='a'), e_forward(name='r'), n(name='b'), rows()]` with `index_policy='force'` on a `gfql_index_all()`-ed graph:

```
{'operation': 'indexed_traversal', 'seam': 'connected_bindings', 'engine': 'polars-gpu',
 'served': False, 'reason': 'unsupported_engine', 'path': 'scan',
 'decision_reason': 'unsupported_engine'}
```

The same query on `engine='polars'` is served by the bypass.

## Impact

**Not a wrong answer** — every value/parity case in `graphistry/tests/compute/gfql/test_rewrite_param_discard.py` passes on `polars-gpu` (45 passed / 2 xfailed on the GPU run). What is lost is the optimization, with no signal at all: a user who indexes a polars graph and runs it on the GPU target gets the scan path and no indication that the index was skipped. That matters because `polars-gpu` is one of the two engines the perf work targets.

## Why it looks like an oversight rather than a decision

[...]

PR #1807, opened 2026-07-28
test(gfql): run the #1788/#1790 suites on all four engines, in a lane that exists

lmeyerov:
CB3 (a) and (b): the two test files the owner flagged, both engine-parametrized properly and both now running in a lane that exists.

## What was wrong

| file | before |
|---|---|
| `test_rows_table_named_middle.py` (#1788) | pandas + polars only; module-level `pytest.importorskip("polars")`; **not** in `bin/test-polars.sh` → the `test-gfql-core` lane installs no polars, so the whole file skipped there and **ran in no CI lane at all** |
| `test_rewrite_param_discard.py` (#1790) | named cuDF but **skipped polars-gpu outright**; also absent from `bin/test-polars.sh`, so every polars assertion in it was dead in CI |

Both files pin *which table a query returns* — engine-agnostic semantics, not a polars question.

## What this changes

Fixed engine list (`pandas`, `cudf`, `polars`, `polars-gpu`) — deliberately **not** `available_nonpandas_engines()`, which builds its list by importability so a missing engine vanishes from the report instead of showing as SKIPPED. Both files added to `bin/test-polars.sh`; the module-level `importorskip` is gone, which also makes 5 pandas cases live in `test-gfql-core`.

## Two probe traps, both hit by trying rather than reasoned about

**1. A probe that runs the shape under test disarms its own file.** My first version smoke-tested with `rows(table='edges')`. Reverting the #1788 `table` guard then made the probe raise, and all 20 parameters reported **SKIPPED instead of failing** — a regression disguised as a missing environment. The smoke query is now a plain traversal with no `rows()` in it.

**2. A probe that swallows every exception is worse than no probe.** A transient `MemoryError: std::bad_alloc: out_of_memory: CUDA error (failed to allocate 32 bytes) … cudaErrorMemoryAllocation` in a fresh GPU container silently dropped cuDF from a run that otherwise passed. A skipped GPU parameter reads as evidence of passing. `engine_skip_reason` (new, in `polars_test_utils.py`) now **classifies**: a missing module skips, a recognisable GPU-stack error skips **with its text quoted in the skip reason**, and any other failure **propagates**.

## GPU receipts

dgx GB10 · `graphistry/test-rapids-official:26.02-gfql-polars` · `docker run --gpus all` · cuDF **26.02.01** / polars **1.35.2**:

```
45 passed, 3 xfailed, 0 skipped        (repeated — all four engines running)
```

**Mutation-checked on the same box**, reverting the #1788 and #1790 `table` guards:

```
24 failed, 21 passed, 3 xfailed, 0 skipped
   6 [pandas]   6 [cudf]   6 [polars]   6 [polars-gpu]
```

Evenly spread across all four engines, so the added parameters are not decorative.

## Two real gaps the new parameters surfaced — pinned, not papered over

* **#1803** — the indexed bindings bypass gate (`index/bindings.py:307`) admits only `(PANDAS, CUDF, POLARS)`, while the native polars chain hands it `Engine.POLARS_GPU` whenever the GPU target is active. polars-gpu therefore **always** reports `served: False, reason: 'unsupported_engine'` and silently runs the canonical scan. Values are unaffected; the optimization is lost with no signal. Strict xfail.
* **#1804** — the native polars bindings builder never receives `alias_prefilters`: 5 rows on pandas/cuDF vs 12 on polars for the same query. It was already xfailed before this change, but **with no issue number**, so there was nothing to close and nothing to find. Now filed and referenced from the marker.

Strict xfails in both cases, so whoever fixes either is told to delete the marker rather than leaving a stale "known gap" behind.

## Runtime delta: zero

No production line changed — the diff is two test files, one shared test helper, `bin/test-polars.sh` and the CHANGELOG. Per CB5 no pyg-bench lane run is required, and that claim is checkable from the diff rather than merely asserted.

## Note for the reviewer

`bin/test-polars.sh` is also touched by #1801 and by the CA4 allowlist PR; the edits are to different lines of the same array and are independently mergeable.


lmeyerov:
**Overlap note for whoever merges this.** #1805 (`fix/gfql-polars-lane-invisible-tests`, opened concurrently) also adds `test_rows_table_named_middle.py` and `test_rewrite_param_discard.py` to `bin/test-polars.sh`, and restructures that array — so the two will conflict textually even though they agree on the outcome.

Left in here deliberately rather than removed: this PR's whole point is that these two files' non-pandas parameters actually execute, and dropping the lane lines would make that contingent on a PR that has not landed. If #1805 goes first, resolve by keeping #1805's array and dropping the two lines from this diff — nothing else here touches that file.

The substance of this PR (the four-engine parametrization, the classified availability check, the GPU receipts, and the two strict xfails for #1803/#1804) is independent of #1805 and does not overlap it.

Also worth flagging against #1797 (the polars coverage cell sitting at its timeout ceiling): this PR's additions measured ~15s per lane run locally, and CI came back green on all 77 checks including every `test-polars` matrix cell — so it did not tip that cell over. That is an observation from one run, not a guarantee.

graphistry/tests/compute/gfql/polars_test_utils.py
```diff
@@ -181,3 +181,58 @@ def assert_surfaces_agree(res_a, res_b, label):
...
+    """
+    if isinstance(ex, ImportError):
+        return f"{type(ex).__name__}: {ex}"
+    text = f"{type(ex).__name__}: {ex}".lower()
+    if any(marker in text for marker in _GPU_ENVIRONMENT_MARKERS):
+        return f"{type(ex).__name__}: {ex}"
+    return None
+
+
+def engine_skip_reason(engine, smoke):
```
lmeyerov: static typing

lmeyerov:
Merged master (`a10173575`) in as `f6fa7fb7e`. Both conflicts resolved, and the conflict exposed two defects in this branch that I fixed rather than carried forward.

**Defect 1 — `bin/test-polars.sh` would have run two modules twice.** This branch adds `test_rows_table_named_middle.py` and `test_rewrite_param_discard.py` to `POLARS_TEST_FILES`. But #1805 — which *is* this branch's merge-base (`49db91cc9`) — already added both, at lines 43 and 50. So the addition was redundant at the moment it was written, and taking the branch side would have listed each path twice and run each module twice. That lane's py3.12 cell is the coverage cell and has been up against its budget: **442s on master's latest green run, cap 600s.** Resolution is master's file verbatim, which also keeps #1814's xdist plumbing that the branch side predates. `bin/test-polars.sh` no longer appears in this PR's diff at all.

**Defect 2 — a false claim in the CHANGELOG.** The entry said *"neither file was in `bin/test-polars.sh`, so `test_rows_table_named_middle.py` … ran in **no CI lane at all**"*. Same root cause: both files were listed at the base, so the polars params did run. Rewritten to the residual that is actually true — the suites were parametrized over pandas + polars only, so **cuDF and polars-gpu never exercised the `table` guards these suites exist to pin**; and the module-level `pytest.importorskip("polars")` skipped the whole file in `test-gfql-core`, so its pandas cases ran only in the polars lane. Every measured claim is untouched (45 passed / 3 xfailed / 0 skipped on GB10; 24 mutation failures, 6 per engine).

**Typing on `engine_skip_reason`, per review.** The engine → required-modules map was a dict literal rebuilt inline on every call inside the `for`; it is now a module-level `_ENGINE_REQUIRED_MODULES: Mapping[str, Tuple[str, ...]]`. Signature is `(engine: str, smoke: Callable[[], object]) -> Optional[str]`. `gpu_environment_reason`'s `"str | None"` string annotation is normalized to `Optional[str]` to match — this module is imported under the py3.8 lanes, so the deferred-annotation form was load-bearing, not stylistic. flake8 and mypy 1.18.2 clean under the repo config.

Net diff vs master is now CHANGELOG + the three files under `graphistry/tests/compute/gfql/` — no production line and no CI-manifest line changed.

PR #1850, opened 2026-08-04
Fix indexing.rst doc-example failures

vaimdevs:
```
##Fix
* Fix indexing.rst doc-example failure — define new_edges_df before use in the invalidation example (was raising NameError).
```

### DIFF
--- a/docs/source/gfql/indexing.rst
+++ b/docs/source/gfql/indexing.rst
@@ -177,6 +177,7 @@ time). Consequences:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


lmeyerov:
Your fix landed on master via #1933 (cherry-pick of `4e7654bce`, authorship preserved) so it didn't have to wait behind #1721's review — this PR will show as resolved/empty once #1721 merges master forward. Thanks for catching the NameError; verified the `amount` column exists in the page's own `edges_df` and that `assign` returning a new frame is exactly the invalidation lesson the example teaches.

PR #1852, opened 2026-08-05
perf(gfql): HOLDBACK — host-thread the dense-count kernel O(E) passes

lmeyerov:
Stacked on #1851 (elision core). Holds the concurrency half of round 3 pending owner judgment: a small process-lifetime thread pool chunks the remaining bincount + gather-sum above 1M edges (both release the GIL). Open questions the owner flagged: library-owned pools vs host pools (oversubscription, fork safety), and whether the thread-free round-4 lever (filter projection pushdown) closes the gap without this. Do not merge unless both answers favor it.

Pins: threaded-vs-plain equality on uneven chunk boundaries; kernel-level threaded/unthreaded/oracle triple.


lmeyerov:
Closing as superseded (owner call). This held back host-threading of the dense-count kernel's O(E) passes; the question it addressed has since been answered differently and better: #1851 (gather-sum elision), #1858/#1859 (per-type facts, receipted -18.6% on q8@100k), and #1870-#1872 (precomputed degree facts) replace the O(E) passes with an O(N) dot rather than parallelizing them -- probe reads q8@100k at ~3ms vs ~13ms, receipted lane in flight. Threading the pass it eliminates is moot, and the original concern about multiprocessing opts inside polars' domain stands.

PR #1856, opened 2026-08-06
perf(gfql): grouped-aggregate fast path declares its plan columns (round-4 follow-up)

lmeyerov:
The declared follow-up from #1853: the single-hop grouped-aggregate fast path (graph-benchmark q1/q3/q4 class) passes `project=` to its cached filters — ids plus exactly the props the compiled plan references — on every engine, under the same static monotone admission rule (`_filter_project`). Both its fused polars lane and the eager/pandas tail were verified to read only those columns.

A referenced prop missing from the node frame stays excluded from the projection, so the existing missing-prop decline fires exactly as before (pinned: decline parity with the eager twin, including the error case).

Tests: narrow-width + decoy-exclusion pins per engine with eager-twin value parity; missing-prop decline parity; the existing 199-test differential net for this lane stays green.

For owner review.


lmeyerov:
DGX container matrix on head `0d11ab18`: **8,372 passed, 53 skipped, 19 xfailed** across the full GFQL tree (real cudf/polars-gpu arms).

lmeyerov:
Cross-platform container matrix (`c4b69aea1`): **1,955 passed** — the projection width/decoy/missing-prop pins and the served-shapes value differential across pandas + polars + polars-gpu + cudf, all on real GPU.

lmeyerov:
**Lock-lane receipts (build `c4b69aea1`, VALID both scales):**

Target cells (grouped-agg class) vs the r4 base lanes — all within slot noise: q1@100k 5.72x (base 5.81x), q3@100k 3.46x (3.57x), q4@100k 1.45x (1.33x), and 20k likewise. **No harm anywhere; no receipted improvement either** — expected in hindsight: this suite's node/edge frames are only 3–5 columns wide, so projecting to ids+props trims almost nothing. The change's thesis (wide production frames) is real but unmeasurable on this benchmark.

Incidental but useful: q8@20k read 2.15ms on this lane vs 2.51/2.59 on the two sibling lanes over the same base — the r4 overhead is real on the cross-lane median (~+0.5ms) but per-lane noise straddles the 2.35 floor, which sharpens the case for #1857's deterministic fix rather than relying on any single lane.

Per the stated merge condition ('non-regression + improvement, locked'): **non-regression ✓, improvement ✗ on this suite** — same decision shape as #1854: (a) merge as architecture with the no-harm receipts (it simplifies the WHEN story and costs nothing measurable), (b) hold for a wide-frame benchmark that can show the benefit, (c) close. Owner's call; CI is a verified 74-job green either way.

lmeyerov:
**Motivation found and measured** (owner's question: any real scenario?). On production-width frames — 60 decoy node columns + 10 decoy edge columns (event-table shape), 200k nodes / 1.5M edges — the grouped-agg projection is **2.18× faster end-to-end**: 52.51 → 24.11ms vs current master, value-identical (exploratory micro A/B on the dev box, engine=polars, PYTHONPATH-pinned trees). The benchmark suite's 3–5-column frames were the one shape with nothing to trim. Proper receipted wide-variant lane available on request before merge.

PR #1873, opened 2026-08-12
feat(gfql)!: AUTO treats polars as a first-class compute engine (+ index gate agreement, typed degree facts)

lmeyerov:
**Default-policy change — needs explicit owner sign-off even when green.**

`resolve_engine(AUTO)` on polars frames now returns POLARS. The legacy semantics — polars is an input format, coerce to pandas — survive under an explicit name, `resolve_input_engine`, used by the surfaces that genuinely compute in pandas (layouts, hypergraph, ComputeMixin's degrees/materialize/topological). Migrating any of those to native polars later = flipping one call site.

## Why

Three AUTO gates had grown around the legacy mapping (legacy resolver, #1849's query router, #1843's index gate) and they **already disagreed** — mixed and edges-only frames routed differently per surface, which is the #1767 cliff class: an index built for one engine meets a query routed to another, every hop declines, answers stay correct so nothing fails.

## Load-bearing details

- **NIE fallback made explicit**: when native polars declines, AUTO serves via pandas *by statement*, not by accident of the legacy mapping.
- **Transitional policy guard at the single public entry**: with a policy attached, AUTO serves via pandas until the polars postload/postchain hook gap is fixed. Commented delete-me-when-fixed; the pins assert the hook *contract*, not this mechanism (owner: the old `policy_disables_the_route` pinned a workaround's side-effect as public interface).
- **A latent wrong-answer surfaced**: `pd.api.types.is_numeric_dtype` returns a *confident False* for polars dtypes — polars Float64 classified as a **string** in the pushdown planner. Polars dtypes now classify first-class.
- Doctrine pins flipped to invariants per the never-pin-accidents rule (mixed-frames bridge, hop-bridges-to-pandas, engine_mismatch-decline, nullable-Boolean omission).

## Folds in #1843 (index AUTO polars rework)

The two halves are atomically coupled — the index-gate agreement needs modern AUTO to exist, and modern AUTO leaves the old index gate red — so they land as one unit (#1843 closed with pointers). Its content here: `resolve_index_engine` *derives from the frames* (narrowing to pandas only when both aren't eager polars — the index can't gather LazyFrame rows) instead of post-correcting pandas on a polars path; typed per-relationship **degree facts** with identity anchored to the bound edge frame; the cross-engine AUTO agreement matrix (`test_auto_engine_agreement.py`).

Two more route pins flipped on the way to green: string predicates and bare float comparisons are now rendered by the WHERE pushdown on polars (the broken classifier was why they weren't), so "query raises upstream" and "float reaches the residual lane" were accidents — the pinned contracts are pandas-oracle parity and NaN-normalization-at-ingest.

Perf receipts live in pyg-bench (q8@100k 13.5→3.90 ms, WIN 2.6× vs warmed Kuzu; lane `graphbench-100k-degree-20260812`).


graphistry/compute/gfql/lazy/engine/polars/dtypes.py
```diff
@@ -22,6 +22,14 @@
     from graphistry.compute.typing import PolarsFrame, PolarsT
 
 
+def is_polars_dtype(dt: object) -> bool:
```
lmeyerov: object bad

graphistry/compute/filter_by_dict.py
```diff
@@ -39,6 +48,9 @@ def _is_numeric_dtype_safe(dtype: Any) -> bool:
 
 
 def _is_string_dtype_safe(dtype: Any) -> bool:
+    from graphistry.compute.gfql.lazy.engine.polars.dtypes import is_polars_dtype, is_stringlike
+    if is_polars_dtype(dtype):
+        return is_stringlike(dtype)
```
lmeyerov: move generic helpers out (dtypes.py?)

Issue #1891, opened 2026-08-14
design: OPTIONAL MATCH semantics rework — seed-gate bypasses, silent inner-joins, data-dependent polars outcomes


lmeyerov:
Round-003 amplification (agent-01: 96 hand-scored cells vs openCypher at the #1883 head) mapped the full extent of round-002's BUG-1 and found two more classes. **9 wrong-value shapes, all engine-AGREEING** (parity suites structurally blind):

- **Seed-gate bypass (silent inner-join)**: the validator gate (`lowering.py:9166`) fires only on `len(matches)==2 && source_alias==seed_alias`; aggregates (grouped and ungrouped — count(*) answers 2 vs Cypher 5), zero-arm aggregates (EMPTY frame), and plain `RETURN x.v` alias-only projections all bypass it and drop unmatched seeds. Paradox: the two-optional-arm twin answers CORRECTLY — the mechanism exists; the gate over-blocks the shape it could serve.
- **Zero-arm WITH-prefix data loss**: `MATCH (p) WITH p OPTIONAL MATCH (p)-[:NOPE]->(x) RETURN p.name, x.v` returns the right row count with **every carried seed property NULLed** — both engines.
- **NEW: polars outcome is DATA-dependent**: identical query text flips ANSWER → NIE('select') when the optional arm empties, and nested optionality flips ANSWER → bare `polars.SchemaError` (null-vs-i64 join key). "Runs natively on polars" is not a stable property of a query.
- round-002 BUG-3 extends to OPTIONAL re-entry (KeyError on carried scalar next to aggregate); typed-gate MESSAGES assert falsehoods (wrong blame, semantics-changing suggestions).

Bright spots hand-verified: OPTIONAL-first zero-match, WHERE-inside, two-arm left joins, ORDER BY null placement. Full matrix + 5-whys in plans/gfql-release-amplification/rounds/round-003/findings/agent-01/. 16 pins (8 strict-xfail, 6 green, 1 invariant sweep, 1 message audit) land via #1883. Absorbs the OPTIONAL-MATCH parts of round-002 BUG-1/BUG-3.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lmeyerov:
[...]

Issue #1888, opened 2026-08-14
design: endpoint-closure contract — five surfaces, five answers for one dangling-edge graph


lmeyerov:
Round-002 amplification (agent-01, 385-cell degenerate matrix at the #1883 head) consolidated what #1808/#1739 were facets of: **there is no defined contract for edges whose endpoints are absent from a bound node table**, and five surfaces answer differently for the same fixture (nodes [0,1]; edges 0→1, 1→2 dangling-dst, 5→6 both-missing):

| surface | answer |
|---|---|
| pandas chain | prunes to 1 edge |
| polars chain (unconstrained fast path) | keeps ALL 3 — never semi-joins edges vs nodes (F-01 HIGH); **attaching a policy flips the value** 3→1 |
| cypher rows (both engines) | keep dangling-dst (2) |
| polars cypher count(*) | drops it (1) — **projection says 2, count says 1 within the same engine** (F-02 HIGH) |
| hop | pandas SYNTHESIZES a phantom NaN node (upcasting int64→float64); polars keeps the edge but not the node (F-03) |

**Also**: #1808 is now stale — the polars row path matches pandas (the side #1808 argues is wrong); only the polars aggregate path kept the Cypher-correct semantics.

**Proposed contract** (matches Cypher + #1808's analysis): when nodes are bound, a pattern edge matches only if BOTH endpoints resolve to node rows — one rule, all surfaces, engines, and policy states; endpoint-closure invariant test fences all five surfaces. Strict xfails asserting the converged answers land via #1883; repros + 5-whys in plans/gfql-release-amplification/rounds/round-002/findings/agent-01/. Supersedes-and-absorbs the fix work of #1808/#1739.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lmeyerov:
**Contract convergence verified on master `e6625ed28`** — the five-surfaces-five-answers matrix now yields ONE answer everywhere; recommending close after review.

Rerunning the issue's dangling-edge matrix (chain / cypher-rows / cypher-count / hop x pandas / polars / cuDF):

```
chain/pandas        -> [(0, 1)]
cypher-rows/pandas  -> [(0, 1)]
cypher-count/pandas -> 1
hop/pandas          -> nodes [0, 1], dtype int64, edges [(0, 1)]
chain/polars        -> [(0, 1)]
cypher-rows/polars  -> [(0, 1)]
cypher-count/polars -> 1
[...]

PR #1894, opened 2026-08-14
fix(gfql)!: OPTIONAL MATCH null-extension semantics (#1891)

lmeyerov:
Umbrella 2 of the fix cycle, stacked on #1893. Routes the seed-gate-bypassed shapes onto the proven left-join mechanism per the round-003 finding; 33 strict-xfail pins are the acceptance record (all flipped; 50/50 in the OM suite; full-tree failure set byte-identical to baseline). Deliberate TYPED declines remain for whole-row-alias+aggregate, DISTINCT-over-alias, varlen-arm aggregates, and polars scalar-carry (AUTO's coerced fallback serves) — #1891 narrows to those + the newly-noted pre-existing mid-stage `WITH p, count(x)` inner-join. Behavior changes CHANGELOG'd.


lmeyerov:
**Known regression to fix before merge (GPU sweep caught it)**: the rework's residual gate now DECLINES `MATCH (n:Single) OPTIONAL MATCH (n)-[r]-(m) WHERE m.num = 42 RETURN m` — an undirected optional arm + WHERE + whole-row projection the old null-fill path SERVED. It regresses on ALL engines; `test_string_cypher_formats_optional_match_projection_on_cudf` is merely the only pin of the shape (which is why local gates missed it — pandas/polars twins of that pin don't exist and will be added with the fix). Fix queued behind the in-flight #1888 work (same worktree); do not merge until this flips back to answering.

lmeyerov:
**Second rework regression (CI tck-gfql @ 671b960bc):** `test_issue142_refreshed_xfail_row_promotions_execute_direct_cypher[match7-24]` now raises `Duplicate Cypher alias 'a' is not yet supported` (lowering.py:3752). The widened connected-OM predicate captures multi-MATCH shapes with repeated aliases that the direct-Cypher promotion path previously executed. Latent since this PR's first push — tck-gfql was *skipped* on earlier heads because it gates behind the lint job that was failing (lane-gated single coverage, same slip pattern as the cudf-only pin). Fix in flight on the stacked #1895 branch alongside the whole-row-projection regression; both remain merge blockers for this PR.

lmeyerov:
**Both rework regressions are fixed on this branch** (`294f30618`, cherry-picked from the stack tip where they were developed):

1. *Whole-row OM projection* — root cause was in `_eligible_optional_projection_query`: the parser mirrors an optional arm's WHERE into `query.where`, so the null-fill planner wrongly declined arm-owned WHERE as if it were a base-MATCH WHERE; the connected-OM lowering defers these shapes to that (absent) plan and the residual gate declined. Eligibility now accepts arm-owned WHERE (`query.where == matches[1].where`), the polars entity renderer learned label-flag rendering, and `_apply_optional_null_fill` went engine-agnostic — repro answers `[{'m': '(:A {num: 42})'}]` on pandas AND polars, with `[{'m': None}]` null-extension when the arm fails. +10 OM tests (5 shapes × both engines).
2. *TCK duplicate-alias routing* — self-loop optional arms (`(a)-[r]-(a)`) now decline off the connected-OM path back to direct-Cypher promotion; bound-alias-only projections go through the row guard. TCK fully green locally (3834 passed / 699 xfailed) incl. three scenarios fixed outright (match7-24, match7-13, match-where6-3); companion 4-key xfail-contract update pushed to tck-gfql branch `fix/gfql-1891-optional-match` (scenarios now *succeed*: match-where6-2, match7-29/30/31).

Merge blocker lifted pending CI. Guards: surface baseline regen for the intentional lowering.py growth; hygiene pins per convention.

lmeyerov:
**GPU verification (dgx GB10, RAPIDS 26.02, @ stack tip b3e69f108):** `test_string_cypher_formats_optional_match_projection_on_cudf` **passes** (1 passed, 3.05s) — the #1650 flattened-projection fix for the null-fill path is confirmed on the shared pandas/cudf projector. OM + divergence + hop pin suites on GPU: 113 passed / 3 xfailed / 0 failed. Both rework regressions are now closed with the column contract pinned on pandas+polars as well (no single-lane holders). Remaining gate: CI green on this head (watching flatten.py coverage floor + tck manifest).

graphistry/compute/gfql_unified.py
```diff
@@ -172,7 +170,7 @@ def _apply_optional_null_fill(
         message="Cypher OPTIONAL MATCH null-row alignment could not recover matched seed identities",
         suggestion="Use a simpler OPTIONAL MATCH projection shape in the local compiler.",
     )["ids"]
-    if not hasattr(matched_ids, "tolist"):
+    if not (hasattr(matched_ids, "tolist") or hasattr(matched_ids, "to_list")):
```
lmeyerov: Avoidable dynamic typing

graphistry/compute/gfql/lazy/engine/polars/projection.py
```diff
@@ -61,18 +62,19 @@ def _native_scalar_text_expr(col: str, dtype: Any) -> Optional[Any]:
...
 
     cols = list(rows_df.columns)
     if alias not in cols:
         return None
-    # single-entity only (no prefixed alias columns), no label rendering
+    # single-entity only (no prefixed alias columns), no node-type rendering
     if any(str(c).startswith(f"{alias}.") for c in cols):
         return None
-    if "type" in cols or any(str(c).startswith("label__") for c in cols):
+    if "type" in cols:
```
lmeyerov: Should this be a symbol or otherwise lifted / structured?

graphistry/compute/gfql/lazy/engine/polars/projection.py
```diff
@@ -87,18 +89,35 @@ def _native_node_entity_text_expr(rows_df: Any, alias: str, exclude: Any) -> Opt
         and not str(c).startswith("__") and not str(c).startswith("label__")
         and (str(c) not in internal or (include_id and str(c) == "id"))
     ]
+    # mirror _node_label_text: ":Name" per true boolean label__Name flag, column order
+    label_cols = [
+        str(c) for c in cols
+        if str(c).startswith("label__")
+        and str(c).split("label__", 1)[1] not in {"<NA>", "None", "nan"}
```
lmeyerov: Should this set be externed DRY ?

bin/ci_cypher_surface_guard_baseline.json
```diff
@@ -13,5 +13,5 @@
       "max_properties": 0
     }
   },
-  "lowering_py_max_lines": 9260
+  "lowering_py_max_lines": 9454
```
lmeyerov: We should queue refactoring lowering.py to ensure utils, helpers, specializations, etc more properly DRY, factored out, etc as 9K loc smells bad

graphistry/compute/gfql/cypher/reentry/execution.py
```diff
@@ -85,7 +85,7 @@ def reentry_validation_error(
         value=value,
         suggestion=suggestion,
         language="cypher",
-        **extra_context,
+        **extra_context,  # type: ignore[arg-type]
```
lmeyerov: These dynamic typings all smell

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -2705,6 +2705,44 @@ def _return_references_optional_only_alias(
     return False
 
 
+def _return_references_only_bound_aliases(
+    query: CypherQuery,
+    *,
+    alias_targets: Mapping[str, ASTObject],
+    params: Optional[Mapping[str, Any]] = None,  # hygiene-ok: explicit-any -- Cypher query parameters are arbitrary user scalars
+    bound_nullable_aliases: Optional[AbstractSet[str]] = None,
```
lmeyerov, line 2735: Any is wider than a scalar...

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -7664,6 +7692,17 @@ def _is_connected_optional_match_query(query: CypherQuery) -> bool:
                     or (el.to_fixed_point if hasattr(el, "to_fixed_point") else False)
                 ):
                     return False
+    # Reject repeated node aliases within a clause (e.g. self-loop
+    # ``OPTIONAL MATCH (a)-[r]-(a)``) — per-clause ``_alias_target`` cannot
+    # represent them; the general lowering serves these (TCK match7-24).
```
lmeyerov: So merging has tck-gfql TODOs? Likely canvas for more etc

graphistry/compute/gfql_unified.py
```diff
@@ -1443,11 +1466,58 @@ def _execute_compiled_query_with_reentry(
...
             empty_result_row=compiled_query.empty_result_row,
             reentry_plan=compiled_query.reentry_plan,
+            aggregate_fill_values=_optional_reentry_aggregate_fill_values(compiled_query),
         )
 
     return result
 
 
+def _optional_reentry_aggregate_fill_values(compiled_query: CompiledCypherQuery) -> Dict[str, Any]:
+    """Cypher empty-group values for aggregate outputs of an optional-reentry suffix.
```
lmeyerov: Narrow Any

graphistry/compute/gfql_unified.py
```diff
@@ -1959,7 +2029,7 @@ def gfql(self: Plottable,
          policy: Optional[Dict[str, PolicyFunction]] = None,
          where: Optional[Sequence[WhereComparison]] = None,
          language: Optional[Literal["cypher", "gremlin"]] = None,
-         params: Optional[Mapping[str, Any]] = None,
+         params: Optional[Mapping[str, Any]] = None,  # hygiene-ok: explicit-any -- Cypher query parameters are arbitrary user scalars
```
lmeyerov: Any again

lmeyerov:
Answering the two process comments; the seven code points are being implemented on this branch and I'll report per-point when they land.

**"We should queue refactoring lowering.py … 9K loc smells bad"** → filed as #1921. Current state is **9,260 lines / 217 top-level defs**, and the surface guard has ratcheted its baseline four times during this campaign alone (9260 → 9772 → 9808 → 9891), each with a written justification — the guard is working as a ratchet, but a ratchet isn't a structure. I put the *evidence* in the issue rather than the aesthetics: bugs in this file were repeatedly hidden by adjacency (the #1915 B-1a `except TypeError: → False` directly contradicted a comment ~200 lines above it in the same module), and fixes kept needing several sites here plus the polars twin, which is how sites got missed (#1888 needed three kernels, #1911 three). Two modules were already extracted successfully during this campaign — `cypher/aggregate_identity.py` and `cypher/identifiers.py` — so the seams are real. Explicitly scoped as post-merge: restructuring this file with a dozen PRs stacked on it would be reckless.

**"So merging has tck-gfql TODOs? Likely canvas for more etc"** → those particular strings are *provenance*, not TODOs. `(TCK match7-24)` and `(TCK match7-13)` annotate **why a decline exists** — the shape is served by the general lowering because that scenario requires it — so they document a deliberate routing choice rather than outstanding work. There is no `TODO`/`FIXME` anywhere in `cypher/` (verified: zero).

But the underlying question is right, and there *is* real cross-repo work in flight worth canvassing. This stack has been promoting openCypher TCK scenarios as fixes make them pass, and each promotion is a companion commit in **tck-gfql** that must land alongside:

| scenario(s) | promoted by | why |
|---|---|---|
| `match-where6-2`, `match7-29/30/31` | #1894 | OPTIONAL MATCH null-extension now served |
| `match5-14/15`, `usecase-countingsubgraphmatches1-2` | #1904 | trail semantics served |
| `expr-comparison2-6-3/6-4` | #1919 | incomparable comparison is null, not false |

Each carries contract-key moves, a regenerated capability-debt manifest, and hand-pinned count updates. They are pushed as **same-named companion branches** in tck-gfql (CI resolves the branch matching the PR's head), so the tck-gfql side must merge in lockstep with this stack or the `tck-gfql` lane fails. Two traps I've already hit and recorded: `~/Work/tck-gfql` is a *different* checkout that lacks these companions (it shows ~10 false failures), and the wrong-row debt bucket is now empty, which flipped several report assertions from `in` to `not in`.

Worth noting the promotions are independent confirmation rather than self-marking: the #1919 pair flipped from `success_wrong_rows` to `success_matches_expected` against a corpus nobody was optimizing toward.

graphistry/compute/gfql/query_types.py
```diff
@@ -6,6 +6,10 @@
 from graphistry.compute.ast import ASTLet, ASTObject
 from graphistry.compute.chain import Chain
 
+# NOTE: the companion `params=` alias is `graphistry.compute.gfql.cypher.ast.CypherParams`.
+# It is deliberately NOT re-exported here: this module is imported by
+# `graphistry.compute.__init__`, and reaching into the `cypher` package would drag the whole
+# Cypher compiler into every `import graphistry`.
```
lmeyerov: comment blongs in CypherParams not here

graphistry/compute/gfql_unified.py
```diff
@@ -510,6 +521,19 @@ def _optional_arm_start_nodes(
                     opt_rows_df = opt_rows_df.join(join_keys, on=join_cols, how="inner")
                 joined = joined.join(opt_rows_df.select(opt_only_cols), on=join_cols, how="left")
             else:
+                # Zero-match (or unjoinable) arm: null-extend the FULL arm
+                # schema with typed nulls, not just the bare alias columns.
+                # The rows op emits the arm's binding schema even at 0 rows,
+                # so property refs (x.v) and downstream join keys (x.id) keep
+                # real dtypes -- the outcome must not depend on whether the
+                # arm happened to match (#1891 F-03).
```
lmeyerov: this kind of comment likely belongs in code & tests, audit & fix style violatons

graphistry/compute/gfql_unified.py
```diff
@@ -1443,11 +1472,58 @@ def _execute_compiled_query_with_reentry(
...
+    """
+    plan = compiled_query.reentry_plan
+    carried = set(plan.scalar_columns) if plan is not None else set()
+    reentry_alias = plan.reentry_alias_name if plan is not None else None
+    ops = list(compiled_query.chain.chain) if compiled_query.chain is not None else []
+    with_map: Dict[str, object] = {}
+    fills: Dict[str, CypherEmptyGroupValue] = {}
+    for op in ops:
+        function = getattr(op, "function", None)
+        params = getattr(op, "params", None) or {}
```
lmeyerov: prefer static typing patterns over getattr's

graphistry/compute/gfql_unified.py
```diff
@@ -1443,11 +1472,58 @@ def _execute_compiled_query_with_reentry(
...
+                func = str(agg[1]).lower()
+                expr = agg[2] if len(agg) == 3 else None
+                if func == "count" and (expr is None or expr == "*"):
+                    fills[alias] = 1
+                    continue
+                source: object = with_map.get(expr, expr) if isinstance(expr, str) else expr
+                if not isinstance(source, str):
+                    continue
+                base = source.split(".", 1)[0]
+                if "__cypher_reentry_" in source or base in carried or base == reentry_alias:
```
lmeyerov: audit & fix externing of special symbol names like this..

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -8968,6 +9088,24 @@ def _attach_graph_context(result: CompiledCypherQuery) -> CompiledCypherQuery:
             # ``reentry_matches=()``, so the recursive call cannot re-enter
             # this branch — recursion terminates after one step.
             return compile_cypher_query(flattened, params=params)
+        # #1891: a pure-carry WITH before OPTIONAL MATCH is a scope-preserving
+        # no-op; flatten it onto the connected optional-match left-join
+        # lowering (which null-extends correctly) instead of the reentry row
+        # assembly (which loses carried seed properties for unmatched rows).
+        # Only flatten when the result actually takes that lowering.
```
lmeyerov: should this be more of a test than a comment?

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -9125,11 +9263,49 @@ def _lower_general() -> CompiledCypherQuery:
...
+        )
+        # Shapes the connected optional-match left-join lowering declined or
+        # deferred (e.g. variable-length optional arms, grouped post-aggregate
+        # expressions) land here, where the aggregate lowering has no
+        # null-extension: it inner-joins before aggregating. That is WRONG
+        # exactly when unmatched seeds must still contribute -- grouped
+        # shapes (their group rows vanish) and count(*) (the null-extended
+        # row itself counts). Ungrouped null-skipping aggregates are
+        # inner-join-safe (nulls contribute nothing), so they stay served
+        # (issue #1367 empty-row synthesis). Decline the wrong ones honestly.
```
lmeyerov: should this be more of a test vs a comment?

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -9228,6 +9391,37 @@ def _lower_general() -> CompiledCypherQuery:
...
+                # the row guard serves them and declines on lost multiplicity.
+                optional_projection_row_guard = _projection._optional_projection_row_guard_plan(
+                    query,
+                    params=params,
+                )
+            # Honest residual gate: shapes with an optional clause that the
+            # connected optional-match left-join lowering declined (e.g.
+            # variable-length optional arms) and that no null-extension
+            # mechanism above covers would silently inner-join here, dropping
+            # unmatched rows. Decline honestly instead of answering wrong.
```
lmeyerov: same, unclear why all these are comments vs tests, clear code & names, etc

graphistry/compute/gfql/agg_types.py
```diff
@@ -43,20 +43,33 @@
...
+from typing import TYPE_CHECKING, Final, FrozenSet, List, Mapping, NoReturn, Optional, Union
 
 from graphistry.compute.exceptions import ErrorCode, GFQLTypeError
 
 if TYPE_CHECKING:
     import polars as pl
 
+    # TYPE_CHECKING only: `gfql.cypher.__init__` eagerly imports the whole compiler, and
+    # this module is imported from inside that import (row.pipeline), so a runtime import
+    # would be circular. The aliases below only need the name at type-check time.
```
lmeyerov: unnecessary comments, esp here? audit and fix, incl. root cause on why you are feeling entitled to do these and why your review flow isn't avoiding them and adjacent issues

graphistry/compute/gfql/agg_types.py
```diff
@@ -43,20 +43,33 @@
...
 
 
 #: Aggregates Cypher restricts to ``INTEGER | FLOAT | DURATION``. ``mean`` is GFQL's internal
 #: spelling of ``avg`` (see ``GFQL_GROUPBY_AGG_METHODS``), so both names must gate.
 GFQL_NUMERIC_ONLY_AGGREGATIONS: Final[FrozenSet[str]] = frozenset({"sum", "avg", "mean"})
 
+#: What an aggregate answers for an EMPTY group, per the Cypher contract documented above:
+#: ``count``/``sum`` -> ``0``, ``count(*)`` over a synthesized null-extended row -> ``1``,
+#: ``collect`` -> ``[]``. Every other aggregate's empty-group answer is ``null`` and is
+#: therefore absent from a fill map rather than present as a value.
```
lmeyerov: this comment can be eliminated or at least drastically shortened per our rules, fix this + root cause why it happened & got through

graphistry/compute/gfql_unified.py
```diff
@@ -152,11 +154,9 @@ def _apply_optional_null_fill(
...
     rows_df = result._nodes
     actual_rows = 0 if rows_df is None else len(rows_df)
-    # The null-fill alignment machinery below (matched-id meta, .iloc row slicing,
-    # per-segment concat) is not yet native on polars: the polars OPTIONAL MATCH
-    # does not populate the matched-seed `_cypher_entity_projection_meta["ids"]`
-    # this path needs. Decline honestly (NO-CHEATING) rather than raising a
-    # misleading "unsupported-cypher-query" validation error — pandas handles it.
+    # polars serves this natively only when its projector recorded the matched-seed
+    # `_cypher_entity_projection_meta["ids"]`; otherwise decline honestly
+    # (NO-CHEATING) rather than raise a misleading validation error.
```
lmeyerov: this comment can be eliminated or at least drastically shortened per our rules, fix this + root cause why it happened & got through

graphistry/compute/gfql_unified.py
```diff
@@ -172,7 +172,11 @@ def _apply_optional_null_fill(
         message="Cypher OPTIONAL MATCH null-row alignment could not recover matched seed identities",
         suggestion="Use a simpler OPTIONAL MATCH projection shape in the local compiler.",
     )["ids"]
-    if not hasattr(matched_ids, "tolist"):
+    # `is_series_like` (pandas/cuDF `.dropna` or a polars Series) rather than sniffing
+    # `tolist`/`to_list`: cuDF has BOTH and both RAISE by design (host-transfer guard),
+    # so attribute presence never proved this was usable. `series_to_pylist` below is
+    # the only thing that reads the values, and it is engine-aware.
```
lmeyerov: this comment can be eliminated or at least drastically shortened per our rules, fix this + root cause why it happened & got through

graphistry/compute/gfql_unified.py
```diff
@@ -209,7 +213,10 @@ def _apply_optional_null_fill(
     concrete_engine = resolve_engine(cast(Any, engine), result)
     df_ctor = df_cons(concrete_engine)
     concat = df_concat(concrete_engine)
-    fill_df = df_ctor({key: [value] for key, value in null_row.items()})
+    # Fill rows span the projected frame's columns (#1650 flattened form:
+    # a whole-entity miss is all-null across its {alias}.{field} columns).
```
lmeyerov: this comment can be eliminated or at least drastically shortened per our rules, fix this + root cause why it happened & got through

graphistry/compute/gfql_unified.py
```diff
@@ -533,6 +557,13 @@ def _optional_arm_start_nodes(
                 opt_rows_df = opt_rows_df.merge(join_keys, on=join_cols, how="inner")
             joined = joined.merge(opt_rows_df[opt_only_cols], on=join_cols, how="left")
         else:
+            # Zero-match (or unjoinable) arm: null-extend the full arm schema
+            # (see the polars twin above) so property refs and downstream join
+            # keys exist regardless of whether the arm matched (#1891 F-03).
```
lmeyerov: this comment can be eliminated or at least drastically shortened per our rules, fix this + root cause why it happened & got through

lmeyerov:
Round 2 pushed as `bd215c4a3`. **Per-thread mapping below so you can verify and resolve — I am not marking anything resolved myself.** Current state by the only measure that counts: **18 unresolved / 22 threads** before this push.

## Two findings that change what you should look at

**1. The `query_types.py` NOTE was FALSE, so I deleted it rather than relocating it.** You said *"comment belongs in CypherParams not here"* and I briefed the agent to move it. It probed the claim first: the note said re-exporting `CypherParams` would drag the Cypher compiler into every `import graphistry`. **Both halves are false** — `gfql_unified.py:38` already imports `cypher.lowering` at import time via `ComputeMixin`, and adding the re-export raises no cycle. Moving it would have laundered a false claim into a more authoritative location. My instruction was wrong; the agent checked instead of complying.

**2. Replacing a comment with a pin uncovered a silent cross-engine wrong answer.** A deleted comment in `gfql_fast_paths.py` asserted *"the polars lanes above resolve or decline the collision themselves."* They do not. `RETURN c.city AS s` on a graph whose src column is `s` returns **the edge src id** on polars (`{'s': 0, 'n': 2}`) while pandas answers correctly (`{'s':'LA','n':1},{'s':'NY','n':1}`). Present on this branch **and on its base**. The guard sat inside the eager branch only; it is now hoisted above the engine split (`gfql_fast_paths.py:2134`). `test_grouped_aggregate_fused_polars.py` could not see it — it compared polars-eager against polars-eager.

## Your comments → what changed

| Your comment | Fix |
|---|---|
| `query_types.py:12` "belongs in CypherParams" | **deleted as false** (above) |
| `gfql_unified.py:529` "belongs in code & tests, audit & fix" | 3 extracted helpers (`_semi_join_prune_arm_rows_to_base_keys`, `_null_extend_full_arm_binding_schema`, `_synthesize_bare_alias_from_prefixed_column`); arm loop 75 lines/5 comment blocks → 11 lines/none; pin `test_optional_zero_arm_null_extends_full_arm_schema_not_just_bare_aliases` |
| `gfql_unified.py:1498` "prefer static typing over getattr" | `isinstance(op, ASTCall)` at `:1497`, `:1200`, `:666`; the last also dropped a `cast(str, getattr(...))` |
| `gfql_unified.py:1518` "audit & fix externing" | 7 literals → 6 constants in **existing** homes (`reentry/naming.py`, `identifiers.py`, `agg_types.py`); no third home |
| `lowering.py:9095/9285/9410` "more of a test than a comment?" | named predicates (`_aggregate_shape_needs_null_extension_this_lowering_lacks`, `optional_arm_has_no_null_extension_mechanism`, `served_by_general_optional_null_fill`) or deletion backed by existing pins |
| `agg_types.py:55` + `:67` | TYPE_CHECKING note deleted; every multi-line `#:` block reduced to one line |
| `gfql_unified.py:159/178/217/562` | named helpers / named booleans / deleted where a pin carries it |

**Audit A (style): 38 sites** — 7 extracted helpers, 7 named booleans, 13 deletions where a pin already carries the meaning, 5 docstrings cut to one line. **Audit B (symbols): 7 literal sites → 6 constants.** Both audits list what was **deferred** as pre-existing repo debt off this PR's diff, rather than sprawling — the deferred symbol list alone is ~30 sites in `lowering.py`/`fast_paths.py`.

## Two caveats I want you to see, not bury

1. **The guard hoist is a behavior change**: polars now *declines* a shape it previously served — it was serving it **wrong**, and both engines now agree, but a fast-path shape was lost and the fused-polars test file's classification changed with it.
2. The local 93-failure set is a GPU-less-box baseline, byte-identical base vs head. It is not a CI claim.

Hygiene ratcheted **down** (explicit-cast `gfql_unified.py` 41→40; total 4543→4542), not merely held flat.

lmeyerov:
Round-3 pass over all 12 open threads. Head is now `914ea50e3`.

Two prior commits are involved, and I want to be exact about which is which:
`bd215c4a3` is the **round-2** commit that landed before this pass — several threads
were already discharged there and I only **verified** them. `c4015855b` and
`914ea50e3` are **this** pass. Every thread below says which.

Every thread was re-checked against the CURRENT head, not against the `outdated`
flag — outdated only means the line moved.

---

**1. `gfql_unified.py:173` — "Avoidable dynamic typing"**

Partly `bd215c4a3`, finished in `c4015855b`. The flagged
`hasattr(matched_ids, "tolist")` became `is_series_like(matched_ids)` in round 2.
But round 2 *introduced a fresh instance of the same smell* 60 lines down —
`rows_df.iloc[...] if hasattr(rows_df, "iloc") else rows_df.slice(...)`. That is
now `_slice_rows(rows_df, start, stop)`, dispatching on `is_polars_df`, typed
`DataFrameT -> DataFrameT`. No `Any`, no `cast`, no `hygiene-ok`.
`grep -n 'hasattr' graphistry/compute/gfql_unified.py` → no engine probes left.

**2. `cypher/reentry/execution.py:88` — "These dynamic typings all smell"**

Already fixed in `bd215c4a3`, verified, not my work. `reentry_validation_error`
went from `value: Any` + `**extra_context: Any` to `value: object` plus a closed
kwarg set (`reason`, `carried_row_count`, `carried_scalar_columns`) — the callers
dispatch on `reason`, so the set really is closed. Hygiene count for that file
20 → 8 in the baseline diff.

**3. `gfql_unified.py:2032` — "Any again"**

Partly `bd215c4a3`, finished in `c4015855b`. Round 2 narrowed the *public*
`gfql(params=...)` to `CypherParams` and dropped the `hygiene-ok` marker — but
left six **internal** signatures still on `Optional[Mapping[str, Any]]` carrying
the very same value (`_gfql_otel_attrs`, `_compile_cache_params_key`,
`_compile_string_query`, `_compile_summary`, `_fire_postcompile_policy`,
`_auto_cudf_polars_gpu_route`). All six are now `Optional[CypherParams]`.
explicit-any in `gfql_unified.py`: **20 → 14**, locked into the baseline
(tightened, never raised).

Also in `c4015855b`: `lowering.py:7690` had
`(el.to_fixed_point if hasattr(el, "to_fixed_point") else False)` on a
`RelationshipPattern`. `to_fixed_point` is a dataclass field with a default, so
that `hasattr` was always `True` — dead dynamic typing. Now `el.to_fixed_point`,
matching the seven other call sites in that file that already read it directly.

**4. `gfql_unified.py:529` — "belongs in code & tests, audit & fix style violations"**

Already fixed in `bd215c4a3`, verified. The 6-line block comment is gone; the code
is now `_null_extend_full_arm_binding_schema(joined, opt_rows_df, opt_only_aliases)`.
The polars and pandas twins that each carried their own copy of that comment
(this thread and #9) collapsed into that one helper plus
`_semi_join_prune_arm_rows_to_base_keys` and
`_synthesize_bare_alias_from_prefixed_column`.

**5. `lowering.py:9095` — "should this be more of a test than a comment?"**

Already fixed in `bd215c4a3`, verified. Both comments there are deleted. The
"#1891 pure-carry WITH is a scope-preserving no-op" narration is now the function
name `flatten_pure_carry_optional_reentry`. The "recursion terminates after one
step" claim is now a test, not a comment:
`test_flatten_pure_carry_optional.py::test_rebind_result_has_no_reentry_matches_so_recompilation_terminates`,
one of 21 tests in that new file.

**6. `lowering.py:9285` — "should this be more of a test vs a comment?"**

Already fixed in `bd215c4a3`, verified. The 8-line comment is now the function
`_aggregate_shape_needs_null_extension_this_lowering_lacks(...)` with a one-line
docstring, called at the decline site.

**7. `lowering.py:9410` — "unclear why all these are comments vs tests"**

Already fixed in `bd215c4a3`, verified. Both comments gone; the residual gate is
now the named boolean `optional_arm_has_no_null_extension_mechanism`, and the
"bound-alias-only projections are arm-independent" comment is carried by
`_return_references_only_bound_aliases(...)`.

**8 / 9. `gfql_unified.py:217` and `:562` — "eliminate or drastically shorten + root cause"**

Already fixed in `bd215c4a3`, verified. `:217` is the named local
`fill_columns_spanning_projected_frame`. `:562` folded into the shared helper from
thread #4 (it was the pandas twin of the same comment).

On the **root cause** you asked for: these were pandas/polars twin branches. The
same fact got written twice because the *code* was duplicated, and a comment is
the cheapest way to keep two copies in sync. Unifying the branches removed both
the duplication and the reason to comment. The style rule caught the symptom; the
duplication was the cause.

I also swept the rest of this PR's added lines for the same class rather than
only the three you pointed at. Two more in `c4015855b`:
- a 3-line comment explaining `structured_projection` → renamed to
  `row_guard_needs_single_column_entity_text`, comment deleted;
- `_optional_reentry_aggregate_fill_values` had a 6-line docstring asserting a
  contract (count→0, sum→0, collect→[], count(\*)→1, carried sources stay NULL).
  That is a test, so it is now a test — see below — and the docstring is one line,
  with the NULL condition carried by the extracted predicate
  `source_is_carried_rather_than_suffix_bound`.

The new pins are `test_optional_match_semantics.py` section E (6 cases). They are
**compile-level on purpose**: the existing end-to-end sweep `return`s on a typed
gate or `NotImplementedError`, so it can pass vacuously and could not have carried
this contract. **I mutation-tested them** — (a) deleting the `count(*) → 1` branch
fails `count_star_counts_the_null_extended_row`; (b) deleting the carried-source
guard fails `collect_over_carried_scalar`; (c) swapping the 0 fill for `[]` fails
`count_suffix_alias_is_zero` and `sum_suffix_property_is_zero`. All reverted; 6/6
green at head.

**10. `query_types.py:12` — "comment belongs in CypherParams not here"**

Deleted in `bd215c4a3`, **not relocated**, and I agree with that call after
re-checking. The NOTE claimed re-exporting from the `cypher` package "would drag
the whole Cypher compiler into every `import graphistry`". That is false at head:
`graphistry/compute/__init__.py:62` imports `query_types`, and
`gfql_unified.py:40,42` already do `from graphistry.compute.gfql.cypher.parser
import parse_cypher` / `...cypher.lowering import ...` at module import time. The
compiler is already imported. Moving a false statement into `CypherParams` would
have preserved a wrong claim in a more authoritative place, so it went away
instead. `sed -n '1,12p' graphistry/compute/gfql/query_types.py` — no NOTE.

**11. `projection.py:96` — "Should this set be externed DRY?"**

Yes. Started in `bd215c4a3`, **finished in `914ea50e3`.** Round 2 moved the flagged
`{"id","labels","type"}` and `"label__"` into
`gfql/row/entity_props.py` as `NODE_INTERNAL_COLS`, `LABEL_FLAG_PREFIX`, and
`label_flag_columns()`, imported by `projection.py`.

But that only fixed the site you pointed at — I grepped wider and **four bare
`"label__"` literals were still out there** in files this PR hadn't touched:
`frontends/cypher/binder.py` (`_catalog_node_labels`, `_catalog_edge_types`) and
`row/pipeline.py` (entity-shape probe, label `filter_dict` probe). Half-externing
a constant is worse than not externing it, so `914ea50e3` points all four at
`LABEL_FLAG_PREFIX`. `grep -rn '"label__"' graphistry/compute/gfql/` now matches
**only** the definition in `entity_props.py`.

**12. `ci_cypher_surface_guard_baseline.json:16` — queue the lowering.py refactor**

Already tracked: **#1921** ("Refactor lowering.py: 9,260 lines / 217 top-level
defs in one module"), open, and it quotes this exact comment as its motivation. It
also explicitly folds in `gfql_unified.py` per your #1897 note, and is marked
post-release. **No refactor in this PR** — twelve-plus PRs are stacked on this file.

One thing I did do: the baseline was carrying slack. Actual `lowering.py` is
**9,425** lines against a baseline of 9,454, so the ratchet had 29 lines of free
headroom. `c4015855b` tightens it to 9,425.

---

**Gates at `914ea50e3`**

- `ruff check graphistry bin` — clean
- `ci_type_hygiene_guard.py` — OK, no growth (baseline *lowered*: `gfql_unified.py` explicit-any 20 → 14)
- `ci_cypher_surface_guard.py` — pass (`lowering_py_max_lines` 9454 → **9425**)
- `mypy graphistry` — 4 errors, byte-identical to the pre-existing set in `polars/hop_eager.py` and `polars/degrees.py` (local polars stub skew). No new errors.
- `pytest graphistry/tests/compute` — 102 FAILED, **`diff` against the base-commit failure set is empty**. Same 102 cudf/optional-dep/env cases, nothing added, nothing masked. Re-run in full after the `914ea50e3` binder/pipeline change, still identical.
- No new `cast(`, no new `Any`, no new `hygiene-ok` markers anywhere in this pass.

Leaving all 12 threads unresolved for you to close.

Issue #1808, opened 2026-07-28
GFQL: a DANGLING DESTINATION is matched by pandas/cuDF and dropped by polars — silent cross-engine divergence


lmeyerov:
## Minimal repro

```python
import pandas as pd, polars as pl, graphistry
nodes = pd.DataFrame({"id": [0, 1, 2]})
q = "MATCH (a)-[]->(b) RETURN count(*) AS c"

# edge (1 -> 9): destination 9 is NOT in the node table
edges = pd.DataFrame({"s": [0, 1], "d": [1, 9]})
```

| case | edges | pandas | cuDF | polars |
|---|---|---|---|---|
| dangling **source** `(9 -> 2)` | 2 | 1 | 1 | 1 |
| dangling **destination** `(1 -> 9)` | 2 | **2** | **2** | 1 |
| both | 3 | **2** | **2** | 1 |

A dangling **source** is dropped by all three engines. A dangling **destination** is dropped only by polars — pandas and cuDF match it and bind `b` to an endpoint that is not a node.

## Which is right

`MATCH (a)-[]->(b)` binds `b` to a **node**. If the destination id has no row in the node table there is nothing to bind, so the pattern must not match. **polars is correct; pandas and cuDF are wrong** — and pandas is this repo's oracle, so the cross-engine parity suites are calibrated against the wrong answer for this shape.

It is also internally inconsistent: the same engine drops a dangling source and keeps a dangling destination, so it is an asymmetry in the endpoint gate rather than a deliberate "edges are authoritative" policy.

## Scope

No error, no warning — just a different count. Found while auditing merged PRs #1782/#1783/#1784/#1790/#1792/#1793 with a pandas-vs-polars differential harness; it reproduces **identically on `84be35fb` (before that stack) and on `233b64c8`**, so it is **pre-existing and not caused by any of them**. It reproduces at every graph size tried and on every seed.

## Why it has not been caught

The polars parity suites compare against the pandas oracle, so a shape where pandas is the wrong one shows up as "polars diverges" and is easy to read as a polars gap. Nothing in the corpus builds a graph with a dangling destination on purpose.

## Suggested fix direction

Make the pandas/cuDF endpoint gate symmetric — require BOTH endpoints to resolve to node rows — then re-baseline any parity fixture that silently encoded the old count. Worth pinning with an explicit dangling-endpoint fixture on all four engines rather than relying on random fuzz to regenerate it.

lmeyerov:
[...]

Issue #1824, opened 2026-07-28
GFQL: Cypher OLAP fast paths silently run on CPU under engine='polars-gpu' (bare LazyFrame.collect() bypasses the GPU-or-error contract)


lmeyerov:
## Summary

`engine='polars-gpu'` **silently runs the Cypher OLAP fast paths on CPU** and reports them as GPU. This violates the NO-CHEATING contract the tree states in its own words.

`graphistry/compute/gfql/lazy/__init__.py:186-198`:

> ``raise_on_fail=True`` is the NO-CHEATING contract for the GPU target: any non-GPU-executable plan node RAISES — **never silently run on CPU and reported as GPU.** […] hard raise forces the truth: **polars-gpu is GPU-or-error**, use polars for CPU.

The fast paths never reach that contract.

## Mechanism (read from `origin/master` @ `1537e4676`)

1. `graphistry/Engine.py:33` — `POLARS_ENGINES = (Engine.POLARS, Engine.POLARS_GPU)`.
2. `gfql_fast_paths.py:1688/1868/2036` — `requested_engine = resolve_engine(...)`; `'polars-gpu'` resolves to `Engine.POLARS_GPU` and is **not** downgraded.
3. `gfql_fast_paths.py:1070` (and `:1449`, `:1519`, `:1713`, `:1907`) — the branch gate is `if requested_engine in POLARS_ENGINES:`, so **POLARS_GPU enters the polars branch**.
4. `gfql_fast_paths.py:871,878` (in `_connected_join_two_star_fused_polars`) — the plan is collected with a **bare** `joined_lf.collect()` / `left_counts_lf.collect()`.

A bare `.collect()` takes no `pl.GPUEngine`, so `ExecutionTarget`, `_engine_for()`, `raise_on_fail=True`, `gpu_executor()` and `cpu_streaming()` are **all bypassed**. The work runs on CPU polars under a GPU engine label.

## Measured (dgx-spark, `graphistry/test-rapids-official:26.02-gfql-polars`, `--gpus all`, polars 1.35.2 in-image, graph-benchmark 100k)

| query | `engine='polars'` | `engine='polars-gpu'` | fast path |
|---|---|---|---|
| q1 | 39.82 ms | 33.77 ms | fused 7/7 |
| q4 | 9.98 ms | 9.11 ms | fused 7/7 |
| q9 | 45.95 ms | 38.04 ms | fused 7/7 |
[...]

Issue #1739, opened 2026-07-20
GFQL polars: HAS_<Label> aggregation shape diverges from pandas on duplicate-id graphs (no disambiguation, no decline)


lmeyerov:
Found by the #1737 review's test-coverage adversarial pass; **pre-existing on master `dbc6f628`** (A/B-verified — NOT introduced by #1737, which covers only the `binding_rows_polars` path).

## Repro

Fixture (same as `_disambiguation_frames()` in `graphistry/tests/compute/gfql/test_polars_rows_entity_groupby.py`): node id 400 duplicated across a Tag row (`t4`) and a Forum row (`f4`); edge `601 -HAS_TAG-> 400`.

```cypher
MATCH (post:Post {id: 601})-[:HAS_TAG]->(tag)
RETURN tag.name AS tagName, count(post) AS c
```

- pandas: `[{'tagName': 't4', 'c': 1}]` (disambiguation narrows the unlabeled `tag` op to the HAS_TAG label row)
- polars: `[{'tagName': 'f4', 'c': 1}, {'tagName': 't4', 'c': 1}]`

Silent divergence — violates the polars parity-or-NIE contract. The single-MATCH **aggregating** shape routes through a polars path that has no HAS_<Label> destination disambiguation and never declined. The non-aggregating RETURN shape on the same fixture is parity-correct (covered by `test_has_label_narrowing_applies_on_reached_collision`).

## Fix options

1. Port the pandas `_gfql_disambiguate_has_edge_destination_nodes` gate to this path (probing pre-dedup base nodes — see #1737's `binding_rows_polars` gate for the base-graph plumbing via `_gfql_rows_base_graph`), or
2. honest NIE decline for the gate shape on duplicate-id graphs (the #1737 resolution).

Either way, add a parity test pinning this exact aggregating shape.

lmeyerov:
[...]

PR #1895, opened 2026-08-14
fix(gfql)!: endpoint-closure contract on every surface (#1888, absorbs #1808)

lmeyerov:
Umbrella 3 of 3, stacked on #1894. One rule — bound node table ⇒ both endpoints resolve — enforced at the three execution kernels; five surfaces converged to the same closed answer, policy-invariant, dtype-preserving. 3 strict-xfail pins flipped (#1888 pair + #1808), 6 closure-invariant greens, ZERO re-baselines needed, failure set byte-identical to baseline. NEEDS-GPU spot-check of the six pin shapes on cudf/polars-gpu queued on dgx. The #1894-noted OM whole-row-projection regression fix follows as the next commit on this stack.


lmeyerov:
**GPU spot-check receipt (dgx GB10, RAPIDS 26.02 container, `--gpus all`, clone verified @ 5ddbd666d):**

- `test_known_cross_engine_divergences.py`: 11 passed / 3 xfailed (pre-existing #1739, #1824, BUG-4 pins) / 0 failed — all #1888/#1808 closure pins green.
- `test_hop_semantics_pins.py` + `test_optional_match_semantics.py`: 92 passed / 0 failed.
- Inline dangling-endpoint fixture, **cudf** (pandas-path kernel): chain=1 edge, nodes [0,1], v stays int64, no nulls; cypher `RETURN a, b` = 1 row; property projection `{av:10, bv:20}`; `count(*)`=1 — ALL PASS.
- Same fixture, **polars-gpu** (`GPUEngine(raise_on_fail=True)` sanity-probed first): chain=1 edge, nodes [0,1], Int64, null_count=0; property projection + count agree — ALL PASS. Whole-entity `RETURN a, b` raises the *documented* polars-projector NIE (separate gap, pinned as such on this branch).
- Known #1891-rework regression `test_string_cypher_formats_optional_match_projection_on_cudf` still fails with the expected residual-gate signature (fix in flight on this branch).

NEEDS-GPU item from the PR description is discharged. Provenance artifacts on dgx: `~/pyg-1888-spotcheck` (HEAD 5ddbd666d), `~/step3_1888_gpu_fixture.py`.

lmeyerov:
Stack tip updated (`f0c995f75`): the two #1894-rework regression fixes were cherry-picked down to their home PR #1894 and merged back up (identical content, clean ort merge). Also on this branch: type-hygiene guard's explicit-cast matcher now flags only module-qualified `typing.cast` — `expr.cast(dtype)` dataframe dtype casts were being miscounted (released 50 false grandfathered findings, 11 files ratcheted down). Local: OM 60/60 (on tip), hop pins 42/42, divergence pins 11P/3xf, guards green, ruff clean. tck-gfql companion branch `fix/gfql-1888-endpoint-closure` pushed (same 4-key contract update).

graphistry/compute/gfql/lazy/engine/polars/chain.py
```diff
@@ -1006,19 +1006,30 @@ def _plain_edge(op):
...
-                    from_ids = filter_by_dict_polars(gf._nodes, n0.filter_dict).select(pl.col(ncol))
-                    edges = edges.join(from_ids, left_on=from_col, right_on=ncol, how="semi")
-                if n2.filter_dict:
-                    to_ids = filter_by_dict_polars(gf._nodes, n2.filter_dict).select(pl.col(ncol))
-                    edges = edges.join(to_ids, left_on=to_col, right_on=ncol, how="semi")
+            # #1888 endpoint closure: with a node table BOUND, an edge matches only if
+            # BOTH endpoints resolve to node rows — one semi-join per endpoint against
+            # its filtered (or full) node universe, parity with the pandas chain's
+            # endpoint prune. Synthesized-from-edges node tables are vacuously closed:
+            # keep the zero-join fast path there.
```
lmeyerov: This pr violates our rules about avoiding comments when naming, tests, etc can obviate them

graphistry/compute/gfql/lazy/engine/polars/hop_eager.py
```diff
@@ -202,6 +202,23 @@ def hop_polars(
 
     FROM, TO, NID, EID, edges_idx, synth_eid, node_dtype = _hop_setup_columns(
         edges, all_nodes, node_col, g._edge)
+
+    # #1888 endpoint closure: with a node table BOUND, an edge can match only if BOTH
+    # endpoints resolve to node rows (Cypher: `(a)-[]->(b)` binds nodes at both ends).
+    # One O(E) pass, parity with the pandas hop's symmetric gate. Nodes synthesized
```
lmeyerov: May violate rules about preferring perf tracking in pyg-bench , audit this + stack and fix

graphistry/compute/hop.py
```diff
@@ -984,7 +995,12 @@ def _undirected_cycle_nodes(edges_df: DataFrameT) -> set:
 
         g_out = g_out.nodes(final_nodes)
 
-    # Ensure all edge endpoints are present in nodes.
+    # Ensure all edge endpoints are present in nodes. Only MISSING endpoints are
+    # appended (#1888): with endpoint closure the edge frame cannot reference an id
+    # outside a bound node table, so for the common case this is empty and the old
+    # unconditional concat — which synthesized NaN-attribute rows and upcast attr
+    # dtypes (e.g. int64 -> float64) even when nothing was missing — never runs.
+    # Wavefront-mode outputs whose node frame omits an endpoint side still backfill.
```
lmeyerov: Another comment violation, review skill clearly failed or  not used

lmeyerov:
Review addressed, plus the stack-wide audit — and the audit surfaced a **perf finding that may matter more than the review**.

## The three comments
| # | Site | Fix |
|---|---|---|
| 1 | `polars/chain.py:1019` | 5-line closure narration deleted; rule now in names — `node_table_bound`, `_ids_endpoint_must_resolve_to()`, `resolvable_ids` — and the two commented branches became a loop over `(endpoint_col, resolvable_ids)` |
| 2 | `polars/hop_eager.py:208` | Gate extracted to `_keep_edges_with_both_endpoints_resolvable()`; the "one O(E) pass" claim deleted, along with an adjacent *"measurably regressed large-edge runs — A/B twice"* note and two `#1892 F-01 …for cost` notes → predicate `only_seeds_can_be_sources` |
| 3 | `hop.py:1003` | 6-line backfill narration → `_endpoint_ids_without_node_rows()` plus an unconditional trailing dedup, which structurally removes the need for the "preserve the historical dedup side effect" note |

## Stack audit (25215ce5c..HEAD — 61 commits, 173 added comment lines)
13 sites fixed across 12 files, **net −43 comment lines**. **Every performance claim in the stack was removed** (hop_eager's A/B note, two `#1892 F-01` cost notes, hop.py's *"O(E) itertuples walks; vectorize if hot"*, `gfql_fast_paths`' lazy-plan note, `partitioned_layout`'s *"Vectorized normalize"*, chain.py's *"every hop falls back to the O(E) scan"*). Kept, with justification, are genuine inexpressible constraints: the polars int/float join-key trap, two import-cycle notes, the strong-ref-not-`id()` note, two pandas-classifier traps, openCypher scoping notes. Two pre-stack blocks left alone and flagged as follow-ups.

## Amplified matrix
New `test_endpoint_closure_matrix.py` — 16 functions, **100 collected cases** across pandas/polars/cudf/polars-gpu. Positive: chain 1-hop/2-hop/var-length, hop in all three directions seeded and unseeded (6-row hand-derived oracle table), Cypher rows, Cypher `count(*)`, index-backed route, serving-lane invariance. Negative: dangling src / dangling dst / both-missing / all-dangling / empty-but-bound node table / self-loop on a missing node / synthesized-from-edges (must NOT be gated). Oracles are hand-computed literals — never engine agreement.

**Anti-vacuity: 33 of 75 executed cases fail at `0064072f7` (pre-#1888), 0 at the fix.** The matrix discriminates by surface: pandas chain already pruned so it passes at base, while pandas cypher/hop fail there (the #1808 asymmetry) and polars chain fails (the #1888 F-01 gap).

## Two real bugs, neither a comment problem
1. **Self-inflicted, caught before push:** a helper insertion landed *between* `@otel_traced("gfql.hop")` and `def hop()`, silently moving the decorator onto the helper and leaving `hop()` untraced. ruff, mypy and the hygiene guard all passed with it broken — worth noting the otel decorator has no test that would catch this.
2. **Pre-existing, from #1888:** the gate's `is_in(<Series>)` is deprecated in polars 1.42 and emitted ~15k warnings per suite run. Switched to `.implode()`, verified value-identical on plain / null-in-ids / null-in-endpoints / empty fixtures. Suite warnings fell **18,335 → 3,197**.

## Perf — flagged, not resolved
Local microbenchmark (loaded 24-core box, **not** a board result), `5ddbd666d` vs parent, 11 interleaved reps, on graphs with **zero dangling endpoints** where the gate provably cannot change the answer:

| 2.4M edges, clean | before | after | Δ |
|---|---|---|---|
| polars chain `n/e/n` | 22.6 ms | 61.2 ms | **+171%** |
| polars seeded hop (1k seeds) | 24.2 ms | 56.0 ms | **+131%** |
| polars unseeded hop | 184 ms | 281 ms | +53% |
| pandas seeded hop (1k seeds) | 197 ms | 291 ms | +47% |

pandas chain within noise; pandas unseeded hop actually *faster* (767→744 ms, the epilogue now skips an unconditional endpoint concat). Structurally the worst shape is a seeded 1-hop: now O(E) regardless of frontier size — the gate costing more than the query it guards.

This needs reconciling with the full-board A/B, which showed polars q1/q2 **flat** and only ~1% aggregate. A dgx board A/B is running with instructions to prove *which lane* serves the board queries rather than infer it, and to re-run these shapes on an idle box. **Not merging this until that lands** — if the gate is genuinely this expensive off-board, the mitigation to weigh is a cheap closure precondition (skip the work when endpoints are already known-closed) rather than paying O(E) on every bound-node-table query. That is a product call, not a cleanup.

graphistry/compute/gfql/lazy/engine/polars/chain.py
```diff
@@ -1035,12 +1041,9 @@ def _plain_edge(op):
...
         added_edge_index = True
-        # with_row_index only PREPENDS a synthetic id column; the indexed src/dst are
-        # preserved by value. Re-point any resident #1658 adjacency index at the new
-        # edge frame so the seeded fast path still engages through the native polars
-        # chain executor (mirrors compute/chain.py — else the identity guard misses
-        # and every hop falls back to the O(E) scan). Enables typed-edge chains on
-        # polars/polars-gpu (untyped already engaged; the frame swap blocked typed).
+        # with_row_index only PREPENDS a column, so the indexed src/dst survive by value:
+        # re-point any resident adjacency index at the new edge frame, else the index's
+        # identity guard rejects it (mirrors compute/chain.py).
```
lmeyerov: should this comment be instead (mostly) tests?

graphistry/compute/gfql/lazy/engine/polars/dtypes.py
```diff
@@ -62,10 +62,9 @@ def is_stringlike(dt: "Optional[PolarsDType]") -> bool:
     return False
 
 
-# --- cross-ENGINE dtype classification (pandas/numpy/arrow/polars), the pushdown
-# planner's contract. Lives with the polars vocabulary because the polars arm is
-# the one that made single-sourcing load-bearing: pandas' classifiers return a
-# confident False for polars dtypes, so per-site fallbacks silently diverged. ---
+# --- cross-engine dtype classification (pandas/numpy/arrow/polars) for the pushdown
+# planner. Single-sourced here: pandas' classifiers return a confident False for polars
+# dtypes, so per-site fallbacks silently diverge. ---
```
lmeyerov: should this comment instead be tests?

graphistry/compute/gfql/lazy/engine/polars/nan_clean.py
```diff
@@ -23,15 +23,11 @@
...
-# is undefined behavior for caches and results; the supported recipe after
-# mutation is REBIND (g.nodes(df)/g.edges(df)) or gfql_clear_caches(). This
-# cache is REGISTERED so that recipe actually works -- the pre-#1883 version was
-# invisible to gfql_clear_caches and the completeness lock, which is what turned
-# a contract violation into an unflushable wrong answer.
+# BOUND-FRAME IMMUTABILITY CONTRACT: frames handed to GFQL are treated as immutable.
+# Mutating one in place (pandas .loc, polars extend/replace_column/insert_column/
+# hstack(in_place=True)) is undefined behavior for caches and results; the supported
+# recipe after mutation is REBIND (g.nodes(df)/g.edges(df)) or gfql_clear_caches().
+# This cache must stay REGISTERED for gfql_clear_caches to reach it.
```
lmeyerov: This comment can be 1-2 lines... or maybe even eliminated via clearer variable naming..

graphistry/compute/hop.py
```diff
@@ -45,6 +45,36 @@ def query_if_not_none(query: Optional[str], df: DataFrameT) -> DataFrameT:
     return df.query(query)
 
 
+def _seed_ids_the_traversal_reencountered(
+    matches_nodes: Optional[DataFrameT], node_col: str,
+) -> Set[Any]:  # hygiene-ok: explicit-any -- node ids are heterogeneous engine scalars (int/str/date)
```
lmeyerov: so narrow to say that vs being Any

lmeyerov:
important that we have sufficient positive/negative testing, incl across engines, given then nature of these changes

graphistry/compute/gfql/lazy/engine/polars/hop.py
```diff
@@ -15,10 +15,8 @@
 
 def hop_lazy_or_eager(self: Plottable, nodes: Optional[Any] = None, hops: Optional[int] = 1, **kwargs: Any) -> Plottable:
     """Run the polars hop: lazy collect-once for a single bounded hop, eager loop otherwise."""
-    # Normalize input eagerness ONCE, mirroring the chain_polars entry (#1740): LazyFrame is
-    # an accepted INPUT format, but the BFS loop mixes user frames with eager wavefronts and
-    # polars joins do not mix eagerness. Direct .hop() reaches here without passing through
-    # chain_polars, so the chain-entry normalization alone leaves this route crashing.
+    # Normalize input eagerness ONCE (mirrors the chain_polars entry): polars joins do not
+    # mix eagerness, and direct .hop() reaches here without passing through chain_polars.
```
lmeyerov: should this be programmatically tested instead of a comment?

graphistry/compute/gfql_unified.py
```diff
@@ -1051,14 +1051,9 @@ def _execute_compiled_query_via_physical_plan(
...
-        # target is established -- so engine='polars-gpu' serves fast-path work
-        # on CPU (#1824: eager arms never collect, and dgx measured 86 polars-gpu
-        # tests green only via that mislabel). The target stays CPU here until
-        # the fast paths are made GPU-or-decline arm by arm (#1824, next cycle);
-        # the collect routing + NIE-decline plumbing below is already in place
-        # and is a no-op on the CPU target. Flip to ExecutionTarget.GPU on
-        # POLARS_GPU resolutions only together with the per-arm decline work.
+        # Fast paths run before the chain route establishes the execution target, so they
+        # serve on CPU even under engine='polars-gpu'. Do not flip this to GPU without
+        # making each fast-path arm GPU-or-decline (#1824).
```
lmeyerov: should this comment be tests, and move remaining comments like flipping to those tests?

graphistry/compute/gfql_unified.py
```diff
@@ -2080,12 +2072,9 @@ def gfql(self: Plottable,
...
         except NotImplementedError:
-            # AUTO must answer: pandas explicitly, since the generic path would
-            # re-resolve these frames to POLARS and re-raise the same NIE. The
-            # frames must be COERCED first -- the pandas executors are
-            # pandas-idiom, and handing them polars frames crashed 7/7 same-path
-            # projection shapes on the default route (round-002 agent-03 BUG-2;
-            # #1885 was one corner of it).
+            # pandas explicitly, not AUTO: the generic path would re-resolve these frames
+            # to POLARS and re-raise the same NIE. Coerce first -- the pandas executors
+            # are pandas-idiom and do not accept polars frames.
```
lmeyerov: unnecessary comment or at least code-encodable?

graphistry/compute/hop.py
lmeyerov: still open

other PRs already figured out how to turn these into types..

lmeyerov:
Round 2 pushed as `c11714cf6`. Mapping below for you to verify and resolve — **6 unresolved / 7 threads** before this push; I am not resolving any myself.

## Your four inline sites

| Site | What changed |
|---|---|
| `chain.py:1046` "should this be (mostly) tests?" | Deleted — its claims are already pinned by `test_index.py::test_rebind_edges_revalidates_after_shallow_augmentation` and `::…drops_index_on_fingerprint_mismatch`. See finding 1 below. |
| `dtypes.py:67` "should this instead be tests?" | Both traps deleted → `test_polars_dtype_classifier_contracts.py` (20 cases): pandas' `is_numeric_dtype` returns a confident `False` **without raising** for polars dtypes — which is *why* polars must be consulted first, since a `try/except` fallback never fires — plus `Decimal` numeric via polars' own `is_numeric()`, and the `"str"`-exact rule with the fallback branch **forced by monkeypatch**. |
| `nan_clean.py:30` "1-2 lines or clearer naming" | `_PL_NAN_CLEAN_CACHE_IDS` → `_nan_free_frame_id_cache`, `key` → `frame_id`, contract cut to 2 lines — **and a perf claim I missed in round 1 removed** ("O(E)-per-call into O(1)… the dominant per-call cost"). |
| `hop.py:50` "narrow to say that vs being Any" | `Set[Any]` + `hygiene-ok` → `Set[NodeId]`, `NodeId` defined in `compute/typing.py`. The comment stated the type; now the type does. |

## Your overall note — "sufficient positive/negative testing, incl across engines"

New pins with **anti-vacuity by mutation** (revert the fix, confirm the pin fails):

| Pin group | Cases | Mutation | Result |
|---|---|---|---|
| dtype classifier contracts | 20, pandas+polars | consult pandas first | **4 fail** |
| `.implode()` value-identity + no-warning | 5 shapes | revert to bare `is_in` | **1 fail** |
| extracted predicates | 14 | — | direct unit pins |
| NaN-cache GC eviction | 1 | drop `weakref.finalize` | **1 fail** |
| `hop()` node-table de-dup | 4 | restore conditional de-dup | **0 fail** |

That last row is reported, not hidden: the mutation failing nothing **proves the else-branch de-dup was already redundant**, so my round-1 refactor is behavior-identical — but those 4 are not anti-vacuous for that edit.

## Re-audit of every round-1 keep, under the corrected bar

**Converted to tests:** the polars int/float join-key trap, the `filter_by_dict` import-cycle note (already covered by `test_pandas_without_polars.py`), the strong-ref note (key renamed `__edges_strong_ref__` + cache-identity pin), the `PolarsDType` metaclass claim.

**Survivors, each with why a test/name cannot carry it:** the cudf-merge note (only checkable on a GPU box — a pin would be a permanent skip on every CPU lane); the `df_to_engine`-not-`ensure_engine_match` note (states which *API* to call, so a test passes either way); the TRANSITIONAL policy-hook guard (carries a **deletion condition** — a test pins behavior, not the intent to remove); and the openCypher scoping notes (cite *which spec rule* the pinned behavior implements).

Added product comments: **173 → 95 → 60**. Zero perf vocabulary. Exactly one issue reference remains, a forward-pointer ("don't flip to GPU without #1824"), not a test substitute.

## Three findings

1. **I deleted a test I wrote.** The `chain.py` integration pin passed with the re-point *removed* — that call changes no observable result or index trace in any shape reachable today. Rather than ship a green test asserting nothing, I dropped it and recorded that **the branch may be dead** or its claim stale.
2. **A rename that would have silently weakened CI.** The clear-caches completeness lock discovers caches by the substring `cache`/`memo`. My first rename removed it, quietly dropping this cache out of that guarantee. Caught pre-commit; the final name stays discoverable.
3. **Real divergence, not introduced here:** pandas `hop()` de-dups its output node table by id, **polars does not** (`[0,0,1]` vs `[0,1]`). Reproduces at `86013f4`. Pinned strict-xfail so the fix flips it.

Perf untouched — the dgx board A/B still gates this merge.

graphistry/compute/hop.py
lmeyerov, line 1: a lot of changes to hop.py so pos/negative testing & amplification->convergence proof obligation as part of the pr

lmeyerov:
**dgx verdict on the closure gate: user-visible but off-board — proven, not inferred.** This is the finding that gates this PR's merge, so the evidence matters.

## Board: 0 of 36 cells move
18 slots/build/scale, frozen sha-verified builds, `--network none`. **Values byte-identical** A vs B, one distinct value per cell across all 36 slots, zero errors. No cell qualifies as significant (p<0.05 *and* bootstrap CI excluding zero).

A polars q5/q6 "+29-30%" in the raw table is a **bimodal sampling artifact**: per-slot values are bimodal in *both* arms (~9 ms and ~13 ms modes), base drew 10 fast/8 slow and fix drew 5/13, so the medians landed on opposite sides. A dedicated 90-iteration-per-build diagnostic — the mode flips *within a single process* mid-run — pools to **q5 −0.5%, q6 −2.6%**.

## Why the board is silent: it never executes the gate

Two instruments, both with positive controls so they cannot fail open:
- **`sys.settrace` line-hit** on the three gate lines plus kernel entries and `index/traverse.py`. On direct `g.hop(seeds,1)` and `g.chain([n(),e_forward(),n()])` the probe fires `KERNEL_pandas_hop`+`GATE_pandas_isin`, `KERNEL_polars_hop`+`GATE_polars_is_in`, `LANE_polars_chain_3opFP`+`GATE_polars_chain_allids`. On **all 9 board queries × 2 engines × 2 builds × 2 scales: zero lines fired.**
- **cProfile**: `compute/hop.py` and `polars/hop_eager.py` frames are entirely absent from every board query.

Lanes actually serving the board: q1-q4 `_execute_single_hop_grouped_aggregate_fast_path`, q5-q7 `_connected_join_two_star_fast_grouped_count`, q8/q9 `_execute_two_hop_count_fast_path` — all fused planners *above* the traversal kernels. The index route isn't involved (default policy `use`, empty registry).

**Counterfactual that rules out "the board is just too small":** the board's own 100k graph is **2.42M FOLLOWS edges — the same scale as the microbenchmark** — with the node table bound and **zero dangling endpoints** (audited per relation). Priced on that exact frame the gate costs pandas 10.4-10.9 ms and polars 16.7-19.2 ms; against polars cells running 7-27 ms that is +65% to +240%. It would have been unmissable had the lane been taken. The board is silent because the code doesn't run.

## Off-board, on an idle box, it is larger than first reported

Synthetic 100k nodes / 2.4M edges, **zero dangling** (gate provably cannot change the answer), 24 slots/build interleaved, values identical in all 12 cells:

| engine | binding | shape | before | after | Δ |
|---|---|---|---|---|---|
| polars | **bound** | seeded hop (1k) | 11.29 ms | 32.90 ms | **+191%** |
| polars | **bound** | chain n/e/n | 10.45 | 26.38 | **+152%** |
| polars | **bound** | unseeded hop | 72.69 | 95.27 | +31% |
| pandas | **bound** | seeded hop (1k) | 55.15 | 75.79 | **+37%** |
| pandas | **bound** | chain / unseeded hop | — | — | ns (see below) |
| any | **unbound** (control) | all three | — | — | **no regression** |

The unbound control gives clean causal attribution. The two flat pandas cells have confirmed mechanisms rather than being noise: pandas `chain n/e/n` never reaches `compute/hop.py` (a chain fast path serves it — the tracer's `KERNEL_pandas_hop` does not fire), and pandas *unseeded* hop pays the gate but also stops paying a previously unconditional endpoint-backfill concat, which offsets it at full frontier but not at 1k seeds.

**What a user actually pays** for a seeded 1-hop on a 2.4M-edge graph with a bound node table: **polars +21.6 ms, pandas +20.6 ms** — O(E) regardless of frontier size, i.e. roughly twice the query it guards.

## Consequence for this PR

Not merging on this. A perf lane is now working the structural fix — fold the closure into the join the traversal already performs (the same shape that fully recovered the q1/q2 regression earlier in this stack), or restrict the check to the frontier rather than all E, or gate on a *proof* of closure such as an index fact. Explicitly not on the table: any heuristic skip that is not a proof, or weakening the contract — it repaired silent wrong answers on five surfaces.

Housekeeping from the same run: a stale self-matching `pgrep` watcher (12h23m old) was found and killed on the box.

lmeyerov:
Review round 3. Head `8a7591b41` (work in `f40148f10`, base merge `1a5d56e1c`, lane fix `8a7591b41`).

Two of these were already discharged before this round; I say so explicitly rather than claiming credit. **Outdated ≠ addressed** — I checked every thread against the current head.

---

**1. `hop_eager.py:208` — "may violate rules about preferring perf tracking in pyg-bench, audit this + stack and fix"**

The flagged comment itself was already gone at head (replaced in `c11714cf6` by the named helpers `_keep_edges_with_both_endpoints_resolvable` / `_ids_an_endpoint_may_resolve_to`) — but the *audit* you asked for had not been done. Done now across the whole stack, in `f40148f10`: every perf/asymptotic/benchmark claim deleted from comments in `hop_eager.py`, `polars/chain.py`, `polars/nan_clean.py`, `compute/hop.py`, `gfql/index/traverse.py`, `gfql_fast_paths.py` (~60 sites: `O(E)`/`O(degree)`/`O(N)`, "measured", "profiled", "dgx measured", "receipts in pyg-bench", ms timings, "hot path", "vectorization-first", "bought for nothing").

The worst offender was the `_idframe_lf` docstring — a 12-line perf essay ("full hash pass … bought for nothing", "O(N) work inside a query whose answer is O(degree)"). Its *correctness* content was already pinned by `test_semi_join_key_frame.py`, so it collapsed to two lines of semantics.

Verified comments-only mechanically: for each of the 6 files I parsed head and working tree, stripped docstrings, and compared `ast.dump` — **AST-identical**, no executable line moved. `grep -nE 'O\(|measured|pyg-bench|hot path' ` over those files now returns nothing in comments.

**2. `hop.py:1003` — "another comment violation"**

Already fixed in `c11714cf6`, before this round: the comment is gone, replaced by `_endpoint_ids_without_node_rows` (`grep -n "_endpoint_ids_without_node_rows" graphistry/compute/hop.py`). What it *claimed* (no NaN-stub rows, no int64→float64 upcast) had no test; it does now — `test_direct_hop_keeps_node_attribute_dtypes`, which **fails at the merge-base** (pre-gate, direct `hop()` on the MIXED fixture returns node 7 with NaN `v` and dtype float64).

**3. `chain.py:1046` — "should this comment be (mostly) tests?"**

Already deleted in `c11714cf6`. The contract is carried by `test_index.py::test_rebind_edges_revalidates_after_shallow_augmentation` and `::test_rebind_edges_drops_index_on_fingerprint_mismatch`.

**4. `dtypes.py:67` — "should this comment instead be tests?"**

Already deleted in `c11714cf6`; the banner became `test_polars_dtype_classifier_contracts.py` (7 tests), incl. `test_pandas_numeric_classifier_says_false_for_polars_dtypes_and_does_not_raise` — the exact claim the comment made.

**5. `polars/hop.py:19` — "should this be programmatically tested instead of a comment?"**

Comment deleted; extracted `_with_every_lazy_input_collected(self, nodes)` — the name states it, the one-line docstring is the constraint (polars refuses to mix a LazyFrame with an eager frame). Also dropped the "pay-as-you-go" perf phrase below it. Pinned by the existing `test_engine_polars_hop.py::test_lazy_eagerness_matrix_matches_pandas` / `::test_lazy_seed_parameter_matches_pandas`. **Mutation-proved**: making the helper a no-op → **15 failures**.

**6. `gfql_unified.py:1072` — "should this comment be tests, and move remaining comments like flipping to those tests?"**

Comment deleted; the constant became `_fast_path_execution_target_ignoring_requested_engine(engine)` — the name carries "the requested engine does not move the target". The adjacent NIE comment was deleted too. New `test_gfql_unified_routing_contracts.py` pins both:
- `test_fast_paths_target_cpu_whatever_engine_was_requested` (7 engine values incl. `polars-gpu`)
- `test_fast_path_body_actually_runs_under_the_cpu_target` — observes `active_target()` from inside the fast path, so it pins the wrapping, not just the constant
- `test_cpu_fast_path_not_implemented_error_is_not_swallowed`

**Mutation-proved**: flipping the target to GPU → **9 failures**.

**7. `gfql_unified.py:2096` — "unnecessary comment or at least code-encodable?"**

Unnecessary — both facts it stated were already visible in the two expressions right under it (`engine=Engine.PANDAS.value`, `_coerce_input_formats(self, Engine.PANDAS)`). Deleted. Pinned by `test_auto_polars_decline_reruns_on_pandas_with_coerced_frames`. **Mutation-proved**: re-entering on AUTO without coercing → **1 failure**.

I also took the neighbouring TRANSITIONAL comment ("move remaining comments like flipping"): it became `_policied_auto_serves_via_pandas_until_the_polars_route_emits_hooks(...)` plus `test_policied_auto_predicate` (5 cases) and `test_mixed_frames_with_a_policy_are_still_routed_to_pandas` — the mixed-frame case the comment warned about. **Mutation-proved**: swapping the predicate for a frame-shape check → **1 failure**.

---

**8. `hop.py` file-level — pos/negative testing & amplification→convergence proof obligation**

**This found a real regression on the branch, and I fixed it.** Details below.

*Cells.* **243** new hand-oracled cells appended to `test_hop_boundary_matrix.py`'s sibling `test_endpoint_closure_matrix.py` (extended, not a parallel file), + 16 routing cells + 1 divergence pin. Axes: hop window (`hops` 1–4, `min_hops`, `output_min_hops`/`output_max_hops` incl. out-of-range), `to_fixed_point` vs bounded, direction (fwd/rev/undirected), seed cardinality (0/1/many/all), topology (path, cycle, self-loop, parallel edges, disconnected, isolated node, empty edge table), endpoint filters on source *and* destination, `edge_match`, hop labelling, `return_as_wave_front`, `include_zero_hop_seed`, `target_wave_front`, and string ids. Every expectation hand-derived from the fixture edge list and written as a literal — engine agreement is never the oracle.

*Anti-vacuity.* **103 of the 188 runnable new closure cells FAIL at the merge-base `470cd5e45`** (85 pass, 55 GPU-skipped). The 85 that pass at base are honest positives/controls — e.g. `test_empty_edge_table_...is_empty_not_a_crash` (crash guard), `test_unreached_seed_is_dropped_on_a_fully_closed_graph_too` (the control proving an empty node frame is pre-existing `hop()` behaviour, not closure eating the seed), and the parity-or-NIE decline pins. The 16 routing cells cannot run at base at all (the symbols don't exist there), so their evidence is mutation, reported above.

*Engines.* pandas + polars on every cell; cudf and polars-gpu parametrized. **Note: this box does have a working GPU and cuDF 25.10**, so the cudf lane really executed — it is not silently skipping.

*Round 2 found things.* Round 1 swept seeds/topology/direction/hops. Round 2 (windows, labelling, wavefront, cycle, string ids, `edge_match`, target wavefront) was **not** empty:
- **A regression introduced earlier on this branch.** Making the endpoint concat conditional also removed the hop-label column that concat attached *as a side effect*; downstream readers assume it exists. Writing it through `.loc` creates it implicitly on pandas but **raises on cuDF**, so seeded hop windows / `min_hops` **crashed on cuDF** (`ValueError: Need to pass dtype when passing pd.NA or None`) — on graphs with zero dangling edges, where the gate is a no-op. Fixed by attaching the column explicitly on every engine.
- To bound it I ran a **1200-cell base-vs-head differential** (6 topologies × 2 engines × 3 directions × seeded/unseeded × 25 kwarg sets). Head is now **value-identical to the merge-base on all 1200** and **fixes 18 cuDF crashes** that exist at base.
- A **pre-existing** cuDF divergence found on the way: under an output hop window cuDF returns the sliced edge without backfilling its source node row (pandas backfills), leaving an edge whose endpoint has no node row. Reproduces identically at the merge-base, so #1888/#1895 did not cause it — pinned as a strict xfail, `test_output_hop_window_backfills_the_source_node_row_on_cudf`.
- Also corrected two of my own wrong oracles rather than bending the code: `include_zero_hop_seed` does not retain an unreached seed on direct `hop()` (verified on a clean graph, so not closure-related), and the polars lane *declines* `include_zero_hop_seed` / output windows with NIE — now pinned as loud-decline contracts so a future silent implementation can't land.

*Convergence.* Round 3 (this one) was mutation-driven rather than shape-driven and found **two decorative pins of my own**: with the backfill deleted and with the label attach removed, my cells still passed. Both are now real — see the mutation table. A further pass over the same axes surfaced nothing new.

*Mutation table* (all reverted after measuring):

| mutation | result |
|---|---|
| M1 closure gate one-sided (source only) | **69 failed** |
| M2 closure gate disabled | **91 failed** |
| M3 endpoint backfill deleted | **4 failed** |
| M4 epilogue hop-label attach removed | **1 failed** |
| M4b per-write hop-label ensure removed | **1 failed** |
| M5 fast-path target flipped to GPU | **9 failed** |
| M6 policy predicate → frame-shape check | **1 failed** |
| M7 AUTO decline re-enters on AUTO, uncoerced | **1 failed** |
| M8 lazy-input collection no-op | **15 failed** |

---

**Gates** (all re-run on the merged tree, after pulling the moved base `914ea50e3`):
- `ruff check graphistry bin` — clean
- `ci_type_hygiene_guard.py` — OK, no growth. My first pass added 5 `Any`s and tripped it; fixed by typing them properly (`Union[EngineAbstract, str]`, `Dict[str, PolicyFunction]`, `DataFrameT`) rather than raising a baseline.
- `ci_cypher_surface_guard.py` — pass
- `mypy graphistry` — 4 errors, exactly the known pre-existing ones in `polars/hop_eager.py` + `polars/degrees.py`
- `graphistry/tests/compute` — **102 failed, 10134 passed**; failure set **byte-identical** to the branch baseline (`comm` both directions empty). One merge conflict, in `gfql_unified.py`, resolved by taking the base's `row_guard_needs_single_column_entity_text` form; `_slice_rows` and `LABEL_FLAG_PREFIX` left intact.

Threads left unresolved for you to close.

lmeyerov:
## Amplification round 4 — verdict: **NOT CONVERGED**

Round 3 was not quiet (two decorative pins). This round was **also not quiet**: it found one real
uncovered contract and one uncovered intersection, both now pinned. Under the two-consecutive-quiet-rounds
standard that means **one more round is still required** — this one cannot be the first of the two.

Head: `5c55db21afbe56cad48731f68e68fec6ec102173` · `mergeable: MERGEABLE` · 2 workflow runs created.

---

### 1. Mutation audit of the existing pins (the headline activity)

I broke the code behind **22 distinct behaviours** this PR claims — every functional site in the
diff across `hop.py`, `polars/hop_eager.py`, `polars/chain.py`, `polars/dtypes.py`,
`polars/nan_clean.py`, `gfql/index/traverse.py`, `gfql/row/pipeline.py`, `gfql_unified.py` — and
checked whether a test went red.

| | count |
|---|---|
| behaviours mutation-tested | 22 |
| caught by an existing test | **20** |
| survived | 2 |
| **decorative pins found** | **0** |

Zero decorative pins: unlike round 3, no test in this PR passed while the code it names was broken.
The two survivors were not decorative pins — they were *unpinned code*:

- **`hop.py` — `_ensure_node_hop_col()` in the `direction == 'undirected'` arm: REAL GAP, now pinned.**
  Deleting that call is silent on pandas (`.loc` creates the column) and raises
  `ValueError: Need to pass dtype when passing pd.NA or None` on cuDF. Nothing in the suite covered it.
  A 1944-shape pandas-vs-cuDF sweep over undirected hops shows **this PR fixes 288 cuDF crash shapes**
  (0 new crashes, 0 new divergences) — an entire fixed family with no test attached to it.
- **`hop.py` — `_ensure_node_hop_col()` inside `if len(seed_endpoint_hops) > 0:`: NOT a gap.**
  That block requires a non-empty `g_out._edges`, which is exactly the condition under which the
  backfill epilogue has already materialised the column, so the call is provably a no-op there.
  Corroborated empirically: instrumented over ~139k `(graph, seeds, kwargs)` hop shapes (43.2k mixed
  direction + 32k undirected + 64k forward/reverse) the site **never once** adds the column, while the
  other two sites fire 384 and 498 times. Left in place as defensive symmetry across the three
  `.loc` writers — reporting it rather than pretending it is covered.

---

### 2. New cells, and how many fail at the merge-base

3 tests / 10 parametrised arms (8 run here, 2 skip: `polars-gpu` needs `cudf_polars`, absent from
this image). **5 of the 8 fail at the merge-base `526976e9`:**

| cell | pandas | polars | cudf |
|---|---|---|---|
| `test_optional_match_null_extends_a_dangling_endpoint` | **fails at base** | passes (control) | **fails at base** |
| `test_optional_match_undirected_null_extends_a_dangling_endpoint` | **fails at base** | passes (control) | **fails at base** |
| `test_undirected_zero_hop_seed_under_an_output_window_labels_before_it_strips` | passes (control) | n/a (typed NIE, already pinned) | **fails at base** |

The 3 base-passing arms are named controls, not padding: polars was already the correct side of #1808
on the OPTIONAL MATCH surface, and pandas never raised on the missing label column because `.loc`
creates one. Every expected value is hand-walked from the fixture tables and written as a literal;
cross-engine agreement was not used as an oracle anywhere.

Verified the new cells earn their place: the cuDF label cell is the **only** thing in the whole suite
that kills the undirected-`_ensure_node_hop_col` mutation, and the OPTIONAL MATCH cells cover a
surface that was returning the wrong answer at base while master was green (so nothing covered it).

---

### 3. What the newly-merged surfaces actually revealed

**#1894 OPTIONAL MATCH × closure — the PR fixes a broad family, none of it was pinned.** A 15-query
Cypher sweep, base vs head, on pandas/polars/cudf: `opt-fwd`, `opt-und`, `opt-count`, `opt-typed`,
`opt-2hop`, `opt-filtered-drive`, `opt-chain2`, `with-opt` all bound the optional alias to an *unbacked*
id on pandas and cuDF at base and NULL-extend correctly at head. (e.g. `count(b)` for the unmatched row:
`1` at base → `0` at head.)

**#1893 `to_fixed_point` == saturated bounded × closure — clean, no new cells warranted.** 216
`(graph, engine, seed, direction, wavefront)` shapes, base vs head: 8 violations at base, the *same*
8 at head, **0 new**; all 8 are the undirected `return_as_wave_front` seed strip. Adding a pin here
would have been vacuous, so I did not.

**#1893 seeded `hops==1` source-filter domain × closure — correct on all three engines at head**
(base kept the dangling edge on all three). Not pinned separately: the gate mutation `M1` is already
caught, so a cell here would kill no mutation the suite does not already kill.

---

### 4. Findings that are NOT this PR's (reported, not pinned — each needs its own issue)

1. **cuDF `sum()` over an empty OPTIONAL MATCH group returns NULL; pandas and polars return 0**
   (openCypher says 0). Reproduces *identically at the merge-base on a graph with no dangling ids at
   all*, so it is independent of #1888. Sibling of the existing ungrouped xfail
   `test_sum_over_empty_match_is_zero`.
2. **polars keeps the seed in an undirected `to_fixed_point` + `return_as_wave_front` hop where
   pandas/cuDF strip it.** Also reproduces at the merge-base on a fully closed 2-node graph.
3. The 576 remaining pandas-vs-cuDF value divergences in the undirected + output-window sweep are all
   the already-xfailed `test_output_hop_window_backfills_the_source_node_row_on_cudf` family. Nothing new.

---

### 5. Gates

- `ruff check graphistry bin` — clean.
- `ci_type_hygiene_guard` OK (no growth) · `ci_comment_density_guard` OK (no growth) · `ci_cypher_surface_guard` ✅. No baseline raised.
- `mypy graphistry` — 4 errors at head, **the same 4 at `526976e9`**. No new errors.
- **Full `graphistry/tests/compute`, GPU box, cuDF 25.10 really executing:** head `102 failed, 10198 passed, 946 skipped, 43 xfailed`; base `102 failed, 9861 passed, 863 skipped, 44 xfailed`.
  **Failure sets are byte-identical — zero delta in both directions** (`comm -13` and `comm -23` both
  empty). The 102 are pre-existing local-environment failures (`gfql/index`, polars conformance,
  umap). The single xfail delta is `test_1808_dangling_destination_pandas_matches_polars`, which this
  PR intentionally flips from strict-xfail to passing.
- Self-review grep on the diff: my change touches only `graphistry/tests/`, so it adds no comments,
  `Any`, `cast(`, `type: ignore` or `hygiene-ok` to product code.

lmeyerov:
## Convergence round 5 — **NOT QUIET**

Head `e8d061a7d` (was `5c55db21a`). Four real coverage gaps, **two decorative pins**, one dead line removed. The trend from round 4 did not hold: the defect class is back to bad tests, not just unpinned code.

### Mutation audit

**40 functional sites in the diff mutated · 30 caught · 10 survived.**

Sites: hop.py closure gate (6), hop.py backfill epilogue (6), `_ensure_node_hop_col` sites + the strip (4), polars `hop_eager` closure (7), polars hop entry (2), polars `chain` single-hop (4), `gfql_unified` routing (4), row-pipeline cache (2), `nan_clean` cache (3), `index/traverse` (2).

Of the 10 survivors:

| | verdict |
|---|---|
| gate forced on with **no** node table bound (pandas / polars hop / polars chain) | **equivalent** — the synthesized table contains every endpoint id, so the gate is the identity. This is the "vacuously closed" case; no test can kill it. |
| polars **lazy seed** collect removed | **equivalent** — verified directly, the seed path never calls a DataFrame-only method |
| third `_ensure_node_hop_col()` | **dead** — removed (below) |
| index-path duplicate-id decline | **out of diff** — comment-only change here, and the index lane cannot run on this box (cupy can't load `libnvrtc.so.12`). Flagged, not fixed. |
| **4 remaining** | **real gaps** — pinned below, each re-verified as now CAUGHT |

### The two decorative pins (round 4 reported zero)

`test_duplicate_node_rows_are_deduped_when_no_endpoint_is_missing` and `..._when_an_endpoint_is_backfilled` **pass with the line they name deleted**. `output_max_hops` defaults to `max_hops`, so on every *bounded* hop the output-window epilogue (`hop.py:1181`) de-dups again and masks the backfill de-dup entirely — 0 differences across a 3360-shape dup-id sweep (pandas + cuDF).

`to_fixed_point=True` is the only shape that reaches the caller un-de-duped (`max_hops` stays `None`, epilogue skipped): `[0, 0, 1]` instead of `[0, 1]`. Pinned there; the two bounded tests are relabelled as the controls they actually are.

### The other three gaps

- **The undirected seed hop-label STRIP.** Round 4 pinned that `_ensure_node_hop_col()` makes the column *exist* (the cuDF crash). Nothing pinned what the strip *writes*. Removing `g_out._nodes.loc[seed_mask_all, node_hop_col] = NA` changes 12 shapes with nothing red. Now pinned by hand oracle on pandas + polars. **cuDF drops the seed's node row entirely** here — pre-existing at merge-base, leaves edge `(0,1)` with no row for `0` on a fully closed graph — strict xfail naming it.
- **The polars single-hop chain's reverse endpoint mapping.** Collapsing `n_from, n_to = (n0, n2) if direction != "reverse" else (n2, n0)` to `(n0, n2)` mirrors the answer on all three reverse shapes and the whole compute suite stays green. This block is rewritten by this PR. Pinned with a hand-walked 4-cycle oracle on every engine.
- **`_with_every_lazy_input_collected`.** No test passes a LazyFrame to `.hop(engine='polars')`. The PR's own closure kernel calls `get_column`, which a LazyFrame has not got — removing the collect raises `AttributeError`.

### Dead code removed

The third `_ensure_node_hop_col()` is unreachable *by proof*, not just by instrumentation: reaching it requires `edge_hop_col in g_out._edges.columns` and a non-empty `seed_endpoint_hops`, hence non-empty edges, hence the backfill block at `hop.py:1029` already ran and added the column (its other three conditions are implied by the enclosing guard). Removed.

### Round 4's three leftovers

1. **cuDF `sum()` over an empty OPTIONAL group** — out of scope (identical at merge-base, fully closed graph). On this box it is worse than reported: cuDF cannot serve **any** grouped aggregate after `OPTIONAL MATCH` — the row pipeline's `_build_grouped` → `grouped.size()` raises *"truth value of a Series is ambiguous"* for `count`/`sum`/`avg`/`min` alike. **Pinned as a strict xfail** on the value (so it flips whether cuDF starts answering `NULL` or answering correctly).
   **polars keeping the seed in undirected tfp + wavefront** — **already pinned**: `test_hop_boundary_matrix.py::REDISCOVERY_XFAIL` carries 9 polars cells with exactly that reason. No new pin; a redundant one would be noise.
2. **`#1893 to_fixed_point == saturated`** — **already covered**, and round 4's "a pin would be vacuous" was right for the wrong reason: `test_undirected_tfp_equals_saturated_bounded` pins it per cell with 10 strict xfails naming the known-wrong pandas set. My own 324-shape sweep reproduces 20 violations, all in that one class. What was *not* covered is **cuDF** — measured, and it lands on **exactly** the pandas cells of both xfail sets, so a cuDF arm is added to that matrix (mirrored, not re-listed, so cuDF-only drift shows up as an unexpected pass).
3. **The third `_ensure_node_hop_col()`** — confirmed independently and removed (above).

### Unexplored intersections

`label_seeds`, multigraph/self-loop, empty edge table, synthesized node table and output windows on cuDF were all already covered. **Traversal `min_hops` was not** — and it is a genuine closure axis: at merge-base `min_hops=4` on LADDER returns the dangling `3->9` edge and node 9; at head it is correctly empty. New axis added.

### Numbers

- **91 new cells**, 0 removed. **7 fail at merge-base** (the `min_hops` axis).
- The other 84 pass at merge-base **by design** and are named as such: they are *mutation-justified*, not base-red — the PR preserves the behaviour, the mutation audit proved nothing was pinning it. 17 are strict xfails that xfail at base too (6 new cuDF + 11 cuDF boundary-matrix cells).
- Engines: **pandas, polars, cuDF** (cuDF 25.10 really executes here). `polars-gpu` skips with a named reason (`cudf_polars` absent); the index-backed lane skips with a named reason (cupy/`libnvrtc`).

### Gates

- `ruff check graphistry bin` — clean
- comment-density / cypher-surface / type-hygiene guards — all clean, run from inside the worktree
- `mypy graphistry` — 4 errors at base, the same 4 at head (line numbers shifted only), **no new**
- `graphistry/tests/compute` full: base `102 failed, 9861 passed`; head `102 failed, 10264 passed, 62 xfailed`. **Failure sets are byte-identical** (`comm` both directions empty). Both runs printed their summary line.

**Verdict: NOT QUIET.** Two consecutive quiet rounds are still required, and none has happened yet.

lmeyerov:
## Convergence round 6 — **NOT QUIET**

Merge base is **`526976e91`** (`ghhttps/master`; the worktree's `origin/master` is ~stale and gives a
wrong base of `54fcba399` — every number below is against `526976e91`).

Round 5's audit did not hold either. Two of its four "provably-equivalent" survivors are not
equivalent, and the shape that breaks them is a **behaviour regression this PR introduces vs the
merge base**, on the one surface the PR is named after.

---

### 1. Re-deriving round 5's four "provably-equivalent" survivors

| survivor | round 5 | round 6 |
|---|---|---|
| gate forced on, no node table bound — **pandas `hop()`** | equivalent | **holds** |
| gate forced on, no node table bound — **polars `hop()`** | equivalent | **FALSE** |
| gate forced on, no node table bound — **polars chain fast path** | equivalent | **FALSE** |
| **polars lazy seed collect** removed | equivalent | **holds** |

The proof round 5 offered was "the synthesized table contains every endpoint id, so the gate is the
identity". Containing the id is not sufficient — **membership disagrees about NULL**:

```
pandas  pd.Series([1,None,3]).isin(pd.Series([1,None,3]))  -> [True, True,  True]
cudf    cudf.Series([1,None,3]).isin(...)                  -> [True, True,  True]
polars  df.filter(pl.col.x.is_in(<same ids>))              -> drops the NULL row
```

`is_in` answers NULL for a NULL input and `filter` drops a NULL predicate, so on polars the gate is
not the identity even over a universe that holds the NULL. Measured on `edges (0,1) (1,2) (NULL,2)`
with no node table bound, `hop(nodes=[0], hops=3, undirected)`: head **3 edges**, gate-forced-on
**2**. Same for the chain fast path, which reaches the question through a semi-JOIN (polars joins
never match NULL to NULL). So `node_table_bound` is load-bearing on both polars surfaces and nothing
was pinning it.

The two that hold, re-derived rather than accepted:
* **pandas** — `materialize_nodes` builds the universe from the *unfiltered* edge column pair, so it
  is a superset of the post-`edge_match` endpoints, and pandas `isin` matches NaN/None to itself.
  Identity.
* **polars lazy seed** — `hop_polars` collects the seed itself at `df_to_engine(nodes, POLARS)`, and
  the index path tolerates a `LazyFrame` (`maybe_index_hop`'s only `nodes.shape` read is inside a
  `try/except AttributeError`; `index_seeded_hop` really did receive a `LazyFrame` in my probe).
  Verified by a **2160-shape** head-vs-mutant sweep (5 topologies × indexed/not × hops × direction ×
  wavefront × tfp × seed sets, both eagerness): **0 differences**.

### 2. The regression that hid behind that wrong proof

The same NULL blindness is not just a pin gap — it **changes an answer versus the merge base**, and
only on polars. Node table `{0, 1, 2, NULL}`, edges `(0,1) (1,2) (NULL,2)` — a graph that is *closed*,
every endpoint including the NULL has a node row, so #1888's gate must remove nothing:

| `hop(nodes=[0], hops=3, undirected)` | pandas | polars | cuDF |
|---|---|---|---|
| merge base `526976e91` | 3 edges | 3 edges | 3 edges |
| head `e8d061a7d` | 3 edges | **2 edges** | 3 edges |

polars is the wrong side of the PR's own stated contract (it over-filters a closed graph), so I fixed
it rather than xfailing a regression: `_keep_edges_with_both_endpoints_resolvable` now treats a NULL
endpoint as resolvable exactly when the id universe holds a NULL. That restores base parity and
cross-engine agreement, and it makes the **polars `hop()`** gate-guard survivor genuinely equivalent —
the polars **chain** semi-join stays null-blind and its guard stays load-bearing, which the new chain
cell pins.

I did not touch the chain semi-joins: `nulls_equal`/`join_nulls` is version-sensitive across the
polars versions this repo runs, and on the *bound-table* chain shape head already agrees with pandas.
The nodes-less chain divergence (polars 3, pandas/cuDF 2) reproduces identically at `526976e91`, so it
is pre-existing and out of this PR — reported, not pinned.

### 3. Mutation audit

**12 mutations · 8 caught · 4 survived · 2 of the survivors were real gaps · 0 decorative pins.**

Round 5's own four new pin targets, re-mutated — all four still catch what they name, and each goes
red on the *right* arms (not via an unrelated assertion in the same body):

| mutation | red | which cells |
|---|---|---|
| backfill trailing de-dup deleted | 4 | only the two `..._under_to_fixed_point` cells, pandas + cuDF |
| undirected seed hop-label STRIP deleted | 4 | the 3 pandas `..._leave_the_seed_unlabelled` cells + the cuDF closure cell as an **XPASS** (proving that strict xfail really is strict) |
| polars chain reverse endpoint swap collapsed | 3 | only the 3 `reverse_*` polars cells of the endpoint-binding oracle |
| `_with_every_lazy_input_collected` nodes arm removed | 2 | the two `lazy_nodes=True` cells |

That covers **39 of the 84** mutation-justified cells. One redundancy found: the `[False-True]`
(lazy **edges** only) parameter adds nothing — removing that arm alone is already caught by the
pre-existing `test_engine_polars_hop.py::TestHopLazyInputEagernessMatrix::test_edges_only_lazy_graph_matches_pandas`.
Redundant, not decorative.

Also mutated this round: polars gate disabled (caught), `_ids_an_endpoint_may_resolve_to`'s
target-wavefront widening removed (caught by `test_endpoint_universe_is_widened_by_the_target_wavefront`),
and the null-aware fix reverted (caught by the new regression cell).

### 4. Round 5's anti-vacuity claim, verified

Its 4 changed test files run at `526976e91` (two import shims needed — `test_hop_kernel_contracts.py`
cannot even collect there, since it imports `_endpoint_ids_without_node_rows`, `_reached_node_ids`,
`_ids_an_endpoint_may_resolve_to`, `_keep_edges_with_both_endpoints_resolvable`; round 5 reported
those cells as "pass at merge-base" without noting they need a shim to be *runnable* there):

**exactly 7 cells red at merge-base, and all 7 are the `min_hops` axis** — `[2|3|4-*-pandas]`,
`[2|3|4-*-cudf]`, `[cudf-4]`. Claim confirmed as stated.

### 5. The two new strict xfails

Both are strict and both name a condition I reproduced at `526976e91`, so neither can start passing
silently:
* `_SEED_LABEL_CUDF_XFAIL` — measured at base *and* head: the labelled undirected hop returns nodes
  `[1, 2]` and edges `[(0,1), (1,2)]`, i.e. edge `(0,1)` with **no node row for 0**. Correctly narrowed
  with `raises=AssertionError`, and demonstrably strict (it XPASSed and reddened the run under the
  strip mutation).
* `_OM_AGG_CUDF_XFAIL` — measured at base *and* head, for `count` and `sum` alike:
  `GFQLTypeError [invalid-node-reference] Error executing 'group_by': The truth value of a Series is
  ambiguous`. One observation, not a blocker: it carries no `raises=`, so an unrelated failure would
  also be swallowed as an expected one.

### 6. Axis enumeration (round 5's "all intersections covered except min_hops")

I re-derived the axis list from the true diff instead of accepting it. Every changed product file
classified by AST-after-docstring-strip: `gfql_fast_paths.py`, `polars/dtypes.py`, `filter_by_dict.py`
and `layout/gib/partitioned_layout.py` are **comments-only**; `index/traverse.py` is comment removal
plus one variable extraction (**behaviour-neutral** — so round 5's "out of diff" disposition is right,
though its stated reason, "the index lane cannot run on this box", is wrong: the CPU polars index lane
runs here fine, `gfql_index_edges` + `index_seeded_hop` both engage — only the cupy/GPU index is
blocked); `nan_clean.py` and `row/pipeline.py` are cache-key renames; `gfql_unified.py` is predicate
extraction already pinned in round 3; `typing.py` adds a `NodeId` alias that **nothing imports**.
`only_seeds_can_be_sources` is already at `526976e91` and is not this PR's.

The real semantic surface is `hop.py` + the three polars files. Bound-vs-synthesized, present-vs-absent
endpoint, direction, seed cardinality, hop windows (incl. traversal `min_hops`), labelling + seed strip,
lazy/eager inputs, chain endpoint binding and target-wavefront widening all have both-sides pins.
**Endpoint id NULL-vs-non-null did not** — that is the new axis this round adds, and it is where the
regression lived.

### 7. New cells + anti-vacuity

**9 new cells** (7 runnable here; 2 `polars-gpu` skip with a named reason — `cudf_polars` absent).
**0 are red at the merge base**, and I am not calling any of them base-red:

* `test_a_null_endpoint_backed_by_a_null_node_row_survives_the_gate` — **regression pin**. The
  `[polars]` arm is **RED at the pre-fix head `e8d061a7d`** and green at base and after the fix. The
  `[pandas]`/`[cudf]` arms are declared **oracles/controls**: they hold at base, at head and under
  every mutation, and they are what makes the polars divergence legible.
* `test_a_null_endpoint_survives_the_vacuously_closed_synthesized_table` — **mutation-justified**:
  red on the pre-fix head under gate-forced-on. After the fix that mutation is genuinely equivalent,
  so this cell is now a behaviour pin only. Saying so rather than claiming a kill it no longer makes.
* `test_the_synthesized_table_does_not_gate_a_null_endpoint_on_the_polars_chain` — **mutation-justified
  and still load-bearing after the fix**: forcing the chain gate on drops `(NULL,2)`.

Every expectation is hand-walked from the fixture edge list (`0 -(0,1)- 1 -(1,2)- 2 -(NULL,2)- NULL`,
so a 3-hop undirected walk from 0 traverses all three edges) and written as a literal. Cross-engine
agreement is not the oracle anywhere — the pandas/cuDF arms assert the same hand-computed literal the
polars arm does.

### 8. Gates

* `ruff check graphistry bin` — clean.
* `ci_comment_density_guard.py`, `ci_type_hygiene_guard.py`, `ci_cypher_surface_guard.py` — all run
  **from inside the worktree**, all OK, no baseline raised.
* `mypy graphistry` — 4 errors at head, the **same 4** at `526976e91` (line numbers shifted only).
* `test_polars_lane_completeness.py` — 11 passed. No new test file and no new source file, so no lane
  registration and no coverage-baseline entry is due.
* Full `graphistry/tests/compute` on this GPU box (cuDF 25.10 really executing): head `e8d061a7d`
  **102 failed, 10264 passed, 952 skipped, 62 xfailed**; this round's head `21167e08f`
  **102 failed, 10271 passed, 954 skipped, 62 xfailed**. **Failure sets are byte-identical**
  (`comm` empty in both directions); +7 passed = the new runnable cells, +2 skipped = their
  `polars-gpu` arms. Both runs printed their summary line.
* `mergeable: MERGEABLE`, 23 check-runs created on `21167e08f`.

**Environment scoping** (declared, not silently skipped): `polars-gpu` cells skip — `cudf_polars` is
absent from this image. The cupy-backed GPU index kernels cannot run here (`libnvrtc.so.12`), so no
claim above rests on them; the **CPU** polars/pandas index lane *does* run and I used it (it is how
the lazy-seed survivor was cleared through `index_seeded_hop`).

### Verdict: **NOT QUIET**

A real answer changed versus the merge base on polars, unpinned, and two of round 5's equivalence
proofs were wrong. Round 6 is not quiet, so two consecutive quiet rounds are still outstanding and
this cannot be the first of them.

lmeyerov:
## Convergence round 7 — **NOT QUIET**, but **no blocking defect found**

Merge base **`526976e91`** (via `ghhttps/master`). Round-7 head **`0a1e8a4f1`** (was `21167e08f`).
Answering the narrowed question as well as the convergence one.

### Category 1 — defects INTRODUCED by this PR (correct at base, wrong at head)

**None that is a new wrong-answer class.** One value change vs base, stated in full so it can be
overruled rather than buried:

Fixture NULLEP — nodes `{0,1,2,NULL}`, edges `(0,1) (1,2) (NULL,2)`; closed end to end, every
endpoint including the NULL has a node row.

| `gfql([n(), e_undirected(), n()])` | pandas | polars | cuDF |
|---|---|---|---|
| merge base `526976e91` | 2 edges | **3 edges** | 2 edges |
| head | 2 edges | **2 edges** | 2 edges |

The polars chain answer moved 3 → 2 because #1888 attached the endpoint gate to the chain fast
path and a polars semi-join never matches NULL to NULL. Why I do not call this a blocker:
at base the polars chain had **no** gate at all — it returned every edge verbatim, which is the
bug #1888 exists to fix (`CHANGELOG`: "polars unconstrained chains returned all edges verbatim").
The base `3` was the ungated pass-through, not a rule producing the right answer; on the same
query pandas and cuDF answered `2` at base and answer `2` now. So head **removes** an engine
divergence rather than creating one, and creates no answer that some engine did not already give.
The residual wrongness — the chain surface answering `2` where `hop()` answers `3` — is
pre-existing on pandas/cuDF (category 2 below). Pinned so it can never move silently:
`test_the_chain_answers_the_same_null_endpoint_question_as_hop`, a strict xfail whose polars arm
**XPASSes at `526976e91`**, i.e. it is red at the merge base and encodes exactly this delta.

Sweep behind that verdict (both sides at base and head, pandas / polars / cuDF): NULL in the id
column, in `src`, in `dst`, both endpoints NULL, NULL-only node table, node table *without* the
NULL, NULL seeds, mixed seeds, forward / reverse / undirected, `to_fixed_point`, `min_hops`,
`output_min_hops`, `label_node_hops`, `label_seeds`, `target_wave_front`, var-length, the
synthesized (vacuously closed) table, Cypher rows and `count(*)`, the #1658 index route, and
string ids as well as float. **The only base→head answer changes anywhere in that sweep are the
polars chain gate above (5 shapes, all polars-only, all converging onto the pandas/cuDF answer)
and the polars hop gate correctly starting to drop `(NULL,2)` when the node table has no NULL
row — which is the PR working.**

### Category 2 — pre-existing (wrong at base too). Not fixed here; pinned.

All three reproduce identically at `526976e91` and are now strict xfails naming the real
condition, in `test_endpoint_closure_matrix.py`:

1. **polars hop keeps the NULL-endpoint edge but drops its node row.** `hop(nodes=[0], hops=3,
   undirected)` returns edge `(NULL,2)` with node set `{0,1,2}` — the output is *not*
   endpoint-closed. Round 6's regression pin asserts edges only, so it was blind to this. Cause:
   the node epilogue `all_nodes.join(needed, how="semi")` is the same null-blind membership as the
   gate, one site further on. pandas/cuDF return `{0,1,2,NULL}`.
2. **polars `MATCH (a)-[x]-(b) RETURN count(*)` answers 4 where pandas/cuDF answer 6** — the two
   orientations of the NULL edge, lost in the chain's endpoint semi-joins. (Control: the same query
   over a NULL-free 3-edge graph answers 6 on all three engines, so 6 is the shape's oracle.)
3. **pandas/cuDF gate a NULL endpoint out of a SYNTHESIZED node table**, answering 2 where polars
   answers the contract's 3 — the vacuously-closed case the matrix header says must not be gated.
   Round 6 reported this in a docstring; it is now a cell.

Reported, not pinned (no new class, uniform across engines, base-identical): an unseeded FORWARD
hop drops a NULL-**source** edge on all three engines while the undirected hop keeps it — the
edge survives undirected only via its other orientation; and cuDF drops the seed's node row under
`min_hops` / `label_node_hops`, which reproduces with **no NULL in the graph at all** and is
already the class round 5 pinned as `_SEED_LABEL_CUDF_XFAIL`.

### Round 6's claims, re-derived rather than accepted

| claim | round 6 | round 7 |
|---|---|---|
| survivor: gate forced on, no node table — **pandas `hop()`** | holds | **holds** |
| survivor: **polars lazy seed collect** removed | holds | **holds** |
| its null-aware fix is pinned by a regression cell | 1 regression pin + 2 mutation-justified cells | **the pin dies as claimed — but only 1 of the 2 cells; the other is a CONTROL** |
| polars **chain** left null-blind on purpose | reported, not pinned | **now pinned; see category 1/2** |

* **pandas gate-forced-on** — `node_table_bound = True`: whole `graphistry/tests/compute` run,
  `102 failed, 10271 passed, 954 skipped, 62 xfailed`, failure set **byte-identical** to head
  (`comm` empty both ways). Mechanism re-derived: with no node table bound, `materialize_nodes`
  builds the universe from the **unfiltered** endpoint pair (the `edge_match` filter is applied
  after), and `pd.Series([1.,nan]).isin(...)` → `[True, True]`, so the gate is the identity.
* **polars lazy seed collect** — line deleted: whole compute suite, `102 failed, 10271 passed`,
  again byte-identical. Mechanism: `df_to_engine(LazyFrame, Engine.POLARS)` returns an **eager**
  `DataFrame`, so `hop_polars` collects the seed itself; the collect in `hop_lazy_or_eager` is
  redundant (defensive for the index hook that runs before it), not load-bearing.
* **the null-aware fix reverted** — `test_a_null_endpoint_backed_by_a_null_node_row_survives_the_gate[polars]`
  goes red, and nothing else does: `1 failed, 302 passed, 91 skipped, 2 xfailed`. Right cell,
  right reason. But `..._survives_the_vacuously_closed_synthesized_table` **does not** die, and
  it does not die under `node_table_bound = True` either: forcing the polars hop gate on now
  leaves the whole gfql suite byte-identical (**93 = 93**), because round 6's own fix made that
  mutation equivalent (the synthesized universe holds the NULL, and a NULL endpoint is now
  resolvable when it does). That cell is relabelled as the control it is — it was a pin at the
  pre-fix head and stopped being one in the same commit.

### Mutation audit

**5 mutations · 2 caught · 3 survived · 2 of the survivors carry proofs · 1 reclassified a cell.**

| mutation | verdict | evidence |
|---|---|---|
| pandas `hop()` gate forced on | survived — **equivalent** (proof above) | full compute suite, 102=102 identical |
| polars lazy seed collect removed | survived — **equivalent** (proof above) | full compute suite, 102=102 identical |
| polars `hop()` gate forced on | survived — **newly equivalent**, was a pin pre-fix | full gfql suite, 93=93 identical |
| polars **chain** gate forced on | **caught**, right cell only | `1 failed, 1320 passed` — `test_the_synthesized_table_does_not_gate_a_null_endpoint_on_the_polars_chain` |
| round 6's null-aware fix reverted | **caught**, right cell only | `1 failed, 302 passed` |

Every line above is a pytest summary line from a real run; no run without one was counted.

### New cells + anti-vacuity

**14 new cells**, 0 removed. **11 runnable here**; 3 `polars-gpu` params skip with a named reason
(`cudf_polars` absent). At head: 7 xfail, 4 green, 3 skip. **1 is red at the merge base** — the
`chain[polars]` strict XPASS, which is the category-1 delta. The 4 green cells are the **declared
pandas/cuDF oracle arms** of the polars xfails beside them (they make the divergence one cell
apart instead of one file apart); they are not claimed as pins, in line with round 5's
decorative-pin finding. Redundancy avoided deliberately: the synthesized-chain cell is
parametrized over pandas/cuDF only, since its polars arm would duplicate the round-6 cell that
the chain-gate mutation already catches.

### Gates

- `ruff check graphistry bin` — clean
- comment-density / cypher-surface / type-hygiene guards — all **exit 0**, run from inside this
  worktree; **no baseline updated** (both offer `--update-baseline` improvements; declined)
- `mypy graphistry` — **4 errors at `526976e91`, the same 4 at head**, no new
- polars lane: `test_endpoint_closure_matrix.py` already registered in `bin/test-polars.sh:61`;
  `test_polars_lane_completeness.py` 11 passed. No new source file, so no coverage baseline moves.
- Full `graphistry/tests/compute` at head: **`102 failed, 10275 passed, 957 skipped, 69 xfailed`**;
  failure set **byte-identical** to the pre-round-7 head (`comm` empty both ways). Deltas are
  exactly the 14 new cells (+4 passed, +7 xfailed, +3 skipped).
- Environment: pandas 2.3.3, polars 1.42.0, cuDF 25.10 (really executes). `polars-gpu` and the
  cupy-backed index kernels **cannot run here** (`libnvrtc.so.12` missing) and are scoped off.

**Verdict: NOT QUIET on the convergence bar — no category-1 blocker on the shipping bar.** The
NULL axis is a real live defect class, but the part of it this PR owns is fixed, and everything
else I could reproduce is pre-existing and now pinned rather than silent.

Issue #1896, opened 2026-08-14
GFQL Cypher: OPTIONAL MATCH + WITH pipeline row semantics (silent-wrong umbrella)


lmeyerov:
Post-fix adversarial re-probe of the OM subsystem (stack tip b645af615, hand-computed openCypher oracles, both engines, twice-run stable) found 4 silent-wrong shapes in the OM→WITH pipeline. Fixture: nodes a1-a4:P v=1..4, b1,b2:C v=10,20, z; edges a1-KNOWS->b1, a1-KNOWS->b2, a2-LIKES->b1, b1-KNOWS->a3.

**1. CRITICAL (cross-engine divergent; polars fabricates data):** `MATCH (a:P) OPTIONAL MATCH (a)-->(b) WITH a, b WHERE a.v <= 2 RETURN a.id AS aid, b.id AS bid` — oracle (a1,b1),(a1,b2),(a2,b1); pandas gives (a1,None),(a2,None); polars gives (a1,'a1'),(a2,'a2') — bid filled with SEED ids. A b-referencing WHERE correctly declines typed (#1273 guard, lowering.py:2228); a seed-only predicate slips past onto a single-alias projection path.

**2. HIGH:** `... WITH a.id AS aid, count(b) AS cnt RETURN aid, cnt` — zero-count groups dropped ((a3,0),(a4,0) missing; direct `RETURN a.id, count(b)` is correct). Breaks `WHERE cnt = 0` ("nodes with no X") silently.

**3. HIGH:** impure-carry reentry null-fill loses seed identity: `MATCH (a:P) WITH a AS p OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p.id, b.id` — 3 unmatched seeds become 2 ANONYMOUS all-null rows. Root: graphistry/compute/gfql/cypher/reentry/execution.py:136-141 fills `prefix_rows − result_rows` null-template copies instead of anti-joining unmatched prefix rows.

**4. HIGH:** `WITH a ORDER BY a.v LIMIT 2 OPTIONAL MATCH ...` — unmatched seed a2's null row never synthesized (`result_rows >= prefix_rows` short-circuit, same code).

Also: ORDER BY on optional-alias property declines typed though the output-alias spelling works with correct null placement — error message should hint the workaround.

[...]

Issue #1341, opened 2026-05-07
Cypher MATCH after WITH rejects re-binding a carried secondary alias as a node variable in trailing MATCH


lmeyerov:
## Summary

Cypher queries that carry a secondary alias through a `WITH` and then bind it as a node variable in a trailing `MATCH` are rejected by the GFQL lowering. This blocks the LDBC SNB IC1 shortestPath shape and similar re-entry patterns.

## Repro (minimal, fails on master `41d865ac9`)

```cypher
MATCH (p:Person {id: $pid}), (friend:Person)
WHERE NOT p = friend
WITH p, friend
MATCH path = shortestPath((p)-[:KNOWS*1..3]-(friend))
RETURN friend.id AS friendId, length(path) AS dist
```

## Expected vs observed

- **Expected**: shortestPath is computed with `friend` already constrained from the prefix `WITH`.
- **Observed**: `GFQLValidationError: [unsupported-cypher-query] Cypher MATCH after WITH does not yet support re-binding a carried secondary alias as a node variable in the trailing MATCH | field: match | value: ['friend']`

## Source

- Guard: `graphistry/compute/gfql/cypher/lowering.py:7181`
- Asserting test: `graphistry/tests/compute/gfql/cypher/test_lowering.py:9082` (`pytest.raises(... match="re-binding a carried secondary alias")`)

## Benchmark evidence

- `pyg-bench` artifact: `results/runs/dgx-spark-ic1-conformance-2026-04-05/probe-results.json`
- LDBC SNB lane: `interactive-complex-1` / `named-shortest-friends`
- Probe `semantic_scope: adapter_shortest_path_join_workaround` — adapter currently splits this into a workaround pass because the direct shape is rejected here.
- Official query at `third_party/ldbc_snb_interactive_v1_impls/cypher/queries/interactive-complex-1.cypher` uses exactly this `WITH p, friend MATCH path = shortestPath((p)-[:KNOWS*1..3]-(friend))` shape.

## Acceptance criteria

- The minimal repro above succeeds end-to-end with a row per `(p, friend)` pair carrying the shortestPath length.
- The secondary-alias rebinding guard at `lowering.py:7181` is retired or scoped down (e.g., admit when the rebound alias is the trailing pattern endpoint of a `shortestPath()` / regular-pattern MATCH).
- At least one regression test under `graphistry/tests/compute/gfql/cypher/` exercises the IC1 shape directly.
[...]

PR #1897, opened 2026-08-14
fix(gfql): OPTIONAL MATCH + WITH pipeline row semantics (#1896)

lmeyerov:
Stacked on #1895. Fixes all four silent-wrong shapes from the #1896 re-probe umbrella:

1. **CRITICAL guard bypass** — terminal WITH after OM rode the single-alias row pipeline (pandas: nulled+collapsed; polars: fabricated seed ids into the optional column). New `flatten_terminal_with_over_optional` routes pure bare-alias carries onto the connected left-join lowering with the stage WHERE as a **post-join binding-row filter** — exact openCypher row semantics, which also makes the canonical `WITH p, x WHERE x IS NULL` anti-join answer correctly on both engines.
2. **Zero-count groups** — passthrough WITH-aggregates fold to direct RETURN aggregates (null-keeping group keys; count(b)=0 preserved).
3+4. **Anonymous null fill** — `apply_optional_reentry_null_fill` count-arithmetic replaced by identity-carrying anti-join of unmatched prefix rows (rename + LIMIT carries fixed); no-identity shapes decline typed, never anonymous rows.
5. ORDER BY decline messages now hint the working output-alias spelling.

21 new green pins (section F, hand-computed oracles, both engines) + 2 decline-shape negative controls. Gates: OM 81P, divergences 11P/3xf, hop 42P, flatten 25P, lowering 1420P (0 new failures), polars 587P, TCK 4145P/0F, ruff+guards green (surface baseline regen for intentional lowering growth 9454→9503).


lmeyerov:
**GPU sweep receipt (dgx GB10, RAPIDS 26.02, --gpus-enforced, @ d502dea58):** full `graphistry/tests/compute/gfql` tree — **8820 passed, 53 skipped, 22 xfailed, 0 failed** (4m40s). First zero-failure GPU sweep of the release campaign; prior tip was 1 failed / 8773 passed. Sanity probes (cudf sum + string to_pandas, polars GPUEngine raise_on_fail collect) green. Minor follow-up noted, not a failure: pandas `FutureWarning` on all-NA concat at `reentry/execution.py:180` in the null-fill path.

graphistry/compute/gfql/cypher/reentry/execution.py
```diff
@@ -133,12 +139,36 @@ def apply_optional_reentry_null_fill(
         null_row=null_row,
         reentry_plan=reentry_plan,
     )
+    if fill_rows is None and carried_output_map:
+        # #1896: never synthesize ANONYMOUS null rows by count arithmetic --
+        # anti-join unmatched prefix rows so the null-extension keeps its
+        # carried seed identity, or decline typed below.
```
lmeyerov: Violates comment rules too

graphistry/compute/gfql/cypher/reentry/execution.py
```diff
@@ -152,6 +182,48 @@ def apply_optional_reentry_null_fill(
     )
 
 
+def _optional_reentry_identity_null_rows(
+    *,
+    prefix_df: DataFrameT,
+    result_df: Optional[DataFrameT],
+    null_row: Dict[str, Any],  # hygiene-ok: explicit-any -- heterogeneous Cypher fill values, mirrors _optional_reentry_carried_null_rows
```
lmeyerov: Avoid Any when narrow available

graphistry/compute/gfql/cypher/reentry/flatten.py
```diff
@@ -283,3 +284,124 @@ def flatten_carried_endpoint_rebind(query: CypherQuery) -> Optional[CypherQuery]
...
+      MATCH filters rows -- matched rows keep their bindings, null-extended
+      rows pass or fail on their own values).
+    - Terminal projection/aggregate stage passed through unchanged by RETURN
+      (``WITH a.id AS aid, count(b) AS cnt RETURN aid, cnt``): fold the stage
+      into RETURN so the direct aggregate lowering (null-keeping group keys)
+      serves it.
+
+    Disqualified shapes return None and stay on the (typed-decline) pipeline:
+    renames/DISTINCT/ORDER/SKIP/LIMIT on the stage, multiple stages, UNWIND/
+    CALL, references to non-carried aliases after a subset carry.
```
lmeyerov: More comment issues

lmeyerov:
Review addressed, with the self-review gate run over the whole PR diff rather than just the flagged lines.

## The three sites
1. **`execution.py:145`** — the 3-line `# #1896: never synthesize ANONYMOUS null rows…` was narration *and* why-a-fix-was-made, so a BLOCKER twice over. Deleted. The rule now lives in the helper name `_optional_reentry_unmatched_identity_null_rows` and **in the type**: the `Optional[Mapping[str,str]]` tri-state (None = decline / `{}` = nothing to do / map = fill) became a frozen `CarriedOutputSources` plus `CARRIED_OUTPUTS_NOT_REPRODUCIBLE`, which also retired a 4-line parameter docstring.
2. **`execution.py:189`** — `Dict[str, Any]` with two `hygiene-ok` pins on brand-new code. Both pins **removed, not suppressed**; real aliases added (`CypherScalar`, `CypherEmptyGroupValue`, `CypherFillValue`, `CypherFillRow`), and five sibling helpers in the file narrowed with them. #1894's commit is not on this branch, so its aliases were copied **byte-identically** so the two merge cleanly.
3. **`flatten.py:311`** — a 20-line docstring narrating the algorithm and re-listing guards the code already states. Now 2 lines plus six named helpers (`_is_terminal_with_over_optional_match`, `_stage_reshapes_rows`, `_query_without_pure_carry_stage`, `_query_with_terminal_stage_folded_into_return`, …). Also DRY'd a copy-paste: `_pure_carry_aliases_ignoring_where` was `_pure_carry_aliases` minus one guard; both now share `_bare_carry_aliases`.

## Self-review gate, whole diff
18 sites given verdicts — including **12 `#` labels inside `parametrize`** converted to `pytest.param(id=…)` and 3 per-test explanations folded into test names (e.g. `test_declines_pure_carry_when_order_by_references_an_uncarried_alias`). Afterwards `grep '^\+\s*#'` returns only the deliberate keeps, and `grep '^\+.*(Any|type: ignore|hygiene-ok|cast\()'` returns **nothing**.

**Kept, with reasons:** `#:` Sphinx doc-comments on the six type aliases (each states an openCypher spec fact no signature can express — the form accepted on #1894); the `TYPE_CHECKING`-only import-cycle note in `agg_types.py`, kept byte-identical to #1894's so the branches merge without conflict; and the test-file section banner matching the existing A–E convention.

## Pins added where a comment was the only record
Deleting a comment without an equivalent pin is information loss, so three went in:
- `test_admitted_query_never_retains_a_with_stage_so_recompiling_it_terminates` — 4 admitted shapes, asserts `with_stages == ()` *and* that re-flattening declines (the recursion-termination claim the deleted comment made).
- New `test_optional_reentry_carried_outputs.py` — 7 named params covering **every** branch of the carried-output mapping, including `p.__cypher_reentry_av__` and the uncarried-marker fallthrough, which had **no** pin at all before.
- `test_optional_match_anti_join_with_where_x_is_null_keeps_only_unmatched_rows`, with the hand-computed oracle in its docstring.

## Gates
Failure list md5 `d68f3d7cddf5bab5d11588e04727d362` — **identical across base, PR head, and post-fix**. Targeted suites 1538 passed; cypher dir 7 failed (all cudf on a GPU-less box, inside the identical 93); chain+hops byte-identical to base. ruff clean; mypy 328 files clean.

**Hygiene improved and ratcheted down**, not held flat: `explicit-any` in `reentry/execution.py` **9 → 0** (baseline entry deleted) and `gfql_unified.py` 21 → 20; total grandfathered 4496 → 4486. The cypher-surface baseline also ratcheted 9503 → 9493 as `lowering.py` shrank.

One judgement call flagged for you: that surface ratchet tightens within the bump this PR itself made (9454 → 9503 → 9493) rather than leaving 10 lines of slack.

graphistry/compute/gfql/cypher/reentry/execution.py
```diff
@@ -133,12 +156,34 @@ def apply_optional_reentry_null_fill(
...
     if fill_rows is None:
-        if result_rows >= prefix_rows:
+        single_prefix_row_matched = prefix_rows == 1 and result_rows > 0
+        if single_prefix_row_matched:
             return result
-        missing_count = prefix_rows - result_rows
-        fill_rows = [dict(null_row) for _ in range(missing_count)]
-    elif not fill_rows:
+        if not carried_outputs.every_output_reproducible:
+            raise reentry_validation_error(
```
lmeyerov, line 168: tck-gfql obligations likely, audit & handle as needed

graphistry/tests/compute/gfql/cypher/reentry/test_flatten_terminal_with.py
```diff
@@ -0,0 +1,150 @@
+"""Direct tests for terminal-WITH-over-OPTIONAL flattening (#1896).
```
lmeyerov, line 1: shouldn't these be more like cypher/reentry/test_flatten.py instead of toplevel?

graphistry/compute/gfql/cypher/reentry/flatten.py
```diff
@@ -283,3 +292,117 @@ def flatten_carried_endpoint_rebind(query: CypherQuery) -> Optional[CypherQuery]
...
+        downstream_texts.extend(item.expression.text for item in query.order_by.items)
+    if stage.where is not None:
+        downstream_texts.append(stage.where.text)
+    for text in downstream_texts:
+        if not _referenced_aliases(text, match_aliases) <= carried:
+            return None
+    return replace(query, with_stages=()), stage.where
+
+
+def _query_with_terminal_stage_folded_into_return(
```
lmeyerov, line 353: seems significant positive/negative test obligations

graphistry/compute/gfql/cypher/ast.py
```diff
@@ -26,7 +26,10 @@ class ParameterRef:
     span: SourceSpan
 
 
-CypherLiteral = Union[None, bool, int, float, str, ParameterRef]
+#: Every value openCypher can write literally: the four primitive types plus null.
+CypherScalar = Union[None, bool, int, float, str]
```
lmeyerov, line 30: is this from lower on our pr stack?

graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -9139,6 +9152,32 @@ def _attach_graph_context(result: CompiledCypherQuery) -> CompiledCypherQuery:
         )
         if compiled_connected_optional is not None:
             return _attach_graph_context(compiled_connected_optional)
+    if query.with_stages and any(m.optional for m in query.matches):
```
lmeyerov, line 9152: more significant pos/neg test amplifications obligations

graphistry/compute/gfql/agg_types.py
```diff
@@ -43,20 +43,33 @@
...
+from typing import TYPE_CHECKING, Final, FrozenSet, List, Mapping, NoReturn, Optional, Union
 
 from graphistry.compute.exceptions import ErrorCode, GFQLTypeError
 
 if TYPE_CHECKING:
     import polars as pl
 
+    # TYPE_CHECKING only: `gfql.cypher.__init__` eagerly imports the whole compiler, and
+    # this module is imported from inside that import (row.pipeline), so a runtime import
+    # would be circular. The aliases below only need the name at type-check time.
```
lmeyerov: avoid unnecessary and overly verbote comments... root cause why this happened, and both fix here, but all the root cause of why authoring & reviewing let this through

graphistry/compute/gfql_unified.py
```diff
@@ -1467,12 +1471,83 @@ def _execute_compiled_query_with_reentry(
...
             aggregate_fill_values=_optional_reentry_aggregate_fill_values(compiled_query),
+            carried_outputs=_carried_output_sources(compiled_query),
         )
 
     return result
 
 
-def _optional_reentry_aggregate_fill_values(compiled_query: CompiledCypherQuery) -> Dict[str, Any]:
+_IDENTIFIER_TOKEN = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
+_REENTRY_MARKER_COLUMN = re.compile(r"__cypher_reentry_(\w+)__")
```
lmeyerov: strange location for these symbols, shouldn't they be externed for reuse by others?

graphistry/compute/gfql_unified.py
```diff
@@ -1467,12 +1471,83 @@ def _execute_compiled_query_with_reentry(
...
+        return None
+    marker = _REENTRY_MARKER_COLUMN.fullmatch(parts[1])
+    if marker is not None and marker.group(1) in scalar_columns:
+        return marker.group(1)
+    if _IDENTIFIER_TOKEN.fullmatch(parts[1]):
+        return src
+    return None
+
+
+def _carried_output_sources(compiled_query: CompiledCypherQuery) -> CarriedOutputSources:
```
lmeyerov: gfql_unified.py is already overly large..

graphistry/compute/gfql_unified.py
```diff
@@ -1467,12 +1471,83 @@ def _execute_compiled_query_with_reentry(
...
+    if plan is None:
+        return CarriedOutputSources(columns={}, every_output_reproducible=True)
+    alias_names = {plan.reentry_alias_name} | {a.output_name for a in plan.aliases}
+    scalar_columns = set(plan.scalar_columns)
+    items: Optional[List[object]] = None
+    grouped = False
+    ops = list(compiled_query.chain.chain) if compiled_query.chain is not None else []
+    for op in ops:
+        function = getattr(op, "function", None)
+        op_params = getattr(op, "params", None) or {}
```
lmeyerov: try to replace dynamic typing pattersn like getattr with more structured & static programming

graphistry/compute/gfql_unified.py
lmeyerov: test amplification concerns?

lmeyerov:
Answering the factual question first, then the rest is dispatched.

**"is this from lower on our pr stack?" (`ast.py:30`)** — **Yes, and deliberately byte-identical.** `CypherScalar` and the `CypherLiteral` re-expression are the same lines #1894 introduced (verified: `git show` on both branches returns identical text). #1894's commit is not an ancestor of this branch, so rather than invent a parallel alias the lane copied it verbatim so the two merge without conflict. Same for the `TYPE_CHECKING` note in `agg_types.py`. That is also why you are seeing the **same** `agg_types.py:55` comment on both #1894 and here — one duplication, two PRs. It is being deleted in both, not just here.

**"gfql_unified.py is already overly large"** — confirmed and now tracked. It is **2,526 lines**; `lowering.py` is 9,493 (#1921). Both are the same problem, and I have added `gfql_unified.py` to that refactor issue rather than opening a competing one. Note the immediate consequence for this PR: your point about the fill-value symbols being in a "strange location" and the file being oversized are the same observation — those symbols belong in an importable home (`gfql/agg_types.py` already exists for exactly this), not inline here.

**Dispatched for fixing**, grouped:
- *Encoding*: `agg_types.py:55` verbose comment (with the root-cause you asked for — see #1924, which now carries seven causes and the corrected keep bar: "a justified keep is still a keep; the test is whether a name or a test can carry it").
- *Structure*: extern the fill-value symbols for reuse; test file moved to mirror the source tree (`cypher/reentry/test_flatten.py` rather than top-level) — you are right that a test's location should follow the module it pins.
- *Typing*: the remaining `getattr` probes replaced with static dispatch on the AST types.
- *Testing obligations* (three separate comments — `flatten.py:367`, `lowering.py:9155`, `gfql_unified.py:1508`): a positive/negative matrix across both engines for the flatten admit/decline branches and the terminal-WITH folding, with hand-computed oracles and an anti-vacuity count at this PR's base.
- *tck-gfql obligations* (`execution.py:171`): audited — this PR's branch has a companion tck-gfql branch carrying the promotion cascade, and I will report explicitly whether any scenario's outcome changes and whether the contract/manifest needs regenerating, rather than assuming it does not.

**Also worth flagging from your earlier note**, since it applies here: the surface-guard ratchet on this PR tightened *within* its own bump (9454 → 9503 → 9493), leaving no slack. Say the word if you would rather it left headroom.

graphistry/compute/gfql/cypher/reentry/execution.py
```diff
@@ -34,6 +37,26 @@
...
+#: Output column -> its value in a synthesized null-extended row.
+CypherFillRow = Dict[str, CypherFillValue]
+
+
+@dataclass(frozen=True)
+class CarriedOutputSources:
+    """Prefix-frame column behind each result output that reads the carried alias."""
+
+    columns: Mapping[str, str]
+    every_output_reproducible: bool
```
lmeyerov: are any of these exernally needed types?

graphistry/compute/gfql/cypher/reentry/flatten.py
```diff
@@ -22,31 +22,31 @@
...
+    ReturnClause,
+    ReturnItem,
 )
 
 
 _BARE_IDENT = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+_IDENT_TOKEN = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
+_AGGREGATE_CALL = re.compile(
+    r"\b(count|sum|avg|min|max|collect|stdev|percentile\w*)\s*\(", re.IGNORECASE
+)
```
lmeyerov, line 44: should these be externed and/or swapped to predicates?

graphistry/compute/gfql/agg_types.py
```diff
@@ -43,20 +43,33 @@
...
 
 
 #: Aggregates Cypher restricts to ``INTEGER | FLOAT | DURATION``. ``mean`` is GFQL's internal
 #: spelling of ``avg`` (see ``GFQL_GROUPBY_AGG_METHODS``), so both names must gate.
 GFQL_NUMERIC_ONLY_AGGREGATIONS: Final[FrozenSet[str]] = frozenset({"sum", "avg", "mean"})
 
+#: What an aggregate answers for an EMPTY group, per the Cypher contract documented above:
+#: ``count``/``sum`` -> ``0``, ``count(*)`` over a synthesized null-extended row -> ``1``,
+#: ``collect`` -> ``[]``. Every other aggregate's empty-group answer is ``null`` and is
+#: therefore absent from a fill map rather than present as a value.
```
lmeyerov: overly verbose..

lmeyerov:
Two things I promised in the dispatch comment above were **not actually delivered** at `fcd46ace5`, and one risk had never been checked. Both are closed now, and I re-audited all 16 threads against the current head rather than assuming the earlier write-up.

## 1. The test relocation was one level short

You asked for `cypher/reentry/test_flatten.py` rather than top-level. The files had only reached `graphistry/tests/compute/gfql/cypher/` — there was no `reentry/` directory at all. `git mv`'d, so history follows:

| old | new | tests |
|---|---|---:|
| `tests/compute/gfql/cypher/test_flatten_carried_endpoint_rebind.py` | `tests/compute/gfql/cypher/reentry/test_flatten_carried_endpoint_rebind.py` | 23 |
| `tests/compute/gfql/cypher/test_flatten_pure_carry_optional.py` | `tests/compute/gfql/cypher/reentry/test_flatten_pure_carry_optional.py` | 26 |
| `tests/compute/gfql/cypher/test_flatten_terminal_with.py` | `tests/compute/gfql/cypher/reentry/test_flatten_terminal_with.py` | 24 |
| `tests/compute/gfql/cypher/test_optional_reentry_carried_outputs.py` | `tests/compute/gfql/cypher/reentry/test_carried_outputs.py` | 7 |

**80 before, 80 after** (23/26/24/7 each side, checked per file, not just the total).

The fourth one is the interesting case, and it is why I checked what each file *pins* instead of moving on filename: `test_optional_reentry_carried_outputs.py` did **not** pin a `cypher/reentry/*` module at all — it imported `graphistry.compute.gfql_unified._carried_output_source_column`. Moving it under `reentry/` would have been the wrong location for the code as it stood. Your #1508/#1515 comments are the same observation from the other end, so I moved the **code** instead (below) and then the test followed it honestly, renamed to mirror its module.

**No `__init__.py`** — checked first: the sibling `cypher/` and `row/` test dirs have none (`index/`, `layouts/`, `same_path/` do), so the new dir matches its immediate parent.

**References swept** (`grep -rn` over CI configs, `bin/*.sh`, `docker/*.sh`, `pyproject.toml`, DEVELOP.md, the guard baselines and `test_polars_lane_completeness.py`): **none of the four files was registered anywhere.** Specifically —
- `test_polars_lane_completeness.py` keys on modules that *mention polars*; all four files contain zero occurrences of the string, so none was in the lane list or the `NOT_POLARS_GATED` dict.
- `.github/workflows/ci.yml` path filters use the **prefix** regex `^graphistry/tests/compute/gfql/cypher/`, which `cypher/reentry/...` still matches — no CI edit needed.
- The polars lane still carries the `#1896` execution tests via `test_optional_match_semantics.py`, which did not move.

## 2. The comment-density guard: it did fail, and the fix was not a cap

You were right that this was live, not hypothetical. Merging `master` (`526976e91`) — which brings #1927, #1924, #1926, #1894 — the guard failed on **two** files:

```
comment-block:    graphistry/compute/hop.py                8 findings, baseline 6
issue-rationale:  .../cypher/reentry/flatten.py            2 findings, baseline 1
```

- **`hop.py`** was inherited, not this PR's: the base branch had already fixed it in `a7c9d6fd5` ("merge: master into #1895, and clear the comment guard it now trips"), which `fcd46ace5` predates. Merging the current base head cleared it — no duplicate fix, no divergence with #1895.
- **`flatten.py`** was ours: `flatten_terminal_with_over_optional`'s docstring cited `(#1896)` as its rationale. Deleted the citation; the rationale is carried by the pin names in `reentry/test_flatten_terminal_with.py`. The grandfathered `1` is the pre-existing `#1341` module docstring.

**No baseline cap was raised.** The only baseline edit is a **tightening**: `gfql_unified.py` `comment-block` 32 → 21 and `issue-rationale` 15 → 11, locked because the externing below made the improvement real. I deliberately did *not* run a blanket `--update-baseline`: 26 other files are now below their caps, but those improvements belong to master/#1895 and locking them here would just manufacture a conflict on a shared file.

## 3. Merge resolution

One conflict, in `lazy/engine/polars/chain.py`. diff3 showed it was the trap you described: our side had trimmed a 5-line comment to 3, and *theirs* looked like it had rewritten the same comment — but the real content of their side was a new `edge_src, edge_dst = _bound_edge_endpoints(self)` line that the code below now depends on. Kept **both**: master's extraction plus our trimmed comment. `_bound_edge_endpoints` is master's new helper (chain.py:855) and is used at the `return g0.nodes(...).edges(...)` below.

`bin/ci_cypher_surface_guard_baseline.json`: `lowering_py_max_lines` resolved to `9464`, which is the **measured** line count of the merged `lowering.py` (master was 9425 and untouched there; the delta is entirely this PR's). The ratchet you flagged still has **zero slack by construction** — I left it that way rather than quietly loosening it. Say the word and I'll add headroom.

## 4. Per-thread disposition (all 16)

| # | Thread | Disposition | Evidence |
|---|---|---|---|
| 1 | `execution.py` "Violates comment rules too" | **Addressed** in `fcd46ace5` | `grep -n "never synthesize ANONYMOUS" reentry/execution.py` → no match. Rule now in the helper name `_optional_reentry_unmatched_identity_null_rows` (execution.py:197). Guard: file at 4 vs cap 5. |
| 2 | `execution.py` "Avoid Any when narrow available" | **Addressed** in `fcd46ace5` | `reentry/execution.py` is **absent** from `ci_type_hygiene_baseline.json`'s `explicit-any` map — absent means it must be zero. Aliases `CypherFillValue`/`CypherFillRow` at execution.py:45/48. Self-review gate over the whole product diff returns no added `Any`/`hygiene-ok`/`cast(`. |
| 3 | `flatten.py` "More comment issues" | **Addressed** in `fcd46ace5`, **+1 residual fixed here** | 20-line docstring is now 2 lines + named helpers (`_is_terminal_with_over_optional_match`, `_stage_reshapes_rows`, `_query_without_pure_carry_stage`, `_query_with_terminal_stage_folded_into_return`). The residual `(#1896)` citation the new guard caught is deleted. |
| 4 | `execution.py:177` "tck-gfql obligations likely, audit & handle" | **Addressed** — and here is the explicit answer I owed you | Companion branch **exists and CI will run it**: `tck-gfql@fix/gfql-1896-om-with-pipeline` = `48ced5951`; `ci.yml:672-703` resolves the tck ref by `head_ref` match. **Yes, scenario outcomes change and yes the manifest needed regenerating** — `09008dd contract: 4 OPTIONAL MATCH scenarios now succeed with correct rows`, `674f3b6 contract: regen capability-debt manifest for the 4 promoted keys`, `48ced59 fully promote match-where6-2 + match7-29/30/31`. |
| 5 | test file location | **Was open — fixed here** | Section 1. |
| 6 | `flatten.py:354` "significant pos/neg test obligations" | **Addressed** in `fcd46ace5` | `reentry/test_flatten_terminal_with.py`, 24 tests, both admit and decline branches at AST level. |
| 7 | `ast.py:30` "is this from lower on our pr stack?" | **Addressed — now reconciled, not just answered** | `git diff ghhttps/master HEAD -- graphistry/compute/gfql/cypher/ast.py` is **empty** after the master merge (#1894 landed in `526976e91`). The byte-identical copy merged away exactly as intended. Same for `agg_types.py`. |
| 8 | `lowering.py:9152` "pos/neg test amplification obligations" | **Addressed** in `fcd46ace5`, verified anti-vacuous here | The `test_1896_*` family in `test_optional_match_semantics.py`: 22 test instances over pandas+polars, including 3 decline pins. **Measured anti-vacuity**: replayed against the base `a7c9d6fd5`, **20 of 22 fail**, 2 pass (positive controls). |
| 9 | `agg_types.py` verbose comment + root cause | **Addressed** in `fcd46ace5` | `git diff ghhttps/master HEAD -- graphistry/compute/gfql/agg_types.py` empty; the TYPE_CHECKING note is gone on both. Root cause is #1924, and it is now a **machine gate** (`bin/ci_comment_density_guard.py`, #1926) rather than a promise — this branch runs it clean. |
| 10 | `gfql_unified.py:1515` "strange location, shouldn't they be externed?" | **Was PARTIAL — fixed here** | The empty-group fill-value symbols did land in `gfql/agg_types.py`. The **carried-output resolution** did not: two regexes and three functions were still inline. Now `graphistry/compute/gfql/cypher/reentry/carried_outputs.py` (167 lines), which is also where `CarriedOutputSources` / `CARRIED_OUTPUTS_NOT_REPRODUCIBLE` now live — with their producer instead of their consumer. |
| 11 | `gfql_unified.py` "already overly large" | **Was open — fixed here** | **2541 → 2422 lines** (−119). Ratcheted and locked: `comment-block` 32 → 21, `issue-rationale` 15 → 11. |
| 12 | "replace dynamic patterns like getattr with structured & static programming" | **Was STILL OPEN at `fcd46ace5` — fixed here** | `getattr(op, "function", None)` / `getattr(op, "params", None)` were still at gfql_unified.py:1553-1554. Now `isinstance(op, ASTCall)` in `_projection_items_and_grouping`, matching what the sibling `optional_reentry_aggregate_fill_values` already did in the same file. |
| 13 | "test amplification concerns?" on `_carried_output_sources` | **Addressed** in `fcd46ace5` | `reentry/test_carried_outputs.py`, 7 named params covering every branch (incl. `p.__cypher_reentry_av__` read-through and the uncarried-marker fallthrough), plus the 22-test execution family. |
| 14 | `execution.py:56` "are any of these externally needed types?" | **Now answered, and acted on** | **Yes** for `CarriedOutputSources` + `CARRIED_OUTPUTS_NOT_REPRODUCIBLE` — `gfql_unified` imports both — so they moved to `carried_outputs.py`. **No** for `CypherFillValue`/`CypherFillRow`: `grep -rn` shows every use is inside `reentry/execution.py`, so they stay private there. |
| 15 | `flatten.py:45` "should these be externed and/or swapped to predicates?" | **Was open — fixed here, both halves** | Externed **and** turned into predicates: `is_bare_identifier()` / `identifier_tokens()` in `graphistry/compute/gfql/identifiers.py`. That retired four duplicates of the same regex — `flatten._BARE_IDENT`, `flatten._IDENT_TOKEN`, `gfql_unified._IDENTIFIER_TOKEN`, and two inline copies in `reentry/carry.py`. `_REENTRY_MARKER_COLUMN` became `reentry_hidden_column_output_name()` in `naming.py`, **built from the same prefix/suffix constants as `_reentry_hidden_column_name()`** so the writer and the parser cannot drift. `_AGGREGATE_CALL` stays local — one call site, flatten-specific. |
| 16 | `agg_types.py` "overly verbose" | **Addressed** in `fcd46ace5` | Same as #9. |

Also added the missing `CHANGELOG.md` entry for #1896 — it had none.

## Gates

| gate | result |
|---|---|
| `ruff check graphistry bin` | clean |
| `ci_comment_density_guard.py` | **OK**, no growth (was FAILING before the merge) |
| `ci_type_hygiene_guard.py` | OK, no growth |
| `ci_cypher_surface_guard.py` | pass; `lowering.py` 9464 = baseline 9464 |
| `mypy graphistry` | 4 errors, 329 files — **byte-identical to the base's 4 errors / 328 files**, zero new |
| `pytest graphistry/tests/compute` | **102 failed / 10243 passed** vs base **102 failed / 10190 passed** — failure lists **md5-identical** (`6a75a38ebb30614a35083426514a76ef`), `diff` empty. The 102 are pre-existing cuDF-lane failures on this box, unchanged. **+53 passed** = the 53 tests this PR adds (22 `#1896` execution + 24 terminal-WITH + 7 carried-output). |
| self-review gate on the product diff | only `#:` Sphinx alias doc-comments (the form accepted on #1894); no added `Any` / `type: ignore` / `hygiene-ok` / `cast(` |

Failure lists captured whole (`… 2>&1 \| grep '^FAILED' \| sort > file`, never through `tail`) and compared as sets against a baseline generated from the merged base `a7c9d6fd5` in a separate worktree.


Pushed as `e08e0dbad` (`fcd46ace5` -> merge `master` `526976e91` -> merge base `a7c9d6fd5` -> the three commits above). `mergeable: MERGEABLE`, base unchanged, CI running. Nothing was resolved, force-pushed, rebased, or re-based.

lmeyerov:
## Base merged forward + amplification round 1 — **NOT QUIET**

### Merge resolution

Merged `ghhttps/fix/gfql-1888-endpoint-closure` (#1895) with `-c merge.conflictstyle=diff3`. #1895 moved twice during this round, so the merge was done twice; the final one is against **`21167e08f`** (round 5 pins + round 6 polars NULL-endpoint closure fix).

- **Zero conflicts.** #1897 already carried #1895 through `325d53fdf`, and the two new commits touch `hop_eager.py` / `hop.py` / four hop-and-closure test modules, none of which #1897 touches.
- `CHANGELOG.md` was the one file both sides appended to. Both entries are kept — #1897's `OPTIONAL MATCH` + terminal `WITH` entry and #1895's round-6 polars closure entry.
- `bin/ci_cypher_surface_guard_baseline.json` → `lowering_py_max_lines`: the merged tree measures **9464** lines of `lowering.py` and the baseline already reads 9464, so no edit was needed. Verified by `wc -l` on the merged tree, not by trusting either side's number.

### Mutation audit (done FIRST, before any new test)

68 mutations over every functional site in this PR's diff — flatten.py 28, carried_outputs.py 16, execution.py 13, lowering.py 8, identifiers.py 2, naming.py 1 — run against the PR's own pins (`test_optional_match_semantics.py` + `cypher/reentry/`).

| | count |
|---|---:|
| behaviours mutation-tested | **68** |
| caught by the PR's own pins | **39** |
| **survived** | **29** |
| now caught by this round's pins | **19** |
| of those, already covered elsewhere (pin dropped) | **1** |
| still surviving, all verified defensive | **10** |

The 10 still surviving produce **byte-identical output on a 30-query battery** covering the whole OM+WITH surface (pure carry, WHERE on either side, aggregate/rename/DISTINCT/ORDER/SKIP/LIMIT stages, reentry with and without LIMIT, scalar and whole-row carries, duplicate carries). They are defensive branches with no Cypher spelling that reaches them, not untested behaviour:
`F13`/`F14` (`carried <= match_aliases`, `carried` non-empty), `F16` (`_bare_carry_aliases` identifier check), `F26` (`bare_carries` needs a match alias), `E02`/`E07`/`E08`/`E09` (identity-fill guards), `E10` (duplicate carried keys — unreachable *because of* the multiplicity defect below), `L05` (the `stage_row_filter is None` recompile arm, which no admitted shape takes).

I also ran the survivors against the whole `gfql/cypher/` + optional-match suite to catch redundancy before pinning. That pass reached 11 of 29 before I cut it for wall-clock, and it earned its keep: `F28` (dropping `collect` from the stage-aggregate pattern) is **already** caught by `test_lowering.py::test_string_cypher_failfast_optional_match_collect_null_whole_row_with_boundary`, so I removed the `collect` param I had written and left a comment naming the test that covers it. Dropping `sum` or `max` reddens nothing, measured the same way, so those two stayed. The other 10 checked (`F01`, `F03`, `F13`–`F16`, `F18`, `F22`, `F24`, `F26`) survived that whole suite.

A methodology note worth recording: my first audit pass reported **68/68 caught**. It was fabricated — `--timeout=600` is not installed here, so every run was a pytest *usage error* that my status check read as a failing test. The numbers above come from parsing pytest's own summary line and nothing else, and a run without a summary line is now classified `NO_SUMMARY_LINE`, never "caught".

### Decorative pins found (3)

1. `test_declines_when_with_precedes_the_optional_match_reentry_shape` — names the `reentry_matches` guard but declines on `any(m.optional)` instead: in `MATCH (a) WITH a OPTIONAL MATCH ...` the OPTIONAL arm moves *into* `reentry_matches`, so `query.matches` holds no optional at all. Deleting the `reentry_matches` guard left it green. Fixed by adding a discriminating sibling (`MATCH ... OPTIONAL MATCH ... WITH a, b MATCH (b)-->(c)`), where every other precondition passes and only that guard can stop the flatten.
2. `test_declines_duplicate_stage_output_names` — declines via the RETURN-passthrough branch, not the duplicate-name check. With the check removed the query *answers*, silently resolving `x` to whichever item is written last; the pin stayed green.
3. `test_1896_limit_carry_null_extension_not_shortcircuited` — its docstring claims it pins "the old `result_rows >= prefix_rows` short-circuit", but restoring exactly that expression leaves it green (the identity anti-join now serves that shape before the short-circuit is reached). The real short-circuit needed a shape where the result is as long as the prefix *and* nothing identifies the rows.

### Vacuous cells

**1 found, and 22 cleared by measurement.** `test_optional_self_loop_arm_projects_edge[polars]` swallowed `NotImplementedError` with a bare `return` — the one cell in `test_optional_match_semantics.py` that could pass without reaching an assertion. It now asserts the typed decline and names the unsupported feature (`cross-entity (same-path) WHERE`).

I specifically went looking for the #1904 pattern in the rest of the file and it is **not** there: the 20-cell `test_optional_match_seed_shapes_gate_or_keep_seeds` sweep and both `_parity_or_nie` call sites were instrumented, and all 22 cells reach their assertion on both engines. Reporting that as measured rather than assumed.

### New cells + anti-vacuity

**72 new cells** (103 collected across the three files, 31 of which are the PR's own).

`graphistry/tests/compute/gfql/test_optional_match_with_pipeline_boundaries.py` — 15 shapes × pandas/polars/cuDF = **45 cells**:

- **30 fail at the merge base** (10 of the 15 shapes): empty edge table, zero-matching typed arm, single binding row, a `WITH…WHERE` over a NULL-valued property (three-valued logic), a carried property named like the edge `s` endpoint column, `RETURN` renaming a folded stage output, and three declines whose merge-base message described a different limitation.
- **12 are base-passing, each named a deliberate control in its docstring**, and each kills a live mutation: empty node table (no fabricated fill row), duplicate stage output names, a WITH pipeline with no OPTIONAL MATCH at all (the hook's precondition), and a single carried row with zero matches.
- **3 are a strict xfail** — see the defect below.

`cypher/reentry/test_carried_outputs.py` and `test_flatten_terminal_with.py` — **+27 unit cells**. Both files are new in *this* PR, so a merge-base comparison is undefined for them; instead each is validated by killing a named mutation that survived both the PR's pins and the broader cypher suite: the `carried_output_sources` branches (`C07` scalar sources, `C09`/`C10` grouping, `C11` no-projection, `C12` the `CARRIED_OUTPUTS_NOT_REPRODUCIBLE` decline, `C13` `with_` as a projection stage, `C16` per-group_by fill reset), the null-fill's two *distinct* decline messages (`E05`), the shared identifier form in `identifiers.py` (`I01`/`I02` — it had no direct pin anywhere in the repo), and the flatten's `reentry_matches` and `CALL` guards (`F01`, `F03`) plus the `sum`/`max` names in its stage-aggregate pattern.

`CARRIED_OUTPUTS_NOT_REPRODUCIBLE` deserves a note: I could not find **any** Cypher query that reaches it. Every irreproducible projection I could write is rejected earlier (backticked property names do not parse; arithmetic over a carried alias hits the #1273 multi-source residual). It and its consumer are pinned at the unit level and labelled defensive in the test docstring rather than presented as covered behaviour.

### Defect found — strict xfail

`MATCH (a:P)-[]->(c) WITH a AS p OPTIONAL MATCH (p)-[:KNOWS]->(d) RETURN p.id, d.id`

The prefix binds `a1` twice and `a2` once, so `WITH a AS p` carries **three** rows and openCypher returns **5**. All three engines return **3**: a whole-row carry deduplicates the prefix frame by node id (`execution.py` returns early at `if not carried_columns:` before the `duplicate_carried_node_rows` decline can fire, which only guards the *scalar*-carry path). Cross-engine agreement is not evidence here — this is precisely the shape where all three agree on the wrong answer.

Pre-existing, and #1897 **improved** it: the merge base returned 2 rows (a silent inner join that dropped `a2` entirely), the PR now null-extends `a2` to 3. The remaining multiplicity half is pinned as a `strict=True` xfail naming the condition, so it flips red the day it is fixed. It is also why `E10` (the duplicate-carried-key guard) is unreachable.

### Engines

pandas, polars, and **cuDF 25.10 on a real GPU** — the cuDF cells were executed here, they did not skip. cuDF had **no** coverage of this feature before this round; all 15 new shapes now run on it, and all 15 agree with the hand-computed openCypher oracle. No engine legitimately declines any of the 15, so no cell is discharged by a skip.

Lane routing, stated precisely: the polars cells run in `test-polars.sh` (the module is registered there, and `test_polars_lane_completeness.py` passes). The cuDF cells skip on the default pandas lane for want of the wheel and execute in `ci-gpu.yml`, which runs the whole suite on `gpu_public` — that workflow is label-gated on `gpu-ci`, so these cells are only as green as that label. They are green on this box today.

### Gates

- `python -m ruff check graphistry bin` — clean
- All three guards, run from inside the worktree: cypher-surface ✅, comment-density ✅ (no growth), type-hygiene ✅ (no growth)
- `python -m mypy graphistry` — 4 errors, byte-identical to the same merge without this round's commits (0 new)
- `test_polars_lane_completeness.py` — 11 passed; the new module is registered in `bin/test-polars.sh`, so its polars cells actually run in the polars lane
- Coverage baselines — no new **source** file in this round; `carried_outputs.py` is already registered in the three baselines whose profile targets it (the `gfql-polars` profile targets `compute/gfql/lazy/**` only, so `ci-polars-py3.12.json` correctly has no entry)
- Full `graphistry/tests/compute`, run twice under `-p no:randomly` — this merge **without** the round (`102 failed, 10324 passed, 954 skipped, 62 xfailed`) and **with** it (`102 failed, 10394 passed, 954 skipped, 65 xfailed`). The two `FAILED` lists, captured whole and sorted (never through `tail`), are **byte-identical**: `diff` is empty. **Zero failures introduced.** The delta is exactly `+70 passed, +3 xfailed` = the 73 new cells that run in that suite (one has since been removed as redundant, leaving 72). Both runs ended with their summary line, so neither was a killed run.
  The 102 are pre-existing and local to this box: `CuPy failed to load libnvrtc.so.12`, concentrated in `gfql/index/` and the polars conformance matrix. I confirmed they are identical at the merge base too, so none of them belongs to #1897 or to the merge.

### Verdict

**NOT QUIET.** Round 1 found 29 unpinned mutations (19 now pinned, 10 shown defensive), 3 decorative pins, 1 vacuous polars cell, a whole uncovered engine, and one real silent-wrong shape now under strict xfail. Convergence needs two consecutive quiet rounds, so this cannot converge today; round 2 should re-audit against the new pins and push on the `E0x` identity-fill guards, which are still reachable only in principle.

lmeyerov:
## Amplification round 2 — closing report

Answering the shipping question directly first, then round 1's three open items.

## 1. Defects INTRODUCED by this PR (the blocking category)

**No wrong answer is introduced by this PR.** Every wrong answer I found is wrong at the merge base too, and in each case this PR either fixes it or converts it into a typed decline. Measured base-vs-head on the same graphs, hand-computed oracles, three engines:

| shape | base | head |
|---|---|---|
| `MATCH path = (a:P)-[:KNOWS]->(b) OPTIONAL MATCH (b)-[:KNOWS]->(c) WITH path,a,b,c RETURN a.id, c.id` | declines | **answers correctly** |
| `MATCH (a:P) OPTIONAL MATCH (a)-[:KNOWS*1..2]->(b) WITH a,b RETURN a.id, b.id` | **wrong, and differently per engine** (pandas 1 row with a null; polars/cuDF join a1 to itself) | typed decline naming the row-synthesis limit |
| `MATCH (a:P) WITH a AS p OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p.grp AS g, b.id AS bid` (grp = g,g,h) | **wrong** — 3 rows with two grp values replaced by null | typed decline |
| `... RETURN p.id AS pid, p.v AS pv, b.id AS bid` over NULL-valued `v` | **wrong** — both unmatched rows lose pid and pv | **answers correctly** |
| `MATCH (a:P) OPTIONAL MATCH (a)-[:KNOWS]->(b) WITH a,b RETURN a.v AS av, b.id AS bid` over NULL-valued `v` | **wrong** (pandas/cuDF collapse to 2 rows; polars answers b.id = a1/a2) | **answers correctly** |
| `MATCH (a:P) WITH a AS p OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p.id AS pid, b.id AS bid` with a NULL node id | **wrong** — a2 comes back as null | **answers correctly** |
| `MATCH (a:P) WITH a AS p LIMIT 2 OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p.v + 1 AS pv` | **wrong** — `[2, None]` where openCypher says `[2, 2, 3]` | typed decline |

**One narrow behaviour is worse at head, and it is a decline rather than a wrong answer.** Reporting it as the closest thing to a category-1 finding so the call is yours, not mine:

```cypher
MATCH (a:P) WITH a AS p OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p.nul AS g, b.id AS bid
-- nodes a1,a2,a3 (:P), b1 (:C); edge a1-KNOWS->b1; column `nul` is NULL on every row
```

openCypher answers `[(null,b1), (null,null), (null,null)]`. The **base answers exactly that** — by accident: it padded with all-null rows and the projected column is null anyway. **Head declines** (`no uniquely-identifying carried-alias columns are projected`), because every anti-join key is NULL and therefore ambiguous. Sound by construction, never wrong, but one shape that used to come back now does not. Pinned as a decline with the openCypher answer written into the docstring so the trade is visible. I do not think it blocks; it is the price of the guard that stops the `grp = g,g,h` row loss in the row above.

Full `graphistry/tests/compute` on the merged tree: **102 failed, 10428 passed, 957 skipped, 74 xfailed**. The 102-name `FAILED` set is **byte-identical** to the same merge without this round — `md5 6a75a38ebb30614a35083426514a76ef`, the same digest round 1 reported. Zero failures introduced. Both runs ended with their summary line.

## 2. Pre-existing, not fixed here

**Whole-entity carried alias is dropped from the null-extended row.**

```cypher
MATCH (a:P) WITH a AS p, a.id AS pid LIMIT 2 OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p, pid, b.id AS bid
```

`p` is bound for every carried row and OPTIONAL MATCH cannot unbind it, so a2's null-extended row must carry a2's own node columns — only `bid` is NULL. pandas and cuDF both return `p.* = NULL`, and they disagree on how: pandas renders the boolean label columns `None`, **cuDF renders them `False`**. Byte-identical at the merge base, so it predates #1897. polars declines the shape outright with the typed scalar-carry `NotImplementedError`.

Disposition: `strict=True` xfail over pandas and cuDF naming the condition, so it flips red the day it is fixed. Same family as round 1's duplicate-carried-prefix-row multiplicity xfail (which I re-verified: it **is** strict — I replaced its expectation with the 3-row answer the engines actually give and got three `XPASS`es reported as `FAILED` — and it **does** name the real condition: I instrumented the reentry state builder and the prefix frame arrives with 2 rows, not 3, so the whole-row carry really is deduplicating by node id upstream). Both want one follow-up issue.

**A vacuous cell round 1 reported as cleared.** Round 1 said both `_parity_or_nie` call sites were instrumented and all cells reach their assertion. I re-measured by replacing every swallow site with `pytest.fail("VACUITY-PROBE")`:

```
FAILED test_optional_match_semantics.py::test_with_carried_scalar_props_after_optional_reentry[polars]
E   Failed: VACUITY-PROBE: parity_or_nie swallowed NIE
```

Fixed here, not deferred: `_parity_or_nie` now requires the caller to **name** the declined feature; an undeclared decline fails, and so does polars *starting* to serve a shape declared as declining. The 20-cell `gate_or_keep_seeds` sweep came back clean under the same probe, so round 1's claim holds there. Re-run after the fix: no probe hit anywhere in the PR's test surface.

## 3. Round 1's three open items

### (a) The redundancy sweep — finished, all 29

Run against the **base branch's own test tree** (`git checkout 21167e08 -- graphistry/tests/`, this PR's three new test modules removed), so "covered elsewhere" means covered by a test that predates this PR.

**5 of 29 are already caught by pre-existing tests.** Round 1 had checked 16 and reported 1; I ran the remaining 13 and re-ran its 16.

| mutation | pre-existing test that catches it |
|---|---|
| `F28` drop `collect` from the stage-aggregate pattern | `test_lowering.py::test_string_cypher_failfast_optional_match_collect_null_whole_row_with_boundary` (round 1 found this, dropped its pin) |
| `E04` single-row guard stops checking `result_rows` | `test_lowering.py::test_issue_1461_optional_reentry_null_extension_does_not_leak_unprojected_scalar` |
| `E08` identity pairs ignore `src in prefix_columns` | `test_lowering.py::test_issue_1047_single_row_prefix_with_optional_match_still_works` |
| `E09` `not pairs` returns a row instead of `None` | same |
| `L04` flatten hook drops its OPTIONAL-MATCH precondition | the pre-existing cypher suite, 151 tests over |

`E04` and `L04` were in round 1's own log and it added pins for them anyway, calling them mutations "nothing else catches". I have **not** removed those pins — the covering tests are pandas-only, and round 1's cells add polars and cuDF — but their docstrings now name what else covers them, so the next reader is not told they are the only thing standing there. The other 24 survived the whole pre-existing suite.

### (b) The ten "defensive branch" survivors — **4 did not hold**

I did not re-use round 1's 30-query battery. I instrumented the five guard sites and ran the whole `graphistry/tests/compute` tree recording, per call, exactly what each guard saw and whether the mutant's condition would have differed — then hunted the shape that breaks each remaining equivalence.

**Four are live, reachable from real Cypher, and had nothing catching them. Now pinned:**

| | shape | real | mutant |
|---|---|---|---|
| **F13** `carried <= match_aliases` | `MATCH path = (a:P)-[:KNOWS]->(b) OPTIONAL MATCH (b)-[:KNOWS]->(c) WITH path,a,b,c WHERE a.v <= 1 RETURN a.id, c.id` | typed decline | **answers** |
| **F26** fold arm's `text in match_aliases` | `... WITH path, a.id AS aid RETURN path.x AS px` | WITH-pipeline decline | "Unknown Cypher alias 'path'" |
| **E10** identity anti-join duplicate-key guard | `... RETURN p.grp AS g, b.id AS bid`, grp = g,g,h | typed decline | **silently drops a2's row** (2 for 3); all-NULL keys drop **both** (1 for 3) |
| **L05** `stage_row_filter is None` recompile arm | `MATCH (a:P) OPTIONAL MATCH (a)-[:KNOWS*1..2]->(b) WITH a,b RETURN a.id, b.id` | decline naming the row-synthesis limit | decline naming the WITH pipeline |

F13 and F26 come from one line: `_match_clause_aliases` walks pattern *elements*, and a **path alias** lives on `MatchClause.pattern_aliases`. It is the only bare carried identifier that no MATCH clause binds. `flatten_carried_endpoint_rebind` reads `pattern_aliases` explicitly for exactly this reason; `flatten_terminal_with_over_optional` does not.

**Two more are reachable but already covered elsewhere** (`E08`, `E09`, above). **Two are equivalent mutants, proven rather than asserted:**

- **E02** — with an empty mapping, `_optional_reentry_unmatched_identity_null_rows` computes `pairs = []` and returns `None` on its first branch; the function is pure. The `and carried_outputs.columns` is a short-circuit with no observable effect. No test can kill it.
- **F16** — the mutant's extra set members are non-identifier strings, and every consumer tests them against `match_aliases` / `prefix_aliases`, which the parser fills from `NodePattern.variable` / `RelationshipPattern.variable` and so contains only bare identifiers. `carried <= match_aliases`, `carried == prefix_aliases` and `carried != prefix_aliases` all take the same branch either way. Measured too: the check differs on exactly one call in the whole tree and the difference does not propagate.

**One I could neither kill nor prove: `E07`.** Reachable only if the executed result frame is None-or-columnless while a carried-output mapping is non-empty; across 14 calls in the whole tree the two conditions were always co-satisfied, because the mapping's keys are exactly the executed projection's output columns. Reported as unproven, not as defensive.

**`F14`** has no Cypher spelling — the parser rejects a WITH with zero items — but is killable from a hand-built AST, and now is, in the style of the existing `row_sequence` / `CALL` pins.

### (c) `CARRIED_OUTPUTS_NOT_REPRODUCIBLE` and `E10` — **neither is dead code. Do not delete either.**

`CARRIED_OUTPUTS_NOT_REPRODUCIBLE` has at least six Cypher spellings. Any expression over a carried alias in a RETURN that names only that alias:

```cypher
MATCH (a:P) WITH a AS p LIMIT 2 OPTIONAL MATCH (p)-[:KNOWS]->(b) RETURN p.v + 1 AS pv
```

and likewise `p.v + p.w`, `p.id AS pid, p.v * 2 AS pv2`, `toUpper(p.name)`, `coalesce(p.v, 0)`, and the scalar-carry form `WITH a.v AS av, a AS p ... RETURN av + 1`. Round 1's near-miss is visible in its own note: every attempt it made also referenced `b` in the RETURN, so the #1273 multi-source residual intercepted it first. The base answers that query `[2, None]` where openCypher says `[2, 2, 3]`, so the decline is a fix, not just a gate. Now pinned end-to-end on all three engines, and the stale "no Cypher spelling that reaches them" comment in `test_carried_outputs.py` is corrected in place.

`E10` is live and load-bearing — see the table above. Round 1 called it unreachable "because of the multiplicity defect"; that conflated it with the `duplicate_carried_node_rows` guard in `compiled_query_reentry_state`, which is a different guard on a different path. `E10` guards the **identity anti-join** and is reached by any non-unique carried identity projection. Removing it does not decline, it loses rows.

## 4. Mutation numbers

Same 68-behaviour catalog round 1 built over this PR's diff, re-run in full at this head. Status comes only from pytest's own summary line; a run without one is `NO_SUMMARY_LINE`, never a catch.

| | count |
|---|---:|
| behaviours mutation-tested | **68** |
| caught by pins in this PR's test set | **62** |
| caught only by pre-existing tests elsewhere (`F28`, `E08`, `E09`) | **3** |
| provably equivalent mutants (`F16`, `E02`) | **2** |
| no reachable shape found, equivalence unproven (`E07`) | **1** |
| newly caught **by round 2** (`F13`, `F14`, `F26`, `E10`, `L05`) | **5** |

`E05` and `C12` were unit-only before this round and now also have end-to-end coverage on three engines.

## 5. New cells and anti-vacuity

**33 new cells**, no new test file.

`test_optional_match_with_pipeline_boundaries.py` — 9 shapes × pandas/polars/cuDF plus a 2-cell strict xfail = **29 cells**. **27 fail at base**; the 2 that do not are the strict xfail, which xfails on both trees. **No cell here is base-passing.** Measured by copying the file onto a base worktree and running it. That same run re-derives round 1's own claim at the current base — 30 of its 45 cells fail there and 12 pass, matching its "30 fail / 12 named controls".

`cypher/reentry/test_flatten_terminal_with.py` — **4 unit cells**. A merge-base comparison is undefined: the module does not import at base, because `flatten_terminal_with_over_optional` does not exist there. Each is validated instead by killing a named mutation that survived both this PR's pins and the entire pre-existing suite. The F13 pair is deliberately a pair — the admitting sibling (same carry, no stage WHERE) exists so the decline is attributable to the WHERE and nothing else.

The full-suite delta confirms the count: `+31 passed, +2 xfailed` = 33.

## 6. NULL axis

Probed on all three engines with hand-computed oracles, never cross-engine agreement: NULL node ids, NULL edge endpoints, NULL carried scalar values, a NULL key shared by a matched *and* an unmatched carried row, NULL booleans, three-valued `WHERE` over a NULL property.

**No divergence and no wrong answer on the flatten and identity-fill paths at head** — a measured negative, and the reason is structural and worth recording: the anti-join keys are compared in **Python** (`_optional_reentry_key_value` normalises NaN to None, membership is a `set`), not through an engine predicate, so the `isin`-vs-`is_in` NULL semantics that produced the silent wrong answer on #1895 never enter this code. Four of the new cells pin that. The one divergence that does exist is on the whole-entity carry path (§2) and is pre-existing.

## 7. Gates

| gate | result |
|---|---|
| `python -m ruff check graphistry bin` | clean |
| `ci_comment_density_guard.py` (run from inside the worktree) | OK, no growth, **no baseline cap raised** |
| `ci_type_hygiene_guard.py` | OK, no growth |
| `ci_cypher_surface_guard.py` | pass |
| `python -m mypy graphistry` | 4 errors / 329 files; this round touches **no source file**, so 0 new by construction |
| `test_polars_lane_completeness.py` | 11 passed; no new test module — the module carrying the polars cells is already registered in `bin/test-polars.sh` |
| coverage baselines | no new source file, so no baseline entry owed |
| full `graphistry/tests/compute` | 102 failed / 10428 passed / 957 skipped / 74 xfailed; `FAILED` set byte-identical to the same merge without this round (`md5 6a75a38e…`), `diff` empty |

Captured with `... 2>&1 \| grep '^FAILED' \| sort > file`, never through `tail`. Nothing weakened, skipped, or xfailed to get green; the one xfail added is `strict=True` and names a real known-wrong condition.

Engines: pandas, polars, and cuDF 25.10 on a real GPU — the cuDF cells executed, they did not skip. **Scoped off and stated:** cupy is broken on this box (`libnvrtc.so.12`), so cupy-backed index kernels and the `polars-gpu` lane could not run here; none of this round's cells touch them.

## 8. Merge and push

Re-fetched immediately before pushing: #1895 had moved to `0a1e8a4f1` (round 7). Merged with `-c merge.conflictstyle=diff3`, **zero conflicts** — it touches `CHANGELOG.md` (both entries kept) and `test_endpoint_closure_matrix.py`, neither of which this round edits. Guards and the full suite above were re-run **on the merged tree**.

Pushed as **`3c825d180`**. `mergeable: MERGEABLE`, base unchanged, CI running. Nothing merged, force-pushed, re-based, or resolved.

## Verdict

**No product defect is introduced by this PR.** Round 2 found four live guards that round 1 had misfiled as defensive (one of which, `E10`, is the difference between a decline and silently losing rows), a decline branch reported as unreachable dead code that six Cypher spellings reach, one more vacuous cell, three redundancy findings round 1's cut sweep never reached, and a pre-existing cross-engine divergence now under a strict xfail. All of that is amplification, not a regression. The one behaviour that is narrower at head is a sound decline on an all-NULL identity projection, called out above.

Open for follow-up, none blocking: `E07` (unproven, no shape found), the pandas-only coverage of `E08`/`E09`, and one issue covering both whole-entity-carry xfails.

Issue #1899, opened 2026-08-14
GFQL Cypher: row-pipeline multiplicity + aggregate identities (round-004 umbrella)


lmeyerov:
Round-004 breadth probe (UNWIND/collect/WITH pipelines, ~70 hand-oracled queries × both engines at stack tip d044e3e84). Fixture: 5 nodes (Ann/Bob/Cat/Dan/Eve, null age/city cases), edges 1→2,1→3,2→3,3→4.

**1. SILENT-WRONG (both engines, top): single-endpoint projection multiplicity collapse.** `MATCH (a)-->(b) RETURN a.id AS id` → [1,2,3]; oracle bag [1,1,2,3]. Also b-side, through bare WITH carries, `WITH a.id AS id`, and post-WHERE. Damning contrast: `RETURN a.id, b.id` is correct (4 rows) and grouped aggregates are correctly multiplicity-weighted; and `WITH a RETURN count(*)` is typed-declined citing exactly this collapse — the non-aggregate projection of the same collapsed rows is served silently.

**2. UNTYPED CRASH + engine divergence:** `MATCH (a)-->(b) RETURN a.id AS id, count(*) AS c` → raw pandas KeyError 'id'; polars correct. Root cause: `_execute_single_hop_grouped_aggregate_fast_path` (gfql_fast_paths.py ~2291) — when the projected prop IS the node-id column, `lookup[out_col] = lookup[prop]` is a no-op and the later `rename(columns={node_col: lookup_key})` clobbers the output; decline-guard (~2270) only checks src/dst collisions. `AS id` is the natural spelling.

**3. SILENT-WRONG (both engines):** ungrouped aggregate over zero-row UNWIND stream eliminates the row: `UNWIND [] AS x RETURN count(*) AS c` → 0 rows; oracle one row c=0 (MATCH-sourced empty aggregation is correct — specific to the pure row pipeline).

**4. SILENT-WRONG (pandas-only-server):** negative list index returns null: `UNWIND [[1,2,3]] AS l RETURN l[-1]` → null, oracle 3 (polars NIEs, so pandas is the only server — wrong).

Known/pinned, not re-filed: sum-over-empty→null (BUG-4 strict xfail). Typed-gap inventory + polars NIE inventory + whole-entity collect() repr-string flag (#1650 review) in the probe report (scratchpad probe-unwind/). Swept clean: extensive UNWIND/collect/WITH inventory in report.

[...]

PR #1901, opened 2026-08-14
fix(gfql): row-pipeline multiplicity + aggregate identities (#1899)

lmeyerov:
Stacked on #1897. Fixes the four #1899 silent-wrongs/crash:

1. **Single-endpoint projection multiplicity** — single-alias projections compiled to the deduplicated node set instead of binding rows; a shared forcing predicate now routes bare `alias.prop` projections over non-optional relationship matches through binding rows (all three lanes), with `WITH a RETURN a.id` served via a new non-optional pure-carry flatten. Forcing exposed and fixed two real polars binding-row gaps (join-key dtype alignment; edges-only node materialization). Deliberately narrow scope; whole-row endpoint multiplicity remains pinned strict-xfail.
2. **`AS id` KeyError** — fast-path rename order fixed (also fixes the reverse `a.city AS id` join-key corruption hazard).
3. **Empty-UNWIND ungrouped aggregates** now emit the identity row (count→0, sum→0, collect→[]); BUG-4 (MATCH-sourced) intentionally untouched and still pinned.
4. **Pandas negative list index** normalized per-row; polars stays honest NIE.
Fold-in: reentry null-fill all-NA concat FutureWarning silenced dtype-losslessly (OM suite green under -W error::FutureWarning).

52 new pins (50P/2 strict-xfail residual). **Zero existing-test changes** — first broad forcing broke 10 tests and was fixed by narrowing the predicate + fixing 2 real polars gaps, not re-baselining. Gates: OM 81P, cypher tree 3181P, polars 742P, lowering 1420P (0 new), TCK 4145P/0F, ruff+guards green (surface baseline 9503→9650).


lmeyerov:
**MERGE BLOCKER — GPU sweep @2fb3f8813: 23 failed / 8867 passed** (prior tip 0F). Engine-uniform (pandas/polars/cudf identical) — the local gate list missed the top-level `graphistry/tests/compute/gfql/` tree. Groups: (A) `seeded_typed_hop` fast path no longer engages (15 tests — this is the benchmark-critical lever from the abstraction-tax stack; perf regression risk, not just test churn); (B) served-path dtype int64→float64; (C) `test_a_shape_neither_path_serves_declines_both` now sees 5 rows vs 3 — possibly the CORRECT multiplicity colliding with an old deduped expectation, needs oracle adjudication not blind re-baseline; (D) two polars shapes now route to the NIE 'rows' op — users regress from served to NotImplementedError; (E) new test file needs POLARS_TEST_FILES registration. Fix in flight; do not merge.

lmeyerov:
Blocker fix pushed (`bf808025e`): the forcing predicate now yields to the seeded typed-hop fast path (exact-shape exclusion mirroring the fast path's own match test) — engagement flags restored, dtype parity back, polars served→NIE regression undone, lane registration added. The 5-vs-3 test was adjudicated by hand oracle: the bag [0,0,0,1,2] is correct openCypher; the old `== 3` encoded the multiplicity-collapse bug (justification comment in the test). Full-tree standing gate: failure set byte-identical to branch-base baseline (93 local-env failures both sides, diff clean). Residual honestly kept: the fast-path shape retains its pre-#1899 multiplicity envelope (documented + strict-xfail pins in force). GPU re-sweep queued.

lmeyerov:
**GPU sweep receipt (dgx, @bf808025e): 8890 passed / 53 skipped / 24 xfailed / 0 failed** (4m44s). The 23 regressions are gone; merge blocker lifted pending CI on the head (which adds only a truth-table-identical mypy narrowing line). Full log retained (sweep1899b.log).

lmeyerov:

Audited head `7c86922dc` against merge base `ef98faaab` (= current `origin/master`). Pushed two commits on top: a narrow fix for two introduced silent-wrongs in the PR's new identity-synthesis lane, plus 50 amplification cells appended to `test_row_multiplicity_semantics.py`.

### Verdict: BLOCKING FINDINGS, found and fixed in this push — SAFE TO LAND at `a425c9383`

**Category 1 (introduced at head, correct-shape at base) — both fixed here:**

1. **Compound aggregate items over an empty stream fabricated a value under a leaked internal column.**
   Repro at `7c86922dc`: `UNWIND [] AS x RETURN count(*) + 1 AS c` returned 1 row `{'__cypher_postagg__': 0}` (openCypher: `{'c': 1}`; master: 0 rows). Mixed shape dropped the compound alias entirely: `RETURN count(*) AS c, count(*) + 1 AS d` gave `{'c': 0, '__cypher_postagg__': 0}` with no `d`. Cause: `_row_only_empty_aggregate_row`'s `len(specs) != len(items)` guard doesn't exclude compound items — one spec per compound item, named by the internal post-agg placeholder. Fix: synthesis declines on `__cypher_postagg__`-named specs (restores master's 0-row shape); residual pinned strict-xfail. The MATCH-rooted lane already handled this correctly (`{'c': 1}`) on both revs, so the correct reference behavior exists in-repo; #1909/#1910's `aggregate_identity` module serves the row-only case properly.

2. **Chained ungrouped aggregates fabricated the final identity.**
   Repro at `7c86922dc`: `UNWIND [] AS x WITH count(*) AS c RETURN count(c) AS n` returned `{'n': 0}` — but `WITH count(*)` yields ONE row `{c: 0}`, so `count(c)` is **1** (`min(c)` → 0, `collect(c)` → `[0]`); master: 0 rows. Fix: synthesis declines when an earlier stage aggregates; residual pinned strict-xfail (#1910's replay serves it).

**Category 2 (pre-existing, same wrong at base) — pinned strict-xfail naming the condition, or listed:**
- Seeded typed-hop fast path dedupes parallel edges: `MATCH (a {name:'Ann'})-->(b) RETURN b.id` with 2 parallel edges gives `[2]`, openCypher `[2, 2]` — the #1899 carve-out routes there (xfail; catches carve-out drift via XPASS).
- Leading `OPTIONAL MATCH (a)-->(b) RETURN a.id` still collapses the bag (`[1,2,3]` vs `[1,1,2,3]`); guarded off binding rows to protect no-match null-extension, which I pinned green as the guard's justification (xfail).
- `SKIP 0` / `LIMIT >= 1` suppress the synthesized identity row (should page it): xfail naming #1909 — **fixed on #1910 in flight; those xfails plus the two residuals above will XPASS strict when #1910 merges master and must be flipped there (deliberate loud signal, called out on #1910 too).**
- `RETURN b` whole-row multiplicity collapse: PR's own strict xfail, verified honest.
- Listed only (no pin, #1910 territory): MATCH-path `sum()` over empty match returns null (openCypher 0); pandas `min` over all-null renders NaN vs polars None.

### Mutation audit: 17 mutants, 14 caught, 3 equivalent with stated proofs
Every catch verified by a pytest summary line (no NO_SUMMARY_LINE runs); baseline noise in every run = 1 pre-existing `cudf/libnvrtc` env failure.
- Caught (failures beyond baseline): forcing-off 25F; seeded-carve-out-off 6F (incl. XPASS of the parallel-edge xfail); pure-carry-flatten-off 3F; count-identity-None 8F; sum-identity-None 5F; collect-identity-None 3F; paging-guard-off 11F; mixed-specs-guard-off 4F; fast-path-rename-revert 4F; neg-subscript-off 4F; polars-materialize-off 2F; polars-endpoint-cast-off 2F; optional-guard-off 3F (via xfail XPASS); chained-guard-off 6F.
- Equivalent (proof stated): (a) neg-subscript out-of-range wrap→keep-negative-key: post-wrap negative keys can never match the non-negative cumcount position column, so the positional merge yields null either way; (b)+(c) flatten's `carried <= match_aliases` and downstream-scope guards: the binder's scope validation rejects those queries before flatten runs — demonstrated by removing the guards and observing the identical `GFQLValidationError [invalid-name-type]` on `WITH q RETURN q.id` / `WITH a RETURN b.id`.

### New cells + anti-vacuity
50 new cells (12 strict-xfail): 9 fail at the merge base (edges-only materialize x2, polars integral-float endpoints, nonintegral typed decline, unseeded parallel edges x2, bag paging x2, cuDF parity); 14 fail at the pre-fix head (the fabrication pins); the remainder are named mutant-killers/controls (paging-removal kills the paging-guard mutant, grouped-empty kills the mixed-specs mutant, null-extension control justifies the OPTIONAL guard, conservative-decline controls pin that forcing opened no unsupported lanes). PR's own 52 cells: 29 fail at base; 21 base-passing are legitimate controls; no swallow sites found by instrumentation.

### Engines
pandas + polars on every group; cuDF added for bag/identity/grouped/paging shapes (dataframe ops only — cupy-backed kernels and `polars-gpu` are out of scope here: this box's cuDF 25.10 works for dataframe ops but cupy is broken, `libnvrtc.so.12`).

### Gates
ruff clean; all three guards rc=0 run from inside the worktree (`lowering.py` 9635 = exactly at cap, **no cap raised** — the fix was compressed to fit); mypy: error set identical to base (4 pre-existing transitive); polars lane: cells live in the PR-registered `test_row_multiplicity_semantics.py`; no new source files. Full `graphistry/tests/compute`: failure SET at head **identical** to master base (102 = 102, both diffs empty; the known cuDF/libnvrtc + optional-dep failures).

### Cross-PR conflict note
#1910 deletes `_row_only_empty_aggregate_row` (replaced by `aggregate_identity.py`). When both land, resolve the `lowering.py` conflict by taking #1910's module wiring — it computes the compound and chained cases this PR declines.

Issue #1903, opened 2026-08-14
GFQL Cypher: var-length path trail semantics + lane-dependent cardinality (round-005 umbrella)


lmeyerov:
Round-005 path probe (76 queries × both engines @ d8668fdf7, hand-computed trail-semantics oracles, twice-verified; scripts in session scratchpad probe-paths/). Fixtures: diamond, directed 3-cycle, self-loop, parallel edges, typed, disconnected.

**SILENT-WRONG:**
1. **Lane-dependent cardinality — three answers for one MATCH** (diamond `[*2]`): `RETURN y.id`→1 row, `count(*)`→2, `RETURN x.id, y.id`→2. Single-alias endpoint projections compile to a reachability lane that collapses path multiplicity.
2. **Single-alias lane BFS visited-pruning drops valid k-trails** to nearer nodes (3-cycle undirected `[*2]` → [] vs oracle [b,c]); the two-alias row lane gets it right — pandas disagrees with itself.
3. **Row/agg lane uses WALK semantics on self-loops** (edge reused within a path: `[*2]` on self-loop returns (s,s), count 2 vs oracle 1; yet `[*3]` → 0 — neither walk nor trail).
4. **Relationship uniqueness not enforced across pattern elements**: `(x)-[]->(m)-[]-(y)` on a single-edge graph reuses the edge → 1 row vs oracle 0; parallel-edge 2-hop 6 vs 4.
5. Undirected self-loop: projection lane drops it, agg lane double-counts orientations.
6. shortestPath unreachable in PLAIN MATCH → null row instead of no row (pandas; indistinguishable from OPTIONAL).
7. polars `[*0..1]` silently drops the zero-hop seed (pandas correct; polars' own NIE text lists include_zero_hop_seed as must-error, but these forms don't raise).

**TYPED-GAP:** 8. `[*..3]` (valid openCypher omitted-lower-bound) → GFQLSyntaxError; parser grammar (~904-953) lacks the `*..M` production; also blocks bounded shortestPath.

Honest NIEs and a solid correct inventory (cycle termination exact, typed alternation, shortestPath ties/typed/min-bound) in the probe report. Net: path counting is only trustworthy via multi-alias row projections or grouped aggregates on acyclic neighborhoods. Successor context: #1748 (polars var-length), #1658 (CSR engagement).

lmeyerov:
**Umbrella verified served on master `e6625ed28`** — recommending close after review, with one bookkeeping dependency noted below.

[...]

Issue #1748, opened 2026-07-20
polars chain: forward/reverse varlen with min_hops>1 returns nodes outside the hop window (silent wrong)


lmeyerov:
## Summary

On the polars chain engine, a **forward or reverse** variable-length pattern with `min_hops > 1` returns nodes whose hop distance is **outside** the requested window. No error is raised — the answer is silently wrong.

Undirected `*2..3` correctly declines with a NotImplementedError. Only fwd/rev silently mis-answer.

## Repro

```python
import pandas as pd, polars as pl, graphistry
e = pd.DataFrame({"s": ["p0","p1","p2","p1"], "d": ["p1","p2","p4","p0"]})
q = "MATCH (a {id:'p0'})-[*2..3]->(b) RETURN b"

graphistry.edges(e, "s", "d").gfql(q, engine="pandas")
# -> p2, p4                      (correct: distances 2 and 3)

graphistry.edges(pl.from_pandas(e), "s", "d").gfql(q, engine="polars")
# -> p0, p1, p2, p4              (WRONG: p0 is distance 0, p1 is distance 1)
```

## Cause

The polars chain gates a node alias by hop distance only when the auto hop-label column is present. For `min_hops > 1` the label is deliberately not requested (those labels come from pandas' layered backward walk, which is not ported), so the alias ends up **ungated** and every reachable node passes.

Undirected happens to escape because it declines the whole shape earlier.

## Scope — measured, not inferred

4-engine differential sweep against pandas as oracle, 144 cases (4 hand graphs + 12 random graphs x 9 patterns), run on dgx-spark under `safe_run.sh`, RAPIDS 26.02 image.

| build | pandas | cudf | polars | polars-gpu |
|---|---|---|---|---|
| master `c355660a` | 144/144 | 144/144 | 79 ok / **49 wrong** / 16 NIE | 79 / **49** / 16 |
| master + #1746 + #1747 | 144/144 | 144/144 | 114 ok / **14 wrong** / 16 NIE | 114 / **14** / 16 |

Set-diff of the mismatch sets: **0 regressions, 70 of 98 fixed.** All 28 residual mismatches (14 per polars engine) are this one shape, `-[*2..3]->`.

So this bug is **pre-existing on master** and is not introduced by the #1741 stack — that stack removes 71% of the silent-wrong surface and leaves exactly this hole.

## Fix direction

[...]

PR #1904, opened 2026-08-14
fix(gfql): var-length TRAIL semantics + lane-consistent cardinality (#1903)

lmeyerov:
Stacked on #1902. The deepest umbrella of the campaign, SERVED not declined:

- **One MATCH = one cardinality**: the multiplicity-forcing router's var-length/seeded exclusions narrowed to exactly the shapes the reachability lane is correct for; benchmark fast paths keep precedence (engagement green).
- **TRAIL semantics implemented on both engines**: relationship identity tracked per path (edge may not repeat; nodes may) — fixes walk-semantics self-loop fabrication, cross-element edge reuse, and undirected self-loop orientation double-count (also in the indexed-bindings lane).
- shortestPath in plain MATCH: unreachable ⇒ no row (OPTIONAL keeps null). polars `[*0..N]` zero-hop seed served. `*..M` grammar production added. pandas computed-NaN comparisons now give openCypher answers (input-null masks distinguish genuine null from NaN).
- A-2 envelope: overclaiming comment corrected + strict-xfail pins documenting the true seeded-shape envelope.

64 new trail pins + 3 strict xfails. Existing-test changes oracle-adjudicated via an independent brute-force trail enumerator (never engine-vs-engine) — walk→trail literals recomputed, 46 shortestPath tests conformed to OPTIONAL MATCH spelling. Gates: full-tree failure set byte-identical to base, TCK 4144P/0F with 3 more scenarios promoted (companion cascade), mypy 328 files clean, ruff+guards green (surface 9759).


lmeyerov:
**GPU sweep receipt (dgx, @870d2defb): 8998 passed / 53 skipped / 27 xfailed / 0 failed** (4m54s) — delta vs prior tip exactly the +64 new trail pins / +3 xfails; trail-identity changes clean on the shared pandas/cudf row pipelines and polars var-length paths. Log retained (sweep1903.log).

lmeyerov:
**Round-006 re-probe (brute-force trail-enumerator oracle, 140 Cypher texts × 2 engines × 2 repeats): this PR's trail work is oracle-exact everywhere it was probed** — deep cycles with chords, unbounded `[*]->` termination, undirected trails counted exactly once, self-loop + parallel-edge combos, zero-hop `[*0..N]` respecting endpoint/edge-type filters, edge-uniqueness across pattern elements, and all 23 shortestPath shapes incl. the new `[*..M]` grammar end-to-end. polars: 75 typed declines, **zero silent divergences**.

Two residuals surfaced, neither a regression in this PR's values (A/B'd against master 25215ce5c):
- **#1906** — undirected + unbounded row projections stay on the source-table lane (the deliberate `lowering.py:2583-2591` early-return), so rows disagree with their now-correct `count(*)` twin and can contain impossible rows. Row values are identical on master; what changed is that the adjacent count lane used to raise a typed error there and now answers correctly. Being fixed-or-declined + pinned in the follow-up #1905 cycle.
- **#1787 hop-window starvation** (pre-existing, byte-identical master vs here): two silent sub-cases sit outside any pin — seeded undirected `*4..4` and unseeded undirected `*3..3`. Pins being added in the same follow-up.

graphistry/compute/gfql/lazy/engine/polars/row_pipeline.py
```diff
@@ -1813,17 +1805,24 @@ def _names(lf: pl.LazyFrame) -> List[str]:
                 payload_renames = {
                     col: f"{edge_alias}.{col}"
                     for col in _names(edges_f)
-                    if col not in (src, dst)
+                    if col not in (src, dst, "__gfql_edge_ident__")
```
lmeyerov: shouldn't we be using externed symbols now for these special vals?

graphistry/compute/gfql/lazy/engine/polars/row_pipeline.py
```diff
@@ -1869,75 +1868,73 @@ def _names(lf: pl.LazyFrame) -> List[str]:
...
                 state = (
                     state.join(oriented, left_on="__current__", right_on="__from__", how="inner")
                     .drop("__current__")
                     .rename({"__to__": "__current__"})
                 )
+                for _used in trail_cols_pl:
+                    state = state.filter(
+                        (pl.col("__gfql_edge_ident__") != pl.col(_used)) | pl.col(_used).is_null()
+                    )
+                _new_trail = f"__gfql_trail_{len(trail_cols_pl)}__"
```
lmeyerov: need externing too?

graphistry/compute/gfql/lazy/engine/polars/varlen_rows.py
```diff
@@ -62,58 +86,42 @@ def _directed_fixed_point_binding_rows_polars(
...
+    reachable cycle still raises the terminating-segments error here — the pandas
+    twin now serves those via trail tracking; this decline is the #1903 residual.
     """
     import polars as pl
     from graphistry.compute.gfql.lazy import collect as _lazy_collect
     from graphistry.compute.gfql.row.pipeline import RowPipelineMixin
 
     pairs_df = _lazy_collect(pairs)
     pairs_lf = pairs_df.lazy()
+    pairs_step = pairs_lf.select(["__from__", "__to__"])
```
lmeyerov: extern?

graphistry/compute/gfql/lazy/engine/polars/varlen_rows.py
```diff
@@ -124,7 +132,7 @@ def _directed_fixed_point_binding_rows_polars(
     while not exhausted:
         hop += 1
         frontier = _lazy_collect(
-            frontier_lf.join(pairs_lf, left_on="__current__", right_on="__from__", how="inner")
+            frontier_lf.join(pairs_step, left_on="__current__", right_on="__from__", how="inner")
```
lmeyerov: extern?

graphistry/compute/gfql/row/pipeline.py
```diff
@@ -3845,41 +3931,69 @@ def _gfql_connected_bindings_state(
...
+            shortest_path_mode = bool(
+                hop_column is not None and str(hop_column).startswith("__cypher_shortest_path_hops__")
+            )
             if edges_df_step is None or len(edges_df_step) == 0:
                 oriented = self._gfql_empty_frame(
                     edges_df_step if edges_df_step is not None else state_df,
                     columns=["__from__", "__to__"],
                 )
+                if "__gfql_edge_ident__" not in oriented.columns:
+                    oriented["__gfql_edge_ident__"] = self._gfql_broadcast_scalar(oriented, None)
```
lmeyerov: extern

graphistry/compute/gfql/row/pipeline.py
lmeyerov, line 1: this file needs:

1. symbol name externing to whatever external file
2. significant positive/negative correctness testing at either sides of its boundary
3. likely pyg-bench signoffs for avoiding perf regressions

lmeyerov:
All six threads addressed at `2ee64b5b5`. Also merged `master` in (the comment-density guard landed there after this PR branched) and the moved base branch, and made every guard pass without raising a cap.

Commits: `7297642f7` externing · `93d26adc3` boundary pins + guard debt · `d1f9340dc` trail-mask helper · `51bd5415b` comment-normalization repairs · `57eaad0b8` / `031727067` merges · `262e91b0c` py3.8 `Literal` lint fix · `5d4723ca3` polars-lane registration · `e15e47284` / `2ee64b5b5` base merges (the base branch moved three times during the review).

## 1-5. Externing (`row_pipeline.py:1808`, `:1935`, `varlen_rows.py:115`, `:135`, `row/pipeline.py:3946`)

Commit **`7297642f7`**. Done as one vocabulary, not five patches. Home is **`graphistry/compute/gfql/identifiers.py`** — its stated concept is "GFQL reserved identifiers", it already hosts a per-concept name (`HIDDEN_ALIAS_COLUMN_PREFIX`), and it is an import leaf, so `row/`, `same_path/`, `lazy/engine/polars/`, `cypher/` and `index/` can all reach it without a cycle. No new module.

| literal | constant | home | call sites |
|---|---|---|---|
| `"__from__"` | `WALK_FROM_COL: Final[str]` | `identifiers.py` | 55 in 9 files |
| `"__to__"` | `WALK_TO_COL: Final[str]` | `identifiers.py` | 54 in 9 files |
| `"__current__"` | `WALK_CURRENT_COL: Final[str]` | `identifiers.py` | 69 in 7 files |
| `"__gfql_prev__"` | `WALK_PREV_COL: Final[str]` | `identifiers.py` | 2 in 1 file |
| `"__gfql_edge_ident__"` | `TRAIL_EDGE_IDENT_COL: Final[str]` | `identifiers.py` | 31 in 3 files |
| `f"__gfql_trail_{i}__"` | `TRAIL_COLUMN_PREFIX` + `trail_column_name(i)` / `is_trail_column()` | `identifiers.py` | 12 in 3 files |
| `"__cypher_shortest_path_hops__"` | `SHORTEST_PATH_HOPS_COLUMN_PREFIX` + `shortest_path_hops_column()` / `is_shortest_path_hops_column()` | `identifiers.py` | 15 in 4 files |

Closed vocabulary gets a real type: `Final[str]` per name, `WALK_SCRATCH_COLUMNS: FrozenSet[str]` for the set, and named predicates (`is_trail_column`, `is_walk_scratch_column`, `is_shortest_path_hops_column`) instead of open-coded `startswith`. `varlen_rows.py`'s local `TRAIL_IDENT_COL` alias is gone.

Sweep proof — after the change only the definitions match:

```
$ git grep -n "\"__from__\"\|'__from__'" -- 'graphistry/**/*.py'
graphistry/compute/gfql/identifiers.py:14:WALK_FROM_COL: Final[str] = '__from__'
```
and identically for `__to__`, `__current__`, `__gfql_prev__`, `__gfql_edge_ident__`, `__gfql_trail_`, `__cypher_shortest_path_hops__` — one hit each, all in `identifiers.py`. That includes files outside the flagged five: `same_path/{bfs,multihop,where_filter,edge_semantics,native_shortest_path}.py`, `index/bindings.py` (its `_CURRENT` now aliases the shared constant), `cypher/{lowering,row_pushdown,shortest_path_aliases}.py` and two test modules.

`is_shortest_path_hops_column` is `isinstance(name, str) and name.startswith(...)`, which is exactly equivalent at all four former spellings (`label_node_hops` is `Optional[str]`), so the routing is unchanged.

## 6. `row/pipeline.py:1` — the file-level ask

### (1) symbol externing — see above, applied across the whole file, not only line 3946.

### (2) both-sides boundary testing

Commit **`93d26adc3`**, new file `graphistry/tests/compute/gfql/row/test_row_pipeline_boundaries.py`.

Every expectation is **hand-computed from the fixture edge list**. No expectation was generated by running the code, and cross-engine agreement is never the oracle — pandas, polars and cuDF are each checked against the same hand-written literal.

| boundary this file decides | inside | outside |
|---|---|---|
| trail tracking engages | plain var-length MATCH binds each edge once | `shortestPath` stays BFS |
| undirected flip-twin dedupe | a self-loop is ONE binding | a non-loop keeps BOTH orientations |
| same-edge return trip | over a PARALLEL edge: legal | over the SAME edge: no row |
| walk scratch columns | live inside the pipeline (`__current__` survives for the caller) | no `__gfql_trail_*` escapes |
| arithmetic input-null mask | computed NaN compares by IEEE (both operand sides) | a genuine null stays null; a non-arithmetic operand takes no override |
| hop-depth concat alignment | mixed depths keep every hop's rows | single depth needs no pad |
| polars var-length gate | served, pandas-identical | typed decline naming `#1748` |

**Cells: 37. Engines: pandas, polars, cuDF** (cuDF really executes here — cuDF 25.10 on this box, `TEST_CUDF=1`).

**Anti-vacuity: 12 of the 37 fail at this PR's merge-base** (`89abb3c11`, measured by running the same file in a base worktree with the new `identifiers` symbols shimmed):

```
test_plain_varlen_binds_each_relationship_at_most_once[pandas, cudf]
test_undirected_self_loop_is_one_binding_and_non_loop_keeps_both[pandas, polars, cudf]
test_cudf_range_varlen_keeps_rows_from_every_hop_depth[mixed_depth, single_depth]
test_computed_nan_comparison_is_false_not_null[pandas, cudf]
test_computed_nan_override_applies_to_the_right_operand_too[pandas, cudf]
test_polars_serves_unseeded_min_one_varlen_with_pandas_parity
```

The cells that pass at base are kept deliberately and say so in their docstrings: the `shortestPath` cells fence that the trail rework did **not** touch BFS; `test_non_arithmetic_operand_keeps_plain_null_semantics` fences the null-mask override against widening to non-arithmetic operands; `test_computed_nan_comparison_matches_nothing_in_positive_form` catches an override that made the comparison TRUE rather than FALSE; the tabled polars declines fence the routing in both directions.

**Mutation audit — existing pins first.** I broke 16 things the PR's pins claim to cover. Findings:

* Three mutations survived the PR's own `test_path_trail_semantics.py` and everything else in `graphistry/tests/compute/gfql`: **removing the pad-before-concat**, **letting the trail columns leak out of `_gfql_connected_bindings_state`**, and **never applying the right-operand arithmetic null mask**. All three are now red — the first only on the **cuDF** lane (pandas and polars null-fill on concat and cannot see it), which is why that pin is cuDF-only.
* **`test_path_trail_semantics.py` was 50% vacuous on polars**: `_assert_bag` returned early on any decline, so **12 of its 24 polars cells asserted nothing**, and a decline did not have to say what was missing. Fixed in place: `POLARS_DECLINED` now tables which `(fixture, query)` may decline (keyed on the pair — `-[*]->` serves on the acyclic DIAMOND and declines on the cyclic TRI) and `DECLINE_PHRASES` requires the message to name the gap. Membership is asserted **in both directions**, so a shape that starts declining *and* a tabled shape that quietly starts serving both go red. Verified by widening the polars var-length gate (`_vl_min >= 3` -> `>= 1`): exactly 5 polars cells go red -- `test_open_min_bound_parses_and_serves`, `test_single_alias_projection_keeps_path_multiplicity[d_range, p_parallel1]`, `test_trail_relationship_uniqueness[cycle_bounded_terminates, selfloop_range]` -- every one of which the old helper passed by returning early.
* Three mutations remain green and I believe they are **equivalent mutants**, not missing pins — reporting rather than papering over them:
  * `trail_tracking`'s `not shortest_path_mode` conjunct: in shortestPath mode the caller builds `step_cols` without the ident column, so the second conjunct is already false.
  * `step_pairs.drop_duplicates(...)`: with the ident column present a duplicate requires a repeated `(from, to, ident)`, which the flip-twin dedupe just removed.
  * `| used.isna()` in the trail mask: the pad is float NaN and `Series.ne(NaN)` is already True on both pandas and cuDF, so the disjunct is redundant *given the pad*.
* No pin was weakened, skipped or xfailed. I added no pin that an existing test already killed — the mixed-depth pandas/polars value was already pinned by `test_trail_relationship_uniqueness[midpattern_varlen_then_hop]`, so that cell is cuDF-only.

### (3) perf

Commit **`d1f9340dc`**. **This is not a pyg-bench signoff.** No perf number, complexity note or benchmark claim was added to any comment or docstring; `bin/ci_comment_density_guard.py` passes, and I removed two pre-existing perf claims from `lowering.py` while I was in there.

**Shapes this change can plausibly regress** (all on the row-binding lane; nothing else in the pipeline is touched):
1. var-length segments — one extra boolean mask per already-bound relationship, per hop;
2. any binding-rows query — the base edge frame gains an `__gfql_edge_ident__` column and `step_pairs` gains a third column plus a dedupe;
3. multi-element patterns — the state frame widens by one trail column per hop, so every subsequent merge/concat carries more payload;
4. polars `pl.concat(..., how="diagonal")` instead of `"vertical"`;
5. shapes the shrunk `#1787` gate now routes to the **native polars** lane instead of declining — a different code path, not a slowdown of an existing one.

**What I did measure** (local, in-process, 2 000 nodes / 8 000 random edges, median of 11, shared box under other load — indicative, not a signoff). pandas, `MATCH (x)-[*1..k]->(y) RETURN count(*)`, merge-base vs this head:

| k | base | head | delta |
|---|---|---|---|
| 1 | 30.3 ms | 32.1 ms | +6% |
| 2 | 36.0 ms | 46.5 ms | +29% |
| 3 | 60.0 ms | 90.6 ms | +51% |
| 4 | 141.0 ms | 297.6 ms | +111% |

Row counts are within 0.1% (668 733 vs 668 195 at k=4 — the difference *is* the trail fix), so this is not "more rows, more time": the cost scales with hop depth, as (1)+(3) predict. `MATCH (x)-[]->(y)` (18.5 → 18.3 ms) and `MATCH (x)-[]->(m)-[]->(y)` (0.99 → 1.07 ms) are flat. polars is flat across all four shapes (5.64 → 5.75 ms at k=2, 8.47 → 8.90 ms at k=3).

`d1f9340dc` recovers part of it by extracting `_gfql_drop_reused_relationship_rows` — one combined mask and one slice per hop instead of one frame copy per already-bound relationship (semantics-identical: a boolean AND of the same per-column predicates; `M15`/`M16` mutations on it are in the audit above). That takes k=3 to ~79 ms and k=4 to ~259 ms. **The remainder is a real regression I am not claiming away** — the dominant term is the widening state frame, not the mask passes.

**What is NOT measured, and needs the dgx box + pyg-bench:** LDBC SF1/SF10 and Pokec at real scale, specifically (a) the IC11/IC6 `-[*1..k]-` undirected shape, (b) graph-bench q3's `-[*1..k]->`, (c) the IS6 unbounded fixed point, (d) the seeded typed-hop fast path (this PR must not have perturbed its precedence), and (e) the shapes the shrunk gate newly routes to native polars, where the comparison is native-vs-decline rather than before-vs-after. I have not run any of those and make no claim about them.

## Gates

* `python -m ruff check graphistry bin` — clean.
* `bin/ci_comment_density_guard.py`, `bin/ci_type_hygiene_guard.py`, `bin/ci_cypher_surface_guard.py` — all pass. The comment guard needed real work after the master merge (the stack was 45 findings over baseline across 16 file/check pairs); every one was fixed by deleting narration or cutting to one line, and the baseline was re-locked **downward** with `--update-baseline` after the fix. No cap was raised.
* `python -m mypy graphistry` — 4 errors in 2 files, byte-identical to the merged base (pre-existing, in `polars/hop_eager.py` and `polars/degrees.py`).
* Full `graphistry/tests/compute`, failure list captured with `... 2>&1 | grep '^FAILED' | sort` (never through `tail`) and compared against `fix/gfql-1900-numeric-conformance` in a separate worktree. Two runs, because the base branch moved under me mid-review:
  * **consistent pair** — base `be4156e99` (109 failed / 10230 passed) vs this head (102 failed / 10301 passed): `comm -13` (only in head) **EMPTY**. The seven base-only entries are exactly the six anti-vacuity probe cells (the new pin file copied into the base worktree, cuDF cells skipped there) plus `test_polars_lane_completeness`, which the lane registration below satisfies.
  * **cross-check against the newest base** — base `dbf86d745` (102 failed / 10416 passed) vs the same head (102 failed): **both directions EMPTY, the two failure sets are byte-identical.**
  * Honest caveat: the head full run was taken at `262e91b0c`, so it predates the final base merge `e15e47284`. That merge adds only the base branch's own commits plus one line in `bin/test-polars.sh`; the final tree is covered by CI, which is green. Nothing this PR changes appears in either failure set.
* `test-gfql-core` caught a real gap on the first push: the new pin file mentions polars but was in no CI lane, so its polars cells would never have run. `5d4723ca3` adds it to `POLARS_TEST_FILES`. Exactly the anti-vacuity class this review is about, and the repo's own guard found it.
* `python-lint-types (3.8)` was failing on the first pushed head with `Incompatible import of "Literal"` in `row/pipeline.py` — a duplicate `typing`/`typing_extensions` import inherited from the **base branch**, not introduced here. Fixed at `262e91b0c`; #1902 has since fixed it on its own side (`aa9ef0c66`) and `2ee64b5b5` merges that forward, keeping the base branch's layout.

lmeyerov:
## Perf lane for #1904 — pandas var-length ladder: **k=4 REGRESSES ~+41%**, k=2/k=3 largely recovered

### Read this first: the protocol box was unavailable, these numbers are indicative

`dgx-spark` is **unreachable** — its Tailscale node key expired (`KeyExpiry 2026-08-15T23:05:48Z`, `Expired: true`, no fallback address; 100% packet loss, SSH timeout on 5 attempts, `tailscale ping` → `peer's node key has expired`). Tailscale is the only route (no LAN IP in `known_hosts`/history). Recovery needs a re-auth on the box (owner `percy@`) or an admin disabling key expiry for that node — I did not touch tailnet config.

So this lane was run **locally, not under `safe_run.sh`**, on a box that also had four other pytest suites running. It is a **within-session, interleaved A/B**, which is the part that survives a contended host; the absolute ms are **not quotable** and are **not** going into pyg-bench as board numbers. cuDF could not be measured at all.

### Method

Arms are frozen `git archive` trees, imported via `PYTHONPATH` (working tree is dirty, so archives-from-SHA were used deliberately):

| arm | commit | role |
|---|---|---|
| A | `aa9ef0c66` | merge base (verified: `git merge-base` of base `fix/gfql-1900-numeric-conformance` and head == `aa9ef0c66`) |
| B | `2ee64b5b5` | PR head |

Graph 2 000 nodes / 8 000 edges (seeded, `default_rng(42)`); query `MATCH (x)-[*1..k]->(y) RETURN x.id AS xid, y.id AS yid`, `engine="pandas"`. **18 slots per arm**, strictly interleaved A,B within each slot index; per slot warmup 2 + 17 timed runs; one process per slot; pinned `taskset -c 20,21`, all thread envs = 1. Stats over the 18 slot medians: bootstrap 95% CI (20 000 resamples), Mann–Whitney two-sided.

### Results

| shape | arm | slots | median ms | CI95 | pooled n | bimodal |
|---|---|---|---|---|---|---|
| k=1 | A | 18 | 31.10 | [30.48, 31.65] | 306 | no |
| k=1 | B | 18 | 31.83 | [31.23, 32.94] | 306 | no |
| k=2 | A | 18 | 116.61 | [112.91, 118.66] | 306 | no |
| k=2 | B | 18 | 121.24 | [118.33, 124.38] | 306 | no |
| k=3 | A | 18 | 178.33 | [176.29, 185.03] | 306 | no |
| k=3 | B | 18 | 178.76 | [166.46, 195.23] | 306 | no |
| k=4 | A | 18 | 406.18 | [390.68, 434.11] | 306 | no |
| k=4 | B | 18 | **572.39** | [550.60, 583.50] | 306 | no |

| shape | A→B | ratio CI95 | p (MW) | verdict |
|---|---|---|---|---|
| k=1 | ×1.024 (+2.4%) | [0.995, 1.061] | 0.017 | **DOES NOT REGRESS** materially — ratio CI crosses 1.0 |
| k=2 | ×1.040 (+4.0%) | [1.008, 1.079] | 0.0068 | **REGRESSES**, small (+4%) |
| k=3 | ×1.002 (+0.2%) | [0.920, 1.098] | 0.89 | **DOES NOT REGRESS** |
| k=4 | ×1.409 (+40.9%) | [1.317, 1.474] | 5.3e-07 | **REGRESSES** — large, unambiguous |

**No cell is bimodal** (306 pooled samples/arm/k; largest interior gap never reached the two-mode threshold). Pooled-sample ratios agree with the slot-median ratios throughout (+2.5% / +4.5% / +0.2% / +41.2%), so this is not a sampling artifact of the kind that produced the old polars q5/q6 "+29%".

### Value check — arms do **not** agree, by design

Within each arm the result is **bit-identical across all 18 slots** (1 distinct row-count tuple and 1 distinct content sha per arm). Between arms:

| k | A rows | B rows | delta |
|---|---|---|---|
| 1 | 8 000 | 8 000 | identical |
| 2 | 39 624 | 39 621 | −0.008% |
| 3 | 164 166 | 164 116 | −0.030% |
| 4 | 653 261 | 652 778 | −0.074% |

Strict value-identity **cannot** hold here and its absence is not a defect: dropping relationship-reusing paths is exactly what this PR implements. The head returns strictly fewer rows while doing more work. Flagging it explicitly because "values identical between arms" is normally the precondition for a perf comparison — here the right check is that the delta is small, one-directional, and matches TRAIL semantics, which it does.

### This contradicts the prior lane

The prior lane reported k=1 +6% / k=2 +29% / k=3 +51% / k=4 +111%, with `d1f9340dc` recovering k=3 to ~79 ms and k=4 to ~259 ms. Measured here at the head (which **contains** `d1f9340dc`): k=2 **+4.0%**, k=3 **+0.2%**, k=4 **+40.9%**.

- k=2 and k=3 are recovered far more completely than the prior lane credited — **k=3 shows no regression at all** (p=0.89). The +51% figure does not reproduce.
- k=4 is still badly regressed, but at ~+41%, not +111%/+83%.

Different graph generator and density (this one yields 653k rows at k=4), so magnitudes are not directly comparable to the prior lane's; the *direction and concentration at k=4* is the reproducible part. I am not smoothing the disagreement — the k=3 number in particular should not be carried forward.

### Why k=4, and what would recover it

`cProfile` at k=4, A vs B — the added time is **not** in a new GFQL function, it is spread across pandas block plumbing:

| pandas internal | A tottime | B tottime |
|---|---|---|
| `internals/managers.py:_merge_blocks` | 0.082 | 0.114 |
| `numpy.ndarray.copy` | 0.066 | 0.109 |
| `array_algos/take.py:_take_nd_ndarray` | 0.066 | 0.096 |
| `numpy vstack` | 0.064 | 0.092 |

`internals/concat.py:concatenate_managers` shows up in B's hot list at 0.163 s cumulative and is absent from A's. That is a **width-driven** cost: trail bookkeeping adds a column per hop, so every per-hop concat/take/copy runs over a frame k columns wider, and at k=4 the depth-4 tier is also ~4× the rows of depth-3 — the two multiply, which is why k=4 blows up while k=3 is flat.

Recovery candidates, cheapest first (from reading the diff, not yet measured):
1. **Skip the NaN pad on pandas.** The `.assign(<nan>)` padding pass before concat exists so cuDF gets aligned columns; `pd.concat` aligns natively. Gating it by engine removes a full O(Σ rows·k) materialization and the extra blocks that `_merge_blocks` then has to merge. Low risk, engine-local.
2. **Pack trail identity into one column** instead of one `__gfql_trail_i__` per hop. Removes the width growth at the root. Bigger change, touches both engines.
3. **Drop the two provably-no-op `drop_duplicates`** on the directed path (no-ops for directed input, but still full O(E) hashing).

(1) and (3) look like they recover a real fraction of the k=4 gap at low semantic risk. I have **not** implemented or measured any of them, and I changed no product code.

### Still unmeasured — the five named shapes

**LDBC IC11, IC6, q3, IS6, seeded typed-hop precedence, and the newly-native polars shapes remain NOT RUN.** They need the dgx box and the benchmark dataset roots (`/home/lmeyerov/data/graphbench*` are not present on this host). Their verdict is **INCONCLUSIVE**, not "fine".

### Bottom line

Under a perf-non-regression bar, **this PR is not clear to merge on k=4**: ~+41% with a tight CI and p≈5e-07, reproducible mechanism, and a plausible low-cost recovery path that has not been attempted. k=1 and k=3 are clean; k=2 is a marginal +4%. Everything here should be re-run under the real protocol on dgx before any number is treated as final, and the five named shapes still need to be run at all.

lmeyerov:
## Perf: pandas variable-length regression — measured, partially recovered

**Measurement context.** dgx-spark was unavailable (expired node key, then reserved by a colleague), so every number here is an **INDICATIVE local measurement** pending a protocol-grade run: frozen `git archive` arms at explicit SHAs, 18 slots/arm (k=4: 16+4 appended after one run was externally killed mid-slot; each slot is a self-contained interleaved block), strict slot-by-slot arm interleaving, warmup 2 + 17 reps/slot, CPU-pinned, thread env vars=1, ratio of slot medians with 20k-resample bootstrap 95% CIs + Mann-Whitney over slot medians. Ratios only — absolute ms are shown to size effects and are not quotable; nothing here goes into pyg-bench. Other agents' suites loaded the box for part of the window; interleaving protects the ratios and the CIs carry the noise.

Graph: synthetic graph-benchmark-schema set, 20k persons, ~40k FOLLOWS (mean out-degree 2, 3 self-loops, fixed seed), `[*1..4]` ≈ 1.53M result rows. **The prior lane's magnitudes did not reproduce at this density** (they reported k=4 ×1.409 on a ~4×-row-growth graph; this graph grows ~2.2×/hop). What does reproduce: the regression concentrates at k=4, and each arm is bit-deterministic.

Arms: `base` = aa9ef0c66, `head` = 2ee64b5b5 (PR before fix), `fix` = head + f0dc66c8b.

### Ratios (pandas, `MATCH (x {node_type:'Person'})-[*1..k]->(y) RETURN x.node_id, y.node_id`)

| k | base→head (the regression) | base→fix (after recovery) | head→fix (my fix's effect) | verdict for pushed head |
|---|---|---|---|---|
| 1 | ×1.044 [1.023, 1.063] p=0.0011 | ×1.052 [1.023, 1.076] p=0.00075 | ×1.008 [0.986, 1.025] p=0.76 | **REGRESSES** ~4–5%, not recovered |
| 2 | ×1.017 [0.986, 1.063] p=0.11 | ×0.995 [0.965, 1.040] p=0.67 | ×0.978 [0.951, 1.010] p=0.14 | **DOES NOT REGRESS** |
| 3 | ×1.042 [1.024, 1.057] p=1.6e-05 | ×1.023 [0.999, 1.041] p=0.0082 | ×0.982 [0.968, 0.995] p=0.0099 | **INCONCLUSIVE** (residual CI touches 1.0) |
| 4 | ×1.076 [1.057, 1.104] p=1e-06 | ×1.060 [1.036, 1.073] p=1.4e-05 | ×0.985 [0.955, 0.995] p=0.00076 | **REGRESSES**, residual ~5–6% |

k=4 replication in an independent later cell (18 fresh slots): base→fix ×1.053 [1.041, 1.062] — consistent. Bimodality flags traced to single contended slots (e.g. one 2.75s base slot among 2.03–2.18s), not genuine bimodal populations; medians-of-medians absorb them.

### What was found and fixed (f0dc66c8b)

cProfile on the head arm reproduces the reported block-plumbing signature (`concatenate_managers`, `_merge_blocks`, `_take_nd_ndarray`, `ndarray.copy`, `vstack`) inside `_gfql_multihop_binding_rows` / the bindings state builder. Three value-identical removals:

1. Per-hop `rename(columns={ident: trail_col})` copied every block of the growing frame; now a metadata-only `.columns` relabel (frames are walk-owned — each is the fresh output of the preceding merge/slice/drop).
2. The pre-concat NaN pad ran unconditionally; `pd.concat` aligns and NaN-fills mismatched schemas natively, so it now runs only for cuDF frames (cuDF expression unchanged). Verified on a real GPU (RTX 3080 Ti): cuDF varlen still runs; k1–k3 hashes identical head-vs-fix and equal to the pandas hashes. This box's cuDF also accepts mismatched-schema concat, but the pad is conservatively retained there so that lane is untouched.
3. The trail-mode `step_pairs.drop_duplicates()` is skipped with a proof: it deduped on exactly `[from, to, ident]`, which the self-loop-twin dedup immediately above already made unique — a duplicate cannot arise. The shortest-path (non-trail) dedup is kept. Plus two `.copy()`-after-`drop()` removals (drop already returns a new frame).

A fourth candidate — shedding the dead trail columns before the alias-attach merges — was implemented, measured **neutral** (fix→shed ×0.998 [0.989, 1.006] at k=4; ×0.992 [0.972, 1.012] at k=1), and reverted so the branch carries only measured recoveries (commits dbd2dbe1b/fe55e8c40 document this).

### What remains, honestly

The unrecovered ~4–6% at k=1/k=4 is the trail identity **payload**, not copy waste: one extra int64 column through every per-hop merge, the per-hop reuse mask, and identity columns riding the downstream binding merges. That is the cost of the TRAIL semantics this PR exists to fix. Next lever if a protocol run shows it matters at real density: packing multi-hop identity into a single column (invasive; not attempted here).

### Value identity (the correctness bar: head unchanged by MY fix; head ≠ base is this PR's intent)

- Row-order-independent sha256 over all columns: **identical head-vs-fix** for k=1..4 (k4 hash `700af2021249ccf6`, 1,526,741 rows) and for q1–q9 + a shared-arm probe on the 20k/100k board datasets. Every timed slot re-hashed its result: one hash per arm per cell across all slots.
- head vs base differs exactly as the TRAIL fix predicts: k1 identical; k2 −3, k3 −33, k4 −210 rows (repeated-edge paths removed), each arm bit-deterministic.
- Internal pre-projection frame: same column order and dtypes head-vs-fix.
- cuDF (GPU): k1–k3 hashes identical head-vs-fix.
- The exact push tree (fix + master fold-forward) re-verified: values identical to the measured fix arm on all 14 checks.

### Gates (run from inside the worktree)

- ruff clean; comment-density, cypher-surface, type-hygiene guards clean. The master fold-forward resolved comment conflicts to the base's canonical one-liners; ratchet baselines resolved to per-file minima (no cap raised); `lowering_py_max_lines` set to the measured `wc -l` of the merged file.
- mypy: same 4 pre-existing errors before/after (none in changed files).
- Full `graphistry/tests/compute`: failure set **byte-identical** across master (bb1fa0d91), PR base aa9ef0c66, head, head+fix, and the final push tree — 102 pre-existing environmental failures (cudf/libnvrtc, umap, dask, polars-version), zero attributable to this branch.

dgx protocol run still owed before any of these ratios are quotable outside this thread.

Pushed head: fa11e0f3f (perf commit f0dc66c8b + neutral-shed revert fe55e8c40 + master fold-forwards; CI now fireable — the branch was CONFLICTING with zero runs before this push).

Issue #1906, opened 2026-08-15
GFQL Cypher: undirected unbounded [*] row projection off the trail lane (impossible rows; count-twin break)


lmeyerov:
Found by the round-006 paths re-probe (brute-force trail enumerator as oracle, 140 Cypher texts × 2 engines × 2 repeats @ d7cf23d1b; scripts in session scratchpad probe-r6-paths/, repro at repro_f1.py).

**Shape:** undirected + unbounded (`to_fixed_point`) var-length, plain row projection, pandas. polars declines typed (safe).

4-cycle a→b→c→d→a:
| query | oracle | actual |
|---|---|---|
| `MATCH (x {id:'a'})-[*]-(y) RETURN y.id` | 8 rows | **4** |
| `... RETURN count(*)` | 8 | 8 ✓ |
| `... RETURN y.id, count(*)` | 4 groups × 2 | ✓ |
| `MATCH (x {id:'a'})-[*1..4]-(y) RETURN y.id` | 8 | ✓ |

So the plain row lane is the only wrong lane — the count and grouped twins are trail-exact, and the *bounded* twin is exact. Breaks the "one MATCH = one cardinality" invariant #1903 established.

**Worse than multiplicity — impossible rows.** With a second element on a triangle: `MATCH (x {id:'a'})-[*]-(y)-[]->(z) RETURN z.id` → `[a,b,c]`; oracle `[a,c]` (reaching `b` requires reusing an edge). The `count(*)` twin is 2 (correct).

Reproduces on every cyclic fixture, seeded and unseeded (TRI 3 vs 6, C4 4 vs 8, C4CH 4 vs 25, C5CH 5 vs 30, SLPAR 2 vs 43, DIASL 5 vs 17; unseeded TRI 3 vs 18, C4 4 vs 32).

**Provenance (A/B vs master 25215ce5c):** the row lane returns the same wrong 4 on master — the row values are NOT a #1903 regression. What changed is the neighbourhood: on master the adjacent `count(*)` twin *raised a typed error* on C4/C4CH (SLPAR returned the old walk value 11); #1903 moved the count lane onto trail semantics and left the row lane on the source-table path, so the shape now presents a correct number beside a silently wrong row set.

[...]

Issue #1905, opened 2026-08-14
GFQL Cypher: count(*) trail uniqueness + list-param equality leak (round-006 umbrella)


lmeyerov:
Round-006 cartesian/params/labels probe (~85 queries × both engines @ d7cf23d1b; scripts in session scratchpad probe-r6-cart/).

**SILENT-WRONG (both engines):**
1. **Bare `count(*)` ignores relationship uniqueness the row lane enforces** — one MATCH, two cardinalities again but in the fixed-hop/comma domain: self-loop fixture `MATCH (a)-->(b)-->(c)` rows=1 vs count(*)=2; branching comma `(a)-->(b), (a)-->(c)` rows=2 vs count(*)=9 (out-deg² products incl. illegal same-edge pairs); grouped branching count equally wrong. `count(a)`/`sum`/`collect`/`count(DISTINCT)`/chain-grouped counts are all correct — only the bare-count lane rides the edge product. A guard exists for aggregates needing path multiplicity (typed decline) — count(*) slips around it.
2. **List-valued `$param` in `=`/property-map runs as isin** (`WHERE n.v = $l`, l=[30,70] → matches members; oracle: scalar=list false → 0 rows). Literal spelling differs too (pandas 0 rows conformant; polars typed error) — param substitution into filter_by_dict semantics leak.

**Typed-gap deviations:** 3. Nonexistent node label = hard GFQLSchemaError column-not-found (incl. OPTIONAL arms, where openCypher says all-null rows); relationships correctly 0-row. Pandas OPTIONAL-arm error leaks binder-internal column name. 4. String-param vs numeric column: pandas typed; polars RAW polars.ComputeError (error-contract divergence).

**NIE inventory:** comma-cartesian is node-only (disconnected relationship arms decline; `MATCH (a:X) MATCH (b:Y)` rejected while comma form works); polars fall-offs (IN $empty/mixed lists, XOR over labels, labels() projection).

QUIET inventory: shared-alias joins with correct trail semantics, label conjunction/OR/NOT, params in all positions incl. shadowing, missing-param typed. Fix order: 1 → 2 → 4 (error contract) → 3 (needs semantics decision: strict-schema vs openCypher missing-as-absent — flag for owner).

lmeyerov:
**Umbrella verified served on master `e6625ed28`** — every item is either fixed or split into its own open issue; recommending close after review.

[...]

PR #1907, opened 2026-08-15
fix(gfql): count(*) trail uniqueness + list-param equality + undirected [*] row lane (#1905, #1906)

lmeyerov:
Stacked on #1904. Closes the count-twin invariant on both lanes.

**#1905-1a — `count(*)` on chain shapes.** The two-hop count fast path answered `Σ_b indeg(b)·outdeg(b)`, which counts the illegal pair `(r,r)` whenever a self-loop at `b` is usable on both arms. Now subtracted in every branch (dense-fact, bincount, fused-polars, eager-polars, pandas); zero cost on self-loop-free frames (one mask + early return), and the O(1) fact branch declines the shortcut rather than assuming zero when `self_loops` is unknown.

**#1905-1b — branching comma patterns.** Per-arm binding rows carried no edge identity (anonymous arm edges emit none), so an edge fitting two arms bound twice → `Σ d²` instead of `Σ d(d−1)`. Anonymous arm edges are now named internally so identity survives the join, cross-arm duplicates are dropped, synthetic columns removed before the post-join chain. **Engages only when arms can share an edge** — provably-disjoint arms (different relationship filters, i.e. the graph-benchmark two-star) keep their existing fast grouped-count/rows path untouched. Also fixed the three-star row projection (13 rows → oracle 0). A first attempt that re-routed these shapes to the general lowering was discarded: it fixed counts but moved ~60 connected-join contract pins off the lane they exist to pin.

**#1905-2 — list-valued `$param` equality.** A list reaching `filter_dict` means `isin`; `=` now maps list/tuple/set/dict to `never_match()` (and `!=` to `notna()`), including property-map entries and edge props. `IN $list` untouched. polars gets a native `NeverMatch` lowering so it answers instead of declining.

**#1905-3 — polars error contract.** Scalar-equality dtype mismatch leaked a raw `polars.ComputeError`; now the same typed E302 as pandas.

**#1906 — undirected unbounded `-[*]-` row lane.** The `to_fixed_point and undirected` early-return is now scoped to shortestPath/allShortestPaths only (they bind one representative route per pair by design). Plain `-[*]-` rides the trail-filtered binding lane: C4 seeded 4 → 8 rows (matching its count twin), triangle two-element `[a,b,c]` → `[a,c]` — no impossible rows.

52 new pins + 6 strict xfails (hand oracles, both engines). 12 existing tests corrected by hand arithmetic quoted in-file (two-star `d²→d(d−1)` family, plus an independent merge-join oracle made trail-exact). Deferred to owner: nonexistent-label hard-error semantics (pinned xfail). Gates: full-tree failure set byte-identical, TCK 4144P/0F, typecheck 328 files, ruff+guards green (surface 9772), fast-path engagements all still fire.


lmeyerov:
**GPU sweep (dgx, @4bcdc8949): 9050 passed / 53 skipped / 33 xfailed / 0 failed** — +52 exact vs prior tip; GPU-probe gated (cudf device groupby + polars GPUEngine raise_on_fail) so it's a genuine GPU run.

**Perf A/B on the real board runner** (pyg-bench `matched_q1_q9/gb_gfql.py`, board datasets 20k/100k, `--binding cold`, warmup 2, 15-41 runs, A/B interleaved so drift can't favor a side, medians of per-rep medians). Counts bit-identical across commits everywhere (board FOLLOWS has zero self-loops), so this is pure overhead:

| cell | @4bcdc8949 | @d7cf23d1b | Δ |
|---|---|---|---|
| **polars q8 @100k** | 15.81 ms | 13.76 ms | **+14.9%** |
| polars q8 @20k | 2.34 ms | 2.02 ms | +15.8% |
| pandas q8 @100k | 90.55 ms | 88.83 ms | +1.9% |
| polars q9 @100k | 38.48 ms | 36.64 ms | +5.0% |
| q5/q6/q7 (two-star) both engines, both scales | — | — | **unchanged** |

Two-star protection confirmed empirically, as designed: the board's arms are `HAS_INTEREST` vs `LIVES_IN`, provably disjoint, so `_connected_join_trail_arms` returns None and the fused fast paths still serve.

Attribution: `_two_hop_equal_domain_dense_total` runs `(src_arr == dst_arr).sum()` unconditionally — an extra full O(E) pass, engine-independent in absolute terms (~+2 ms at 2.42M edges), so the % hit lands hardest on the fastest lane, which is the board's one contested cell vs Kuzu (~9.7 ms @100k). **Not merging on this head**: a perf follow-up is in flight to make the correction near-free (fuse into the existing degree pass / use the precomputed `DegreeFact.self_loops`), with the correctness pins and fast-path engagement as hard gates. The correction itself stays — reverting to the wrong count is not on the table.

lmeyerov:
**GPU sweep @3cf304b81: 9058 passed / 0 failed** (= 9050 + the 8 new multi-block pins), GPU-probe gated (cuDF device op + groupby, polars `GPUEngine(raise_on_fail=True)`).

**Board re-measure, three-way interleaved** (read-only frozen `git archive` builds, `chmod -R a-w`, one slot at a time through safe_run, `--network none`, box idle; 11 slots per build @100k / 7 @20k in balanced A-B-C / C-B-A / B-A-C orders; median of per-slot medians). **All returned values bit-identical across all three builds, every cell, every slot** (q8@100k 58,431,994; q9@100k 45,562,239; q5/q6/q7 row-sets hash-identical) — no correctness alarm.

| cell | A 3cf304b (fix) | B 4bcdc89 (regressed) | C d7cf23d (floor) | A vs B | A vs C |
|---|---|---|---|---|---|
| **polars q8 @100k** | **14.77** | 15.64 | 13.59 | **−5.6%** | **+8.6%** |
| polars q9 @100k | 38.19 | 37.76 | 36.39 | +1.1% | **+5.0%** |
| pandas q8/q9 @100k | 89.45 / 164.52 | 90.94 / 163.85 | 90.52 / 167.24 | — | at/below floor |
| polars q8 @20k | 2.42 | 2.42 | 2.08 | +0.0% | +16.1% |

Per-slot distributions barely overlap between A and C, and an independent round-3 reproduced it standalone (A 14.49 / B 15.42 / C 13.26).

**Verdict: real but partial.** #1905 cost +2.04 ms on the contested cell; this recovers 0.87 ms (43%), leaving +1.17 ms over floor. Vs the warmed Kuzu ~9.7 ms goalpost: floor 1.40× → regressed 1.61× → now 1.52×.

Two findings beyond the brief, both now being chased: (1) **q9 is untouched** (+5.0% / +1.80 ms over floor, A≈B) even though it is also a two-hop `count(*)` — so the self-loop pass is not the whole cost of #1905; (2) the cache-blocking is **scale-gated** — zero recovery at 20k. pandas is clean at both scales, so the entire residual is polars-side.

Measurement-integrity note: an initial q5/q6/q7 pass showed a "+40% on q6" that was purely a slot-order artifact (unbalanced A/C interleaving against a machine that slowed in later slots); re-run balanced it vanished, and the two-star cells are confirmed unchanged. polars q5/q6 are intrinsically bimodal within every slot (~6.8-16.9 ms in both builds), so their per-slot medians are not usable statistics — pooled distributions are shape-identical.

lmeyerov:
**GPU sweep @3ec5d1be6: 9060 passed / 0 failed** (GPU-probe gated: cuDF device sum, cupy device count, polars `GPUEngine(raise_on_fail=True)`).

**Board A/B/C @3ec5d1be6 vs 3cf304b81 vs floor d7cf23d1b** — 27 slots (9/build), 3 balanced rounds with rotated order, frozen read-only builds, `--network none`, Mann-Whitney over per-slot medians. **All values bit-identical across builds and slots** (q8@100k 58,431,994; q9@100k 45,562,239; q5/q6/q7 identical) — no correctness alarm.

| cell | NEW | prev head | floor | vs prev | vs floor |
|---|---|---|---|---|---|
| **polars q9 @100k** | **34.91** | 38.14 | 36.44 | **−8.5% SIG** | **−4.2% SIG (beats floor)** |
| **polars q9 @20k** | **8.09** | 9.53 | 9.10 | **−15.1% SIG** | **−11.2% SIG (beats floor)** |
| polars q8 @100k | 15.10 | 14.61 | 13.78 | +3.4% (ns) | **+9.6% SIG** |
| pandas q8 / q9 @100k | 92.20 / 169.92 | 91.88 / 169.13 | 88.93 / 162.61 | ns / ns | +3.7% / +4.5% SIG |
| polars q8 @20k | 2.50 | 2.51 | 2.46 | ns | ns |

Per-slot spreads on the contested cell show NEW and prev interleaving completely while the floor sits cleanly below both (floor max 14.25 < both medians).

**Verdict: closed on q9, open on q8.** The fused-lane change does exactly what it claims for the non-dense two-hop count lane — q9 now beats its pre-regression floor at both scales with disjoint distributions. The dense q8 lane is unchanged-to-nominally-worse and still +9.6% over floor (1.56× vs the warmed Kuzu ~9.7 ms goalpost); pandas still carries +3.7/+4.5% and this commit moved neither. Confirms the local x86 prediction rather than contradicting it.

q5/q6/q7 two-star cells unchanged (judged on pooled p25/p50/p75 — a −10.8% on polars q6 in the per-slot table is the known bimodality artifact, not a gain). 20k q8 shows no measurable regression in this run and the prior session's 2.08 floor did not reproduce, so no 20k q8 delta should be quoted in either direction.

One final targeted pass is in flight to account for the dense-lane and pandas residuals; if they prove irreducible without giving up the correctness fix, this merges on documented cost — the silent-wrong count is not coming back.

lmeyerov:
**FINAL board lock @9fb1b77dc** — GPU sweep **9068 passed / 0 failed** (probe-gated: cuDF device buffer op + polars `GPUEngine(raise_on_fail=True)`). Board: 54 slots per scale (**18 per build**, doubled from the prior protocol because MWU at n=9 lacks resolving power for a sub-3% effect), frozen read-only builds verified by sha256 against their git blobs, `--network none`, bootstrap CIs + Mann-Whitney over slot medians. **Values bit-identical across all three builds and all 54 slots**, both scales.

**Headline (release-quotable), polars q8 @100k, 2.42M-edge FOLLOWS, cold binding, bind excluded:**
> **14.14 ms**, 95% CI **[13.98, 14.42]** — **1.038× the pre-regression floor** (13.62), **1.46× the warmed-Kuzu goalpost** (9.7 ms).

| cell | NEW 9fb1b77 | prev 3ec5d1b | floor d7cf23d | vs prev | vs floor |
|---|---|---|---|---|---|
| polars q8 @100k | **14.14** | 14.81 | 13.62 | **−4.5%, p=1.1e-4 SIG** | +3.8%, SIG |
| polars q9 @100k | 34.81 | 34.96 | 36.40 | ns | **−4.4%, p<1e-6 (beats floor)** |
| polars q9 @20k | 8.00 | 7.98 | 9.19 | ns | **−12.9%, p<1e-6 (beats floor)** |
| polars q8 @20k | 2.39 | 2.35 | 2.24 | ns | +6.7%, ns |

Per-slot spreads: NEW and prev overlap only at NEW's top-2 vs prev's bottom-2 slots; direction stable across both independent batches.

**Net:** this recovers a bit over half the polars-q8 regression the #1905 correctness stack introduced (+8.7% → +3.8% over floor), values identical throughout, and q9's wins over floor are intact at both scales. It does not return q8 to the floor, and does not change the Kuzu verdict on that cell.

Two measurement-integrity notes: (1) the floor reproduced exactly (batch 1 alone: 13.781 vs the prior session's 13.78), but **the prev-head number drifted between sessions** (15.10 → 14.81), so the previously posted "+9.6% over floor" for 3ec5d1be6 is better stated as **+8.7%** — all verdicts here are within-session and interleaved, so the drift doesn't touch them; (2) the board's −0.67 ms is *larger* than the local −0.28 ms and the local value sits near the lower CI edge — consistent, but the board should not be quoted as confirming −0.28 specifically.

Carried forward as follow-ups, not blockers: polars q8 @20k nominally +6.7% (not resolvable at this n) and a small consistent ~+2% pandas cost @100k that this commit doesn't address (the two pandas "SIG" cells are bimodal with CIs straddling zero — pooled distributions are the honest read).

graphistry/compute/gfql_fast_paths.py
```diff
@@ -560,26 +570,30 @@ def _two_hop_equal_domain_degree_counts(
...
         filtered_edges = (
             edge_domain.lazy()
             .join(domain_ids, left_on=src_col, right_on=node_col, how="semi")
             .join(domain_ids, left_on=dst_col, right_on=node_col, how="semi")
         )
-        in_counts, out_counts = _lazy_collect_all([
+        in_counts, out_counts, loops = _lazy_collect_all([
             filtered_edges.group_by(dst_col).len("__in_count__"),
             filtered_edges.group_by(src_col).len("__out_count__"),
+            filtered_edges.select((pl.col(src_col) == pl.col(dst_col)).sum().alias("__self_loops__")),
```
lmeyerov: extern symbol on these + others?

graphistry/compute/gfql_fast_paths.py
```diff
@@ -2567,6 +2706,86 @@ def _dense_interval_from_fact(fact: Optional["ColStatsFact"]) -> Optional[Tuple[
...
+    return total
+
+
+def _gather_sum_and_self_loops(
+    in_counts: "ArrayLike", src_arr: "ArrayLike", dst_arr: "ArrayLike",
+    *, xp: "ArrayNamespace", blocked: bool,
+) -> Tuple[int, int]:
+    """``(sum_e in_counts[src(e)], self-loop count)`` from ONE blocked pass over src.
+
+    The correction's cost here is the SECOND read of dst, and that read is bandwidth
```
lmeyerov: violating our conventions like what goes into comments vs into code, how to handle perf bits, etc... audit full pr

graphistry/compute/gfql_fast_paths.py
lmeyerov, line 1: this file likely comes with positive+negative test amplification, till convergence, ogligations

graphistry/compute/gfql_unified.py
lmeyerov, line 1: same pos/neg proof obligations , amplification, & convergence

graphistry/compute/gfql_unified.py
```diff
@@ -612,6 +612,118 @@ def _optional_arm_start_nodes(
     _execute_two_hop_count_fast_path,
 )
 
+_TRAIL_EDGE_IDENTITY_COL = "__gfql_edge_index_0__"
+_TRAIL_EDGE_ALIAS_PREFIX = "__gfql_trail_arm_"
```
lmeyerov: shouldn't these be externed to managed central files with others?

lmeyerov:
## Full-PR audit — you were right, and it was wider than the five threads

Head `f3bec40a0`. Audited the whole `36b7d26e5..HEAD` diff, not the flagged lines. **21 comment sites / 64 added comment lines / 21 new suppressions / 18 new `cast()`** in the submitted diff. **16 of the 21 sites were violations.** After remediation the same diff carries **8 sites / 15 comment lines / 3 suppressions / 5 casts**, and one surviving issue-ref line.

### 1 + 5 — externing, swept not spot-fixed

`lowering.py:1272` was writing `NODE_IDENTITY_COLUMN if isinstance(target, ASTNode) else "__gfql_edge_index_0__"` — the constant on one arm, the bare literal on the other, in one expression. That is the partial-externing pattern this stack keeps hitting.

| literal | constant | home (why) | call sites updated |
|---|---|---|---|
| `__gfql_edge_index_0__` | `EDGE_IDENTITY_COLUMN: Final[str]` | `gfql/same_path_types.py`, beside `NODE_IDENTITY_COLUMN` — same concept, same importer set | 5 (lowering ×3, row/pipeline, row/entity_props) + gfql_unified |
| `__gfql_trail_arm_` | `TRAIL_ARM_EDGE_ALIAS_PREFIX: Final[str]` | `gfql/identifiers.py`, beside `HIDDEN_ALIAS_COLUMN_PREFIX` | 3 |
| `__in_count__` | `_TWO_HOP_IN_COUNT_COL: Final[str]` | moved to `gfql_fast_paths.py` top — the constant existed at line 1628, **after** 4 of its own uses | 6 |
| `__out_count__` | `_TWO_HOP_OUT_COUNT_COL: Final[str]` | same | 6 |
| `__self_loops__` | `_TWO_HOP_SELF_LOOP_COL: Final[str]` | same (new) | 3 |

Sweep proof — `git grep '"<literal>"' -- graphistry ':!graphistry/tests'` returns only the definition for each of the five. (`generate_safe_column_name.py:27` still shows the string inside a docstring *example* of the naming scheme; it is prose about the convention, not a use.)

Also `Final[str]` on the pre-existing `identifiers.py` constants, and the filter-dict domain now has a real name (`FilterDict` in `compute/typing.py`) instead of `Dict[str, Any]` + suppression.

### 2 — the convention audit (site-by-site)

| site | verdict | carrier |
|---|---|---|
| `gfql_fast_paths.py` `_two_hop_equal_domain_degree_counts` docstring | **perf claim** (`+0.4..0.5ms at 200k/2.4M edges`) | deleted; value contract kept, `test_two_hop_count_*` pins |
| same, pandas branch comment | narration + null fact | deleted; `_self_loops_in_all_domains` one-liner + `control_null_endpoint` |
| `_eager` / `_self_loop_edges` (+6 `hygiene-ok` casts) | new suppressions | **restructured** into typed polars/indexable arms per `compute/typing.py`'s own rule; functions gone |
| "Arm frames are projected to endpoints…" | narration | helper `_edges_passing_both_relationship_filters` |
| `ids_of` comment | **perf claim** (`wasted O(N) pass`) | deleted; `test_two_hop_count_ignores_duplicate_rows_in_a_node_domain` |
| fused-lane decline comment | narration | name `correction_needs_both_relationship_filters` + a decline pin |
| `loops_lf` comment | **perf claim** | deleted; sub-plan named `self_loops_in_all_domains_lf` |
| `_edge_cols_bounds_polars` docstring | **perf claim** (`+0.60/+0.81/+1.27ms`, `-0.28ms 95% CI [-0.45,-0.16]`, `225 reps`, `43% of residual`) | deleted |
| `(False, 0)` null comment | narration | deleted; `control_null_endpoint` |
| `_edge_cols_bounds_scan` docstring | **perf claim** (`+1.8ms`) | deleted |
| `_TWO_HOP_SELF_LOOP_BLOCK` 6-line `#:` | **perf claim** (sweep record `16k..1M`) | deleted; value-identity carried by the existing monkeypatch pin |
| `_self_loop_count` / `_gather_sum` docstrings | **perf claims** (`-0.88ms`, `-3.76ms CI [-6.2,-2.5]`) | deleted |
| `_gather_sum_and_self_loops` docstring | **perf claim** — 12-line essay (`1.16..1.29ms`, `9.68 vs 9.40ms`) | deleted; value-identity kept in one line |
| `scanned_self_loops` comment | restates `Optional[int]` | deleted |
| 7-line TRAIL comment in the kernel | narration | deleted |
| "…before the buffer exists" | **factually wrong** — `xp.subtract(..., out=buf)` writes `buf`, never `src_arr`/`dst_arr`, so the ordering it claims to protect is not a constraint | reverted to the pre-existing one-liner |
| dispatcher 8-line comment | **perf claim** (`+4.9ms`) | deleted; name `counting_lane_derives_correction` |
| eager-polars fallback comment + duplicated logic | narration + DRY | folded into `_self_loops_in_all_domains` |
| `lowering.py` shortestPath comment | narration + `#1903`/`#1906` | helper `_binds_one_route_per_pair_undirected` |
| `predicates.py` `_schema_memo` | **perf claim** (hot lanes) | deleted |
| `predicates.py` NeverMatch / E302, `registry.py` `self_loops` | multi-line + issue-as-rationale | one-liners, refs dropped |

`gfql_unified.py`'s 5 remaining `hygiene-ok` casts were also restructured (local narrowing for the Optional edge frame, a typed polars arm, `isinstance` narrowing on the AST clone seam). **New suppressions 21 → 3**: one pre-existing `type: ignore` on a line the PR edited, and two `hygiene-ok` on `pl.Expr.cast` — a guard false positive, since it matches any call named `cast`.

### 3 + 4 — amplification

**Mutation-audited the existing pins first.** 20 mutations; **7 survived** everything the PR shipped plus `test_lowering.py`: the degree-fact correction, the `self_loops is None` guard, `build_degree_fact`'s own count, the both-relationship-filter probe, the node-domain restriction, and the shortestPath veto. Those were decorative coverage, not covered lanes.

Round 2 added 8 more mutations; 5 survived round 1's new pins (count-vs-flag on two lanes, the fused decline, the identity-column reuse, the empty scalar sink).

**36 new cells. 28 fail at the merge-base; 8 are declared controls** (the `allShortestPaths` parser decline, empty-edge-frame ×3, null-endpoint ×3, empty-sink). Every oracle is hand-enumerated in-file over ordered tuples of distinct relationships — no cross-engine agreement anywhere. The count-correction cells run **pandas, polars and cudf**.

Two findings worth your call rather than a fabricated pin:

- **`_polars_scalar_int`'s empty-frame branch is unreachable.** Both call sites are `select` of an aggregate, and polars returns height 1 for those (verified). It is dead defensive code.
- **The shortestPath veto has no end-to-end observable.** `_forces_relationship_multiplicity_projection_bindings` returns False earlier because a `shortestPath` pattern binds a path alias, which is not an `ASTNode`/`ASTEdge`. I could not construct a query where the scoping changes an answer, so `M14` (deleting the veto call) stays unkilled. The scoping itself is pinned at compile level.

**Round 2 is NOT QUIET** — it found 5 new gaps, 4 now closed. **Convergence not reached**; a third round is owed.

### Gates

- `ruff` clean. **All three guards exit 0** — including the comment-density guard, which this branch had never run: it failed on 16 file/check pairs when I merged master, and is now fully green.
- `mypy`: 4 errors, byte-identical to the control (`hop_eager.py` ×2, `degrees.py` ×2) — no new errors.
- Full `graphistry/tests/compute` vs a control built from the merge-base + master: **102 vs 102, failure sets byte-identical**, zero regressions. After the base moved I rebuilt the control at the new base and re-ran `tests/compute/gfql`: **93 vs 93, byte-identical**. All are pre-existing cudf failures from a missing `libnvrtc.so.12` in this environment.
- `lowering_py_max_lines` 9734 → 9746: the convention-mandated extraction of the seeded typed-hop guard into a named helper costs a signature and a docstring. No other baseline was raised.

### Perf — measured vs not

I deleted the prose but **did not re-measure any of it**, and two changes could plausibly move a cell:

- `_two_hop_trail_illegal_pairs` now early-exits through `_self_loops_in_all_domains(frame, ())` instead of an inline filter, and the eager-polars fallback calls the shared helper rather than its own loop. Same reductions, one extra Python call per query.
- `ids_of` losing `.unique()` is the PR's own change, not mine; its value-equivalence is now pinned, its cost is not.

**Not measured**: any wall-clock effect on the 200k/2.42M-edge equal-domain and distinct-domain count lanes. To claim no regression you would need the two-hop count lanes on dgx at those two sizes, interleaved reps, pandas + polars, against `36b7d26e5`. I am not claiming it.

Nothing merged, no force-push, base unchanged, no threads resolved, no test weakened or skipped.

lmeyerov:
## Perf lane for #1907 — polars count lanes **DO NOT REGRESS**; **pandas q9 regresses +4.1% post-lock** (CI excludes 1)

### Read this first: two limits on what this can and cannot settle

1. **`dgx-spark` is unreachable.** Its Tailscale node key expired (`KeyExpiry 2026-08-15T23:05:48Z`, `Expired: true`, no fallback address; 100% packet loss, SSH timeout on 5 attempts). Tailscale is the only route. Recovery needs a re-auth on the box (owner `percy@`) or an admin disabling key expiry — I did not touch tailnet config.
2. **The graphbench dataset roots are not on this host** (`/home/lmeyerov/data/graphbench`, `-100k` do not exist). So this lane runs a **synthetic Person/FOLLOWS proxy**, not the Pokec root.

Consequence, stated plainly: **the specific blocking question — does polars q8 @100k move outside `[13.98, 14.42]` — is NOT answered here.** My polars q8 @100k lands at ~1.17 ms against the board's 14.14 ms, i.e. a ~12× different cost profile, so a different path dominates on the proxy and the board CI cannot be checked against it. That cell needs dgx + the Pokec root. What this lane *does* answer is the relative question: **did the commits after the `9fb1b77dc` lock move these query shapes at all.**

### Method

Arms are frozen `git archive` trees imported via `PYTHONPATH` (the working tree is dirty, so archives-from-SHA were used deliberately):

| arm | commit | role |
|---|---|---|
| M | `e15e47284` | merge base (verified via `git merge-base`; note it is *behind* the current #1904 tip `2ee64b5b5`) |
| L | `9fb1b77dc` | the board-lock commit — **inside this PR's own range**, with 7 commits after it |
| H | `f3bec40a0` | PR head |

`MATCH (a {node_type:'Person'})-[{rel:'FOLLOWS'}]->(b …)-[{rel:'FOLLOWS'}]->(d …) RETURN count(*)` (q8) and the same + `WHERE b.age < 50 AND d.age > 25` (q9), verbatim from `gb_queries.py::GFQL`. Scales 20k / 100k edges (nodes = edges/4, seeded). Cold binding: a fresh `Plottable` is constructed inside every timed iteration; engine frames are materialized once, outside the timed region. **18 slots per arm**, strictly interleaved M,L,H within each slot index; warmup 2 + 17 timed runs per slot; one process per slot; `taskset -c 20,21`, all thread envs = 1. Stats over the 18 slot medians: bootstrap 95% CI (20 000 resamples), Mann–Whitney two-sided. pandas q8 is carried as a **negative control**.

### Value check — L and H are identical; M differs by design

Every arm is internally deterministic (one distinct value per arm across all 18 slots). Across arms:

| cell | M (base) | L (lock) | H (head) |
|---|---|---|---|
| q8 @20k | 80 553 | 80 548 | 80 548 |
| q8 @100k | 399 593 | 399 591 | 399 591 |
| q9 @20k | 31 997 | 31 995 | 31 995 |
| q9 @100k | 161 024 | 161 023 | 161 023 |

(identical for both engines). **L and H agree bit-for-bit on all 8 cells** — so the L→H comparison below is a clean same-answers perf test. M→H differs only because that is the #1905 count(*) trail-uniqueness fix doing its job (the removed paths are the relationship-reusing ones). No arm disagreement is unexplained.

### Results — L→H (post-lock drift, value-identical, the quotable comparison)

| cell | L median ms | H median ms | H/L | ratio CI95 | p (MW) | verdict |
|---|---|---|---|---|---|---|
| polars q8 @20k | 0.71 | 0.74 | ×1.046 | [0.932, 1.117] | 0.65 | **DOES NOT REGRESS** |
| polars q8 @100k | 1.14 | 1.17 | ×1.027 | [0.956, 1.150] | 0.35 | **DOES NOT REGRESS** |
| polars q9 @20k | 3.59 | 3.57 | ×0.995 | [0.960, 1.046] | 0.99 | **DOES NOT REGRESS** |
| polars q9 @100k | 10.32 | 10.47 | ×1.015 | [0.978, 1.052] | 0.21 | **DOES NOT REGRESS** |
| pandas q8 @20k (control) | 3.10 | 3.10 | ×0.998 | [0.932, 1.009] | 0.69 | flat |
| pandas q8 @100k (control) | 7.97 | 8.24 | ×1.034 | [0.994, 1.063] | 0.048 | marginal, CI crosses 1 |
| pandas q9 @20k | 8.17 | 8.50 | **×1.041** | **[1.010, 1.073]** | **0.0031** | **REGRESSES +4.1%** |
| pandas q9 @100k | 19.73 | 20.31 | ×1.030 | [0.974, 1.063] | 0.44 | not established |

### Results — M→H (whole PR)

| cell | M | H | H/M | ratio CI95 | p |
|---|---|---|---|---|---|
| polars q8 @20k | 0.65 | 0.74 | ×1.141 | [0.955, 1.195] | 0.035 |
| polars q8 @100k | 1.09 | 1.17 | ×1.066 | [0.983, 1.188] | 0.064 |
| polars q9 @20k | 3.43 | 3.57 | ×1.041 | [0.952, 1.089] | 0.48 |
| polars q9 @100k | 10.73 | 10.47 | ×0.976 | [0.940, 1.010] | 0.026 |
| pandas q8 @20k | 3.08 | 3.10 | ×1.005 | [0.965, 1.014] | 0.56 |
| pandas q8 @100k | 7.97 | 8.24 | ×1.034 | [0.990, 1.065] | 0.045 |
| pandas q9 @20k | 7.84 | 8.50 | **×1.085** | **[1.018, 1.123]** | **0.0025** |
| pandas q9 @100k | 19.16 | 20.31 | **×1.060** | **[1.014, 1.096]** | **0.0012** |

**No polars cell has a ratio CI that excludes 1.0, on either comparison.** The two sub-ms polars q8 cells show the largest point ratios (×1.14 at 20k) but at 0.65–0.74 ms they are noise-dominated and their CIs comfortably straddle 1 — I am not calling those regressions, and I am also not calling them clean beyond "no effect established". **No cell in this lane is bimodal** (306 pooled samples per arm per cell); pooled-sample ratios track the slot-median ratios throughout, so none of this is the q5/q6-style sampling artifact.

### The pandas q9 regression was predicted from the diff, and the control holds

Reading the post-lock diff first, the only change that adds real work on a reachable path is in `_two_hop_trail_illegal_pairs`: the self-loop correction now runs its O(E) filter **twice** where it previously ran once. That path is reachable from **pandas q9** and *not* from polars q8/q9 (which take a different lane) nor from pandas q8 (dense/degree-product lane, no trail-illegal-pair scan).

That is exactly what the numbers show: q9 pandas moves, q8 pandas does not (×0.998 at 20k), and no polars cell moves. The effect appearing on precisely the predicted cell while the control stays flat is the strongest evidence in this lane. The synthetic graph does contain self-loops (4 at 20k, 2 at 100k), so the branch is genuinely exercised rather than early-returning.

Split of the cost: pandas q9 @20k is +8.5% M→H and +4.1% L→H — so roughly half arrived with the original #1905 fix and **half arrived after the board lock, in the commits that were never re-measured**.

### Recovery

The doubled filter looks straightforwardly recoverable: compute the self-loop mask once and reuse it, rather than applying the O(E) filter on both sides of the correction. That is a local change in `_two_hop_trail_illegal_pairs` with no semantic surface — values are already identical L→H, so a fix has to preserve them, which is easy to pin with the existing count twins. Cost: small. I have **not** implemented or measured it, and I changed no product code.

Also flagged from the diff but **not measured here**: a guard in `_execute_single_hop_grouped_aggregate_fast_path` was hoisted to function entry and made stricter, which can stop the fast path being taken for outputs named like edge endpoints — that would be a large regression rather than a small one, on q3/q4/q6/q7, not on the q8/q9 cells measured above. Those lanes should be run before merge.

### Bottom line

- **polars q8/q9 at 20k and 100k: DOES NOT REGRESS** relative to the `9fb1b77dc` lock on this proxy — consistent with the code reading that no post-lock change touches that path.
- **The board-lock cell itself is INCONCLUSIVE**: `q8 @100k` vs `[13.98, 14.42]` cannot be checked without dgx + the Pokec root. That check still owes a real run.
- **pandas q9 REGRESSES +4.1% post-lock** (CI [1.010, 1.073], p=0.0031) and +8.5% across the whole PR, with a diagnosed and cheap-looking cause.
- q3/q4/q6/q7 fast-path engagement: **unmeasured**, and is the larger of the two open risks.

Under a perf-non-regression bar this is close but not clear: the polars lanes the PR was suspected of moving are clean, but pandas q9 has a real, small, recoverable regression, and two checks (the board q8 cell, the grouped-aggregate guard) are still owed a measured answer. Absolute ms here are from a contended local box on a synthetic proxy — **not quotable**, and deliberately not recorded in pyg-bench as board numbers.

lmeyerov:
## Perf: pandas q9 regression — root-caused, mostly recovered; fast-path audit

**Measurement context.** dgx-spark was unavailable (expired node key, then reserved), so every number is an **INDICATIVE local measurement** pending a protocol-grade run: frozen `git archive` arms at explicit SHAs, 18 slots/arm, strict slot-by-slot interleaving, warmup 2 + 17 reps/slot, CPU-pinned, threads=1, bootstrap 95% CIs + Mann-Whitney over slot medians. Ratios only; nothing goes into pyg-bench. Board queries are the matched gb_queries set on a synthetic graph-benchmark-schema dataset (20k/100k persons, FOLLOWS ~120k/600k edges, 3–4 self-loops, fixed seed).

Arms: `mid` = 2ee64b5b5 (#1904 head, this PR's base), `head` = f3bec40a0, `fix` = +7eaa8e1aa (dedupe the O(E) scan), `fix2` = +ae1cca9b7 (stop hashing domain columns). The pushed head carries both.

### Root cause, in two parts (measured separately)

The predicted double O(E) self-loop filter in `_two_hop_trail_illegal_pairs` was real and is fixed — but measurement showed it was the *smaller* part. The dominant cost was that `_self_loops_in_domains`' pandas arm called `loops[src].isin(domain[node_col])`: pandas `isin` hashes its **argument**, so counting a handful of self-loop pairs built a hash table over each ~10–20k-row domain, three times per query. fix2 probes each unique loop value with one vectorized equality scan instead (hashing the loop side past a pinned probe limit); the cuDF arm keeps device-side `isin` unchanged.

### Ratios (pandas q9 = distinct-domain two-hop count; q8 = equal-domain control)

| cell | pair | ratio | 95% CI | p (MW) | note |
|---|---|---|---|---|---|
| q9 @20k | mid→head | ×1.043 | [1.038, 1.052] | 5.3e-07 | reproduces the reported ×1.041 |
| q9 @20k | head→fix | ×0.994 | [0.987, 1.000] | 0.056 | double-scan removal alone |
| q9 @20k | mid→fix2 | ×1.023 | [1.000, 1.040] | 0.02 | pushed head, residual |
| q9 @100k | mid→head | ×1.018 | [1.007, 1.028] | 0.0012 | smaller than reported ×1.060 at this density |
| q9 @100k | mid→fix2 | ×1.017 | [1.009, 1.028] | 5.6e-06 | pushed head, residual |
| q9 @100k | fix→fix2 | ×0.987 | [0.981, 1.000] | 0.14 | fix2's increment |
| q8 @20k | mid→head | ×0.995 | [0.983, 1.014] | 0.94 | control flat, as reported |
| q8 @20k | mid→fix2 | ×1.001 | [0.991, 1.006] | 0.94 | control stays flat |

Polars, head→fix (my change touches shared helpers; confirming no disturbance — values identical in every cell):

| cell | ratio | 95% CI | p |
|---|---|---|---|
| q8 @20k | ×1.001 | [0.992, 1.008] | 0.65 |
| q8 @100k | ×1.006 | [0.969, 1.026] | 0.76 |
| q9 @20k | ×1.002 | [0.970, 1.043] | 0.74 |
| q9 @100k | ×0.999 | [0.987, 1.018] | 0.81 |

**Verdicts.** pandas q9: **REGRESSES (residual)** — reduced from ×1.043/×1.018 to ×1.023 [1.000, 1.040] @20k and ×1.017 [1.009, 1.028] @100k. The remainder is the irreducible corrected-semantics work: one O(E) `src==dst` scan plus tiny-side membership probes that pre-PR code never did — pre-PR q9 was over-counting (see value evidence). pandas q8: **DOES NOT REGRESS**. polars q8/q9: **DOES NOT REGRESS**, values identical.

### Fast-path disengagement audit (the flagged unmeasured risk)

The flagged claim — a stricter guard hoisted to entry of `_execute_single_hop_grouped_aggregate_fast_path` — **does not exist**: that function is byte-identical between 2ee64b5b5 and f3bec40a0 (and across #1904's diff). Engagement was asserted with call-count spies, not timing: q3/q4 serve `_execute_single_hop_grouped_aggregate_fast_path` on head and fix exactly as on base; q6/q7 (disjoint-rel two-star) still serve `_connected_join_two_star_fast_grouped_count` — the new `_connected_join_trail_arms` correctly declines when arm filters are provably disjoint.

The real disengagement surface is `arms_may_share_an_edge` in `_apply_connected_match_join`: when two arms' edge filters are NOT provably disjoint (e.g. both `FOLLOWS`), both two-star fast paths are bypassed. Measured: ×5.84 [5.64, 6.04] on that shape — but the value also changes (918,038 → 797,992 = minus exactly one self-pairing per FOLLOWS edge), i.e. **the fast path was returning a wrong answer under relationship uniqueness and is disengaged for correctness, not by accident**. That cost is the price of the corrected answer; a trail-aware fast lane for the shared-rel two-star shape is follow-up work.

### Value identity (bar: bit-identical head vs pushed head)

- Canonical values for q1–q9 + the shared-arm probe: **identical head vs fix vs fix2** on pandas at 20k and 100k; every timed slot re-hashed its result (one hash per arm per cell, stable across all 18 slots).
- cuDF (RTX 3080 Ti): q9 = 307,600 on head, fix, and fix2. (cudf q8 fails on this box's missing `libnvrtc` — pre-existing env gap, symmetric across arms.)
- head-vs-mid deltas are this PR's *intended* corrections and agree across pandas/cuDF: q9 307,602 → 307,600 (the 2 self-loops passing both age filters), q8 −3 (the 3 self-loops).
- The exact push tree (fixes + base fold-forward) re-verified value-identical on the full board.

### Gates (run from inside the worktree)

- ruff clean; comment-density, cypher-surface (`lowering_py_max_lines` = measured merged `wc -l`), and type-hygiene guards clean.
- mypy: same 4 pre-existing errors before/after (none in the changed file).
- Full `graphistry/tests/compute`: failure set **byte-identical** across master (bb1fa0d91), this PR's base, head, and the push tree — 102 pre-existing environmental failures, zero from this branch.

dgx protocol run still owed before these ratios are quotable anywhere else.

Pushed head: a9fc97b6e (perf commits 7eaa8e1aa + ae1cca9b7 + base fold-forwards).

Issue #1909, opened 2026-08-15
GFQL Cypher: ungrouped-aggregate identity row lost when a later stage empties the stream (round-007)


lmeyerov:
Round-007 aggregation/ordering probe: 151 queries × both engines, hand-derived oracles (scripts in session scratchpad probe-r7-agg/). **Mostly quiet** — grouping keys, DISTINCT, aggregate identities, ORDER BY null placement, paging, determinism, and OPTIONAL-MATCH aggregation all matched oracles on both engines. One silent-wrong family:

**1. SILENT-WRONG (both engines): an ungrouped aggregate returns 0 rows instead of its identity row whenever a LATER stage empties the stream.** openCypher: an aggregate with no grouping keys always yields exactly one row.
```
MATCH (m) WITH m.name AS n WHERE n = 'Zed' RETURN count(*) AS c
  oracle: 1 row {c: 0}   actual: 0 rows (both engines)
```
Four independent emptying mechanisms, all wrong: post-WITH WHERE, HAVING-style filter on an aggregate, `LIMIT 0`, `SKIP 99`. Also wrong for `collect` (expected `[]`), `count(DISTINCT)`, multi-aggregate returns, post-aggregate expressions (`WITH count(*) AS c RETURN c+1` → expected 1), and edge-rooted matches.

Root cause confirmed by compiler introspection: the single-stage MATCH path attaches `empty_result_row` (lowering.py:7394/7401 via `_empty_aggregate_row`) and the UNWIND row-only path attaches one (:5591 via `_row_only_empty_aggregate_row`), but **the multi-stage MATCH→WITH→RETURN lowering never does**. Controls confirm the localization: `MATCH (m) WHERE m.name='Zed' RETURN count(*)` → correct 1 row, and the connected-join variant is also correct.

Impact: the natural "filter, then count what survived" idiom returns an empty frame instead of a zero — callers doing `df['c'][0]` get an IndexError, or worse read 0-rows as "no data" rather than "the count is zero".

[...]

PR #1910, opened 2026-08-15
fix(gfql): ungrouped aggregates always yield their identity row (#1909)

lmeyerov:
Stacked on #1907. Round-007's one silent-wrong family, plus two bugs the probe had not seen.

**Fix 1 — multi-stage identity row.** `compile_cypher_query`'s `with_stages` branch only set `empty_result_row` for the shortest-path/OPTIONAL case, so any later stage that emptied the stream (post-WITH WHERE, HAVING, `LIMIT 0`, `SKIP 99`) dropped the row openCypher guarantees. New shared `aggregate_identity.py` scans compiled row steps backwards for the last *ungrouped* aggregate producer, seeds identities (count/sum→0, collect→[], else null), then applies the compiled suffix — symbolically for projections/ORDER BY/DISTINCT/SKIP/LIMIT, via a one-row pandas replay for post-aggregate expressions. It **declines** when the suffix holds a filter/unwind/join, because survival then depends on the real aggregate value.

**Fix 2 — paging.** The row-only path bailed whenever the final stage had SKIP/LIMIT. Now the identity row is synthesized and *paged*: `LIMIT 0`→0 rows, `LIMIT n≥1`→the row, `SKIP 0`→the row, `SKIP ≥1`→0 rows. **Probing found the MATCH-rooted equivalents were not actually correct either** (the issue assumed they were): `... RETURN count(*) LIMIT 0` and `SKIP 1` both still emitted the row — fixed. Also found `sum` identity was null on the single-stage path; now 0 everywhere.

**Fix 3 — diagnostics, at compile time on both engines.** Unsupported aggregates (`stDev`, `percentileCont`, …) rejected in the expression-constraint walker; aggregate-in-ORDER-BY rejected via a dedicated check that runs *after* projected-output rewriting, so `ORDER BY count(*)` alongside `count(*) AS cnt` still resolves (an over-rejection this caught in a first attempt). Both now raise `GFQLValidationError [unsupported-cypher-query]` with byte-identical messages across engines, replacing execution-time `invalid-node-reference` / `NotImplementedError`.

**BUG-4 retired:** `test_sum_over_empty_match_is_zero` was a strict xfail from round-002; the `sum→0` identity makes it pass, marker removed with openCypher 9 §3.2 quoted in the docstring. That is the only existing-test change.

Kept deliberately: `sum()`/`avg()` over booleans stays permissive (useful dataframe-native idiom, coherent answer) — now pinned as a documented extension rather than left ambient.

96 new pins (91 passed / 5 strict-xfail). polars NIE families pinned rather than served: all three reduce to one root cause (no native cypher expression engine in the polars row pipeline), each paired with a pandas test asserting the hand oracle so the pins flip loudly when polars lands. Residual pinned: post-aggregate `WHERE` on the aggregate value is undecidable at compile time.

Gates: full-tree failure set byte-identical, TCK 4144P/0F (companion branch pushed), typecheck 329 files, ruff + hygiene guards clean, surface baseline 9772→9808 (≈200 lines were extracted to a new module specifically to keep lowering.py from absorbing them). Compile-time A/B: hot lanes unchanged; only post-aggregate-expression queries pay the one-row replay (compile-time, cached). No two-hop/self-loop perf code touched.


lmeyerov:

Audited head `f375fe3ea` against merge base `f3bec40a0` (base `fix/gfql-1905-count-and-params`, re-fetched and re-merged immediately before this push). Pushed a fix for one introduced silent-wrong plus 50 amplification cells in `test_aggregate_identity_row_semantics.py` and 3 rewritten/added unit pins in `test_aggregate_identity_branches.py`.

### Verdict: BLOCKING FINDING, found and fixed in this push — SAFE TO LAND at `cb11e0731`

**Category 1 (introduced): chained ungrouped aggregates fabricated the final identity.**
`ungrouped_aggregate_identity_row` pivoted at the LAST aggregate seed, assuming its input stream is empty — but an earlier ungrouped aggregate feeds it exactly ONE row (its own identity). Repro at `f375fe3ea` (MATCH-rooted lane; base returned 0 rows, head fabricated a value):
```
MATCH (m) WHERE m.name = 'Zed' WITH count(*) AS c RETURN count(c) AS n
-- head: {'n': 0}   openCypher: {'n': 1}  (the WITH yields one row {c: 0})
```
Likewise `min(c)` → null (true 0), `collect(c)` → `[]` (true `[0]`). The row-only variants (`UNWIND [] AS x WITH count(*) AS c RETURN count(c)`) show the same wrong values but were already wrong at this base via the older `_row_only_empty_aggregate_row` lane (category 2 there; #1901's round declines them on its branch).

**Fix (this push):** the pivot now prefers the EARLIEST seed whose whole suffix is replayable, so the existing replay lane chains the true values through the later `group_by` (`count(c)` → 1 on pandas/polars/cuDF). A filter between the aggregates still pivots at the last seed — a dropped row and an empty upstream both leave the final aggregate an empty input, so `WITH count(*) AS c WHERE c > 99 RETURN count(*)` keeps its correct `{'n': 0}` (pinned).

**Wrong pin (test-quality):** `test_the_last_aggregate_producer_wins` codified the fabrication (asserted `collect` identity `[]` where the chained truth is `[0]`). Rewritten as three pins: chained replay (`count(c)` → 1), chained collect (`[0]`), and the filter-between fallback (last seed, `n=0`). `aggregate_identity.py` stays at its 100.0 coverage floor after the change (re-measured).

**Category 2 (pre-existing / by-design, verified honest):**
- Post-aggregate `WHERE` over an empty stream: the PR's own strict xfail is honest — synthesis genuinely declines (`where_rows` is outside the replay set), and my chain rule deliberately does not fire across it.
- polars decline families: pinned strict xfail with `raises=NotImplementedError` — verified.
- rapids-25.02/26.02 coverage baselines don't list `aggregate_identity.py`; consistent with existing practice (6 other post-baseline modules also absent — those lanes are refreshed on periodic DGX runs), listed for that refresh, not a gate break. `ci-pandas` floor registered by the PR at 100.0; `ci-polars` targets only polars-engine files.

### Mutation audit: 15 mutants, 15 caught (0 equivalence claims needed)
All catches verified by pytest summary lines. sum-identity-null 11F; count_distinct-null 3F; SKIP off-by-one 11F; LIMIT off-by-one 11F; count_table-seed-off 2F; paging-made-transparent 21F; extend-ignored 1F; replay-off 36F; temp-strip-off 6F; ORDER-BY-aggregate-rejection-off 6F; unsupported-aggregates-allowed 10F; general-row-paging-off 4F; compile-wiring-off 29F; pivot-last-revert 14F (the introduced defect, killed by the new chained cells incl. cuDF); graph-lane-identity-revert 3F (incl. the flipped `test_sum_over_empty_match_is_zero`).

### New cells + anti-vacuity
50 new cells: 30 fail at the merge base; 20 are named controls — `sum_of_count` (correct-by-coincidence at base, kept as value-stability control), negative-SKIP/LIMIT validation errors, mid-pipeline paging, DISTINCT/ORDER-BY transparency, `$param` empty UNWIND (pandas serves + polars typed NIE), constant-key grouping stays empty. 14 chained/compound cells fail at the pre-fix head (the pins bite on the introduced defect). PR's own files: 96 semantics cells, 54 fail at base; the branches file fails wholesale at base (new module); no swallow sites (no bare excepts / early-return declines / skips beyond the typed-NIE asserts).

### Engines
pandas + polars throughout; cuDF added on the chained and compound groups (dataframe ops only — this box's cuDF 25.10 works for dataframe ops, cupy is broken `libnvrtc.so.12`, so cupy-backed kernels / `polars-gpu` are out of scope; the ~102 known local failures are that plus optional deps).

### Gates
ruff clean; all three guards rc=0 from inside the worktree (`lowering.py` 9780 <= 9782 cap, untouched; the comment-density guard caught my own 3-line comment — cut to one line, no cap raised); mypy: error set identical to base (4 pre-existing transitive); polars lane: `test_aggregate_identity_row_semantics.py` registered by the PR, `test_polars_lane_completeness` green; `aggregate_identity.py` re-measured at 100% for its `ci-pandas` floor. Full `graphistry/tests/compute` failure-set diff vs base: `identical failure SET (102 = 102, both directions empty; the known cuDF/libnvrtc + optional-dep failures)`.

### Cross-PR notes
- `lowering.py` will conflict with #1901 (it deletes `_row_only_empty_aggregate_row`, which #1901's round patched). Resolve by taking THIS branch's module wiring — it computes what #1901 declines.
- When this branch merges master after #1901 lands, three strict xfails in `test_row_multiplicity_semantics.py` will XPASS (this PR fixes them): `test_empty_aggregate_identity_row_survives_noop_paging`, `test_empty_compound_aggregate_identity_residual`, `test_chained_ungrouped_aggregate_identity_residual`. Flip them to plain asserts in that merge — that's the designed loud signal, not a regression.

Issue #1911, opened 2026-08-15
GFQL Cypher: WITH rebind onto a live alias is unguarded; edge-ident column collision (round-008)


lmeyerov:
Round-008 scoping probe: ~175 query-cases × both engines over 6 fixtures, oracles pre-registered in writing before any code ran, every finding re-verified in a fresh process with different label/type names. Scripts in session scratchpad probe-r8-scope/.

**Scoping is strong where it counts** — scope narrowing (17 cases) errors clearly on every dropped node/rel/path alias with **zero** silent re-binding; OPTIONAL cross-clause visibility (11), parameter/alias collision (8), degenerate scopes (9/9), alias-kind confusion (11, declines only), and an internal-column leak audit were all QUIET. Four silent-wrongs survived.

**1. SILENT-WRONG (pandas, incl. default engine): `WITH X AS Y` where `Y` is already bound returns all-NULL properties.**
```
MATCH (a:P)-[:K]->(b:P) WITH a AS b RETURN b.name
```
oracle `Sa, Sb` · pandas `None, None` · polars correct. The renamed alias takes its ROWS from the source alias but resolves `.property` against the SHADOWED alias's node table — on disjoint node sets that is 100% data loss. Symmetric in both directions, both dtypes, 3 fixtures, survives ORDER BY. Discriminators: whole-entity `RETURN b` is correct on both engines (only property access corrupts), and `WITH a AS fresh` (non-colliding) is a clean typed decline — so the guard exists and is bypassed exactly when the target name is another live alias.

**2. SILENT-WRONG (BOTH engines): simultaneous rebind ignored.** `WITH a AS b, b AS a` returns the unswapped values, byte-identical to the no-WITH control (openCypher WITH projections are simultaneous; oracle is the swap). Same with item order reversed. Also `WITH a, b WITH a AS b RETURN b.name` returns the shadowed original. Note `test_ir_query_graph.py:345` shows the IR already models `WITH a AS b` correctly — the defect is downstream in execution. Cross-kind rebind IS guarded (#1357); same-kind onto a live alias is not.

**3. SILENT-WRONG (pandas) + HARD CRASH (polars): user edge column named `__gfql_edge_ident__`.** Introduced by this campaign's trail work. `gfql_unified.py:671` `_with_edge_identity` tests the hardcoded literal instead of using `generate_safe_column_name`, contrary to the convention `lazy/engine/polars/reserved_columns.py` mandates:
[...]

PR #1912, opened 2026-08-15
fix(gfql): guard WITH rebind onto live aliases; safe edge-identity column (#1911)

lmeyerov:
Stacked on #1910. Round-008's four silent-wrongs.

**Fix 3 (done first — a regression this campaign introduced, and a polars hard crash).** Not one site but **three** stamped or tested a *hardcoded* identity-column literal: `polars/row_pipeline.py:1775` (`with_row_index` unconditional → `duplicate column name` on **every** relationship query for a graph carrying that column), `row/pipeline.py:3891` (pandas *adopted the user's column as the identity*), and the sibling arm-join identity in `gfql_unified.py`. All now resolve through `generate_safe_column_name` against the user's edge columns, with the resolved name threaded through the varlen/multihop/trail-arm call chain; base names centralized in a new `gfql/identifiers.py`. Hardening that helper also fixed a latent pyarrow bug (`.columns` is arrays, `.column_names` is names) which surfaced as a real failure in an existing arrow test.

**Fixes 1+2 — typed decline, deliberately.** `WITH X AS Y` onto a live alias now raises a typed error. The reasoning: lowering has **no** whole-entity alias renaming at all (`WITH a AS fresh` has always declined), so the colliding spelling was never *supported* — it was unguarded; polars' apparently-correct answer was an accident of row-column representation, and it got `WITH a AS b, b AS a` wrong too. Real simultaneous rebind needs alias renaming end-to-end in lowering and both engines. A typed decline beats silent 100% data loss. The guard keys off leading-MATCH pattern aliases (so `WITH a AS x MATCH (x)-->(b)` stays a consumer, not a shadow) and off the query-global alias set, matching how `alias_targets` itself is scoped.

**Fix 4 — partial real fix plus cross-engine consistency.** The alias marker no longer shadows a user column at the property-lookup seam (node alias frames and single-hop edge frames), so pandas returns the user value and matches polars. `MATCH (id:P) RETURN id.id` — where the alias overwrites the node-ID binding — now raises one typed E108 on both engines instead of a raw `ValueError` on pandas versus `True` on polars. Shapes needing the marker itself relocated are pinned as residuals.

Also: the misattributed "Unresolved identifier" now names the visible scope. **Not** done: the `WITH $x AS x` value leak — the compiled plan is clean, so it is a runtime re-substitution issue needing its own investigation, not the cheap item it appeared to be.

64 new pins including the probe's discriminating controls. **Zero existing-test changes** — nothing asserted the old wrong behavior. Two of the agent's own draft oracles were wrong and were corrected against measurement rather than the other way round. Gates: full-tree failure set byte-identical, TCK 4144P/0F, typecheck 329 files, ruff + hygiene clean (fixed by using `Optional[DataFrameT]` per convention, not by raising the cap), surface baseline 9808→9891 (entirely the new validation).


lmeyerov:
**GPU sweep @d11120730: 9328 passed / 53 skipped / 37 xfailed / 0 failed** (probe-gated: cuDF device op + device groupby, polars `GPUEngine(raise_on_fail=True)` — a genuine GPU run, no CPU fallback).

**cuDF lane confirmed live**, which matters because this change threads a resolved edge-identity name through pandas/cuDF-shared varlen/multihop/trail-arm paths: `-k "cudf or gpu" -rs` → **1344 passed, 0 failed**, the single skip being an unrelated missing igraph. Nothing was skipped for absent GPU/cuDF, and the executed set includes exactly the touched paths (lowering cudf/dense/selfloop_multi engine-matrix parity, `test_engine_polars_gpu.py`, index polars-gpu parity, row-pipeline ops).

**Baseline correction:** I had quoted 9160 as the parent tip; that was the sweep taken *before* this stack's 104 coverage tests landed. The measuring agent re-ran the actual parent `2b7a837ff` under the identical wrapper and got **9264**, making the delta here exactly **+64 passed with skips/xfails unchanged (53/37)** — i.e. the ~64 new pins, with no test silently converted to a skip or xfail. 9160 + 104 = 9264 closes the arithmetic.

Incidental, pre-existing, not a failure: polars `is_in` DeprecationWarnings from `hop_eager.py:217` and `pattern_apply.py:240`.

lmeyerov:

Audited head `49c12a8dd` against its base; the base branch had moved to `cb11e0731` (#1910's audit push), so this push first folds that forward (clean merge, no conflicts — the three flagged XPASS-on-merge xfails live on #1901's branch, not this lineage, so nothing to flip here). Then adds 15 amplification cells to `test_alias_scoping_semantics.py`.

### Verdict: SAFE TO LAND (no introduced defects) at `239fe099a`

Probe battery of 12 adversarial rebind/shadow shapes (rebind chains, swaps, cross-kind, scalar shadowing, UNWIND redeclare, path-alias rebind, aliases named like s/d columns, self-renames) diffed head-vs-base: the ONLY behavioral change is `WITH r AS q` over two edge aliases going from a silent split-read (`[None, 2, None]` — rows from `r`, properties from the shadowed `q`) to the new typed decline. Everything else is unchanged. The guard opened no unsupported lanes and broke no working shape.

### Mutation audit: 14 mutants — 10 caught, 2 equivalent with proofs, 1 conservative-improving, 1 uncaught-low-risk
All catches verified by pytest summary lines.
- Caught: guard-off 20F; self-rename-declines 2F; safe-name-counter-off 3F; pyarrow-`column_names`-revert 1F (killed by a new cell — it TypeErrors on arrow's unhashable column arrays); unshadow-off 4F; edge-unshadow-call-off 1F; id-shadow-guard-off 2F; trail-col-ignores-user 1F; polars-fixed-ident 4F; binder-scope-suggestion revert 2F.
- Equivalent with proof: (a) UNWIND carve-out removal — a colliding UNWIND alias is rejected upfront ("UNWIND alias collides", lowering.py:4824/7303) on every path, pinned by a new precedence cell, so the carve-out only ever pops aliases whose queries are already rejected; (b) `clause.kind != "with"` skip — `with_stages` is built exclusively from `expected_kind="with"` projection stages (parser.py:1576/1653), so the branch is a parser-invariant restatement.
- Conservative-improving (noted, not a gap): making the guard ALSO decline cross-kind rebinds passes every pin — i.e. the guard could be widened to give `WITH a AS r` the good compile-time message instead of today's execution-time `GFQLTypeError`. Suggestion, not blocking.
- Uncaught-low-risk: forcing `_with_edge_identity`/trail-arms wiring back to the fixed `EDGE_IDENTITY_COLUMN` name was unobservable in two lane-engaging shapes (the two-star fast path serves them before the trail-drop consults identities), and the risky "reuse user values as identity" branch is dead in the gfql lane at head (the generated name never pre-exists on the frame). Documented here so the next round knows it was hunted.

### New cells + anti-vacuity
15 new cells: 4 fail at the merge base (arrow identity-column resolution; edge-onto-edge rebind decline x2 — the silent split-read; cuDF unshadow+guard parity), 11 are named controls/killers (cross-kind + scalar stay-errors, scalar-shadow WHERE keeps the bag `[30, 30, 35]`, UNWIND collision precedence, edge-alias-named-like-source schema-error residual). PR's own 64 cells: 35 fail at base; the base-passers are the declared controls and residual pins (all residuals honest — verified `{True}` marker leaks and the multihop divergence reproduce exactly as pinned).

### Category-2 findings (pre-existing, listed)
- Cross-kind rebinds (`WITH a AS r` / `WITH r AS b`) surface as execution-time `GFQLTypeError` rather than a compile-time decline — see the conservative-improving mutant above for the free upgrade.
- An edge alias named after the edge SOURCE column has its marker destroy that column (typed `GFQLSchemaError` "column s does not exist" — confusing attribution); pinned as a residual, the node-ID-style typed decline would be the upgrade.
- **Cross-stack flip warning:** `test_with_non_rebind_shapes_are_unaffected[WITH a AS a ...]` pins the DEDUPED bag `[Alice, Bob, Carol]` — the #1899 multiplicity collapse, which #1901 (targeting master) fixes for this exact single-endpoint shape. When this stack merges a master containing #1901, that pin becomes `[Alice, Alice, Bob, Carol]` and must be updated in the merge — expected signal, not a regression.

### Engines
pandas + polars on every group; cuDF added for the unshadow read and guard decline (dataframe ops only — cupy is broken in this env, `libnvrtc.so.12`, so cupy index kernels / polars-gpu are out of scope).

### Gates
ruff clean; all three guards rc=0 from inside the worktree; mypy error set identical to base (4 pre-existing transitive); test file already registered in the polars lane (bin/test-polars.sh:56), `test_polars_lane_completeness` green; no new source files. Full `graphistry/tests/compute`: failure SET at head **identical to base** (102 = 102, both diffs empty) — base reference measured at `b5ec5800e`, which differs from `cb11e0731` only by a one-line comment.

Issue #1913, opened 2026-08-15
GFQL: legitimate .edges() rebind resurrects a STALE adjacency index (silent wrong answers, both engines) + documented recovery fails


lmeyerov:
**Highest-severity finding of the amplification campaign — release-blocking.** Round-009 lifecycle probe, ~40 scenarios + 585 differential checks, both engines, twice-verified. Scripts in session scratchpad probe-r9-lifecycle/.

## 1. SILENT-WRONG on both engines: rebind resurrects a stale index

The declared-index identity+fingerprint guard is *correct*, and `get_valid()` correctly reports the index stale after a rebind. The chain executor then hands it a new lease on life. `compute/chain.py:1220-1223` and its polars twin `gfql/lazy/engine/polars/chain.py:1055-1058` re-point the resident index at the new frame **unconditionally**:
```python
_reg = get_registry(g)
if not _reg.is_empty():
    g = set_registry(g, _reg.rebind_edges(indexed_edges_df))
```
`GfqlIndexRegistry.rebind_edges` (`gfql/index/registry.py:290-328`) validates only the NEW frame's *structural* fingerprint — (row count, bound column names, engine) — and never checks the index was still valid for the frame it is augmenting. Its own docstring states that precondition; **no caller enforces it**. After `g.edges(other_frame)` the identity check has already failed, but the shallow copy re-passes the structural check, so a CSR built over the OLD edges is marked valid for the NEW ones.

```python
E1 = pd.DataFrame({'s':[0,1,2,3,4], 'd':[1,2,3,4,5]})
E2 = pd.DataFrame({'s':[0,3,1,4,2], 'd':[3,0,4,1,5]})   # same row count
g  = gfql_index_edges(graphistry.nodes(N,'id').edges(E1,'s','d'))
g.chain([n({'id':0}), e_forward(), n(), e_forward(), n()])   # warm
gr = g.edges(E2, 's', 'd')                                   # documented rebind
gr.chain([n({'id':0}), e_forward(), n(), e_forward(), n()])
```
oracle (fresh bind of E2) nodes `[0,3]` edges `[(0,3),(3,0)]` — **actual `[2]` / `[]`**. Control without an index is correct. At N=4000 with the index confirmed engaged via `index_trace`: 2-hop `[2]` vs oracle `[0,1582,1707]`.

**Causality proven**: monkeypatching `rebind_edges` to DROP indexes instead of re-pointing makes every wrong answer correct on both engines, nothing else changed.

[...]

PR #1914, opened 2026-08-15
fix(gfql)!: stale adjacency index must not survive an edges rebind (#1913)

lmeyerov:
Stacked on #1912. **The campaign's highest-severity find, and it is not ours — it shipped in 0.58.0.**

**Provenance (asked for explicitly, answered with commits).** `git log -S` over both call sites and `rebind_edges` returns exactly two commits, both 2026-07-19: `9b121e4fb` (index engagement for seeded chains, pandas/cuDF — introduced `rebind_edges` and the pandas call site) and `129d72cb4` (the polars twin). Both are ancestors of master and contained in tag **0.58.0** (2026-07-21). So users of the released version are exposed today: **if you use `create_index`/`gfql_index_*` and rebind edge frames, results after the rebind may be wrong.** That warning is now in the CHANGELOG rather than only in a fix line.

**Fix 1.** `rebind_edges` validated only the NEW frame's structural fingerprint and never that the index was still valid for the frame it augments — its own docstring stated the precondition, no caller enforced it. Now `rebind_edges(new_edges, old_edges)` with `old_edges` **required**, migrating only when `get_valid(...) is idx` and `idx.source_ref is old_edges`. Shape (a) over gating the call sites because it enforces the precondition where it is documented, makes an unmigrated caller fail loudly rather than silently, and keeps exactly one definition of "valid".

One refinement over the issue's sketch: a lineage failure **leaves the entry resident** instead of dropping it. It still truthfully describes its own frame, `get_valid` misses it on the new frame (safe scan), and diagnostics keep reporting resident-and-stale rather than absent — dropping regressed the foreign-engine explain pins, which have a legitimately different lineage. Structural mismatch still drops, as pinned.

**Fix 2 — recovery docs reconciled deliberately.** After Fix 1 the "rebind fresh frames" arm **recovers on both engines** (measured before/after). `gfql_clear_caches()` still does not, and deliberately so: per-graph registries live on user-held `Plottable`s and a process-global function has no handle on them short of a GC walk. So `gfql()`'s docstring now names rebinding / `drop_index` / a fresh binding and says plainly that `gfql_clear_caches()` is not a recovery here, with a reciprocal cross-reference — and **the non-recovery is itself pinned** so the two docstrings cannot drift apart again.

**Fix 3.** The completeness lock now scans the other execution-path modules and notices process-global mutable state regardless of name; the two previously-invisible globals are handled (`_OFFENGINE_BRIDGE_WARNED` made clearable, `_COST_GATE_FRAC_OVERRIDES` exempted with a written reason).

**Honest residual:** a user-level `g.edges(E.assign(w=1))` now falls back to scan — correct, just not indexed. Unavoidable rather than chosen: there is no O(1) discriminator between `E1.copy(deep=True)` and a same-shaped different edge set, so keeping the index there *is* the unsound behavior being fixed, and both suggested fix shapes share this consequence. The optimization's real target — the executor augmenting a frame the index is live for — still engages, pinned by **engagement** assertions (`gfql_explain` + `index_trace`), not merely correctness.

Perf: engagement matrix identical base-vs-fix; seeded 1-hop over 200k nodes / 1.6M edges within noise on both engines (the guard is two O(1) fingerprint compares per chain). Gates: full-tree failure set byte-identical, TCK 4144P/0F, typecheck 329, ruff + guards clean. 8 new index pins verified real by stashing only the product files (8 fail at base); the one existing-test change is a required-parameter migration, not a re-baseline.


lmeyerov:
**GPU sweep @4f0bf4f69: 9347 passed / 0 failed**, parent `d11120730` measured on the same box under the identical wrapper at **9328 / 0 failed** → delta exactly **+19**, the new #1913 pins. (Both commits are clean at 0 failures under `26.02-gfql-polars`; the "93 failed" figure in the commit body is the GPU-less local box, not this image.) cuDF lane genuinely executed: `-k "cudf or gpu" -rs` → 1344 passed, single skip is a missing igraph.

**Targeted verification of the bug itself on GPU — the headline: the GPU lanes carried it too.** Same script at both commits, oracle = a brand-new graph built from the rebound frame with no index:

| check | cuDF @parent | polars-gpu @parent | @fix |
|---|---|---|---|
| diff-values 2-hop | **FAIL** `[2]` vs `[0,65,3966]` | **FAIL** same | PASS / PASS |
| `sort_values` permutation 2-hop | **FAIL** `[2]` vs `[0,1,2]` | **FAIL** same | PASS / PASS |
| var-length (1..3) | PASS | PASS | PASS |
| fewer-rows, `assign(w=1)`, `copy(deep=True)` | PASS | PASS | PASS |

So GPU agreed with CPU both before and after — broken pre-fix, correct post-fix. The fix is not CPU-only in effect even though its pins were. Engagement preserved: `gfql_explain(index_policy="use")` reports `used_index=True` on cuDF and polars-gpu at the fix commit.

Two nuances worth recording: only the fixed-2-hop shape routed through the stale index on GPU — variable-length was already correct pre-fix — and `show_indexes()["valid"]` reported `[False, False, True]` after a bad rebind on **both** commits, i.e. the diagnostic was already honest while queries were being served wrong.

Given this proof, I am extending the #1913 pins to the GPU engine params so the dgx sweep guards these lanes permanently rather than relying on an ad-hoc script.

lmeyerov:

Audited head `d05073934` against merge base `49c12a8dd`; the base branch moved to `239fe099a` (#1912's audit push) and is folded forward here (clean merge). Adds one killer pin to `test_index.py`.

### Verdict: SAFE TO LAND (no introduced defects) at `8cd3650d3`

The lineage fix is sound in every probed direction: user rebind to a different frame / a `sort_values` permutation / a different row count all stay miss-to-scan (correct answers, base was silent-wrong); the executor's shallow-augmentation migration still engages (pinned by engagement tests that the pass-new-as-old mutants prove are load-bearing); no other `rebind_edges` caller exists (grepped — chain.py and polars/chain.py only, both updated), so the breaking signature change is fully absorbed in-repo.

### Mutation audit: 10 mutants, 10 caught
Baseline noise in every run = 38 pre-existing cudf/cupy env failures (`libnvrtc.so.12`); catches counted on the CPU lanes, all with pytest summary lines.
- lineage-check-off 7F (both resurrection families + the stale-refusal unit pin); foreign-index-dropped-instead-of-left 3F; structure-check-off 1F; pandas-chain passes-new-as-old 4F (engagement + typed/untyped index pins + in-place recovery); polars-chain same 3F; registry-self-exemption-off 1F; warn-ledger-unregistered 2F; cost-gate-exemption-off 1F; thin-reason-check-off 1F (`test_registry_fails_loud`).
- lineage-identity-only (drop the `get_valid` conjunct) was initially uncaught — the observable difference is exactly the docstring's promise that a shape-mutated frame's index stays RESIDENT (reported stale) rather than migrating or dropping. Added `test_rebind_edges_leaves_index_resident_after_in_place_shape_mutation`, which kills it (1F verified).

### New cells + anti-vacuity
1 new cell (the S2-killer above); red at the merge base trivially (the two-arg `rebind_edges` does not exist there). PR's own additions: 11 CPU-lane cells fail at the merge base (rebind-to-different-values x2, row-permutation x2, in-place recovery x2, stale-refusal, fingerprint-drop, shallow-augmentation revalidate, warn-ledger clear, cache-lock discovery) — the silent-wrong resurrection pins are real; the remainder are engagement/negative controls, each justified by a mutant above.

### Engines
pandas + polars CPU index lanes fully served (174 passing). The PR's own GPU cells (`[cudf]` variants of the correctness pins) cannot run here — cupy is broken (`libnvrtc.so.12`), so cupy-backed index kernels are out of scope on this box; those cells are the only 3 additions to the full-suite failure set (see below) and are environment-only.

### Gates
ruff clean; all three guards rc=0 from inside the worktree; mypy error set identical to base (4 pre-existing transitive); `test_polars_lane_completeness` green (changed test files already registered); no new source files. Full `graphistry/tests/compute`: 105 failed vs base 102 — the 3 extra are this PR's own NEW `[cudf]` correctness cells failing solely on the broken-cupy env; the base 102 carry over exactly (both set-diffs otherwise empty).

### Notes
- `register_exempt`'s thin-reason rule and the `_REGISTRY` self-exemption are both mutation-verified as load-bearing.
- Category-2 (documented, unchanged): a value-only in-place mutation of the indexed frame still fingerprints as valid (row count + columns + engine cannot see values) — the documented immutability contract; the recovery matrix (fresh frame / `drop_index`) is now pinned end-to-end on both CPU engines.

Issue #1915, opened 2026-08-15
GFQL Cypher: datetime() filters silently match zero rows; temporal + is string concatenation (round-010)


lmeyerov:
Round-010 probe: ~282 queries × both engines (~510 executions) @ 3746c6f30, oracles hand-written to a file BEFORE the sweep ran. Scripts in session scratchpad probe-r10-union-temporal/.

**Substantially QUIET where it counts** — UNION multiplicities, empty-branch handling, per-sub-query ORDER BY/SKIP/LIMIT binding, 3-way unions, mixed MATCH×UNWIND branches, typed rejections for mismatched names; and on the temporal side ORDER BY null placement, **nanosecond precision round-trips intact** (no µs truncation), tz retained, min/max/count, temporal GROUP BY keys with null groups, `duration.between`, ISO normalization. The #1909 identity-row fix also survived UNION correctly. Findings below.

## B-1 — SEVERE, silent: `datetime()` vs a tz-naive datetime column matches ZERO rows (pandas)
```
MATCH (n) WHERE n.ts > datetime('2021-01-01T00:00:00') RETURN n.id
```
Oracle 2 rows; **actual 0**, across **10 spellings** (`>`,`<`,`>=`,`=`, property-map, `count(*)`, AND/OR-chained, edge-endpoint alias). `NOT (...)` returns all 4 non-null rows, proving the predicate evaluated **False**, not null. The intent is clearly supported — `localdatetime(...)`, `date(...)`, and a plain string all return the correct 2 rows on the same column, and a tz-AWARE column with `datetime()` works. Only the canonical Cypher spelling is broken.

Root cause chain: `datetime(...)` lowers to Z-suffixed ISO text; `row/ordering.py:227` `order_detect_temporal_mode` returns None for a real `datetime64` column (it recognizes only TEXT temporals), so the temporal path at `row/pipeline.py:702` is skipped; pandas then raises TypeError comparing datetime64 to str, which `pipeline.py:489-492` routes into `_gfql_safe_mixed_comparison_op`, whose handler does `except TypeError: out_values.append(False)` (`pipeline.py:623-626`). That silent-false is a **general** trap beyond temporals and contradicts this same file's stated intent 200 lines earlier (`pipeline.py:434-437`: incomparable cross-type ordering must be **null**, "never a coercion").

## B-2 — silent: temporal `+` is STRING CONCATENATION (pandas answers; polars declines)
[...]

PR #1919, opened 2026-08-15
fix(gfql): temporal comparison and arithmetic conformance (#1915)

lmeyerov:
Stacked on #1914. Round-010's temporal/UNION findings, B-1 through A-3.

**B-1 (severe) — `datetime()` filters silently matched ZERO rows** on tz-naive datetime columns, across ~10 spellings. Two compounding defects:
(a) `_gfql_safe_mixed_comparison_op` turned a per-element `TypeError` into **`False`**; now `pd.NA`, per openCypher and per this same file's own stated intent 200 lines above. **This is general, not temporal** — it is also why `RETURN '1.0' < 1.0` answered `false`. Independent confirmation: two openCypher TCK scenarios flipped from `success_wrong_rows` to `success_matches_expected`, and the wrong-row debt bucket is now **empty**.
(b) A native-datetime detector feeding the temporal comparison path. Deliberately kept **separate** from `order_detect_temporal_mode` rather than extending it: that function also drives ORDER BY, and routing real datetime columns through the temporal-key sort would have changed null placement for a working feature.

**B-2 — temporal `+` was string concatenation.** Chose real arithmetic over a decline. Temporal literals lower to ISO text before the AST exists, so provenance is gone by evaluator time — but it is intact at **fold** time, so the arithmetic is constant-folded before engine dispatch and **both engines get identical values with no per-engine implementation**. Duration month/day/second groups stay separate, so `date + PT25H` is a no-op, `date + P1D` advances a day, and `date('2020-01-31') + P1M` clamps to 02-29. The non-foldable column form now declines typed exactly like its `-` sibling.

**B-3/B-4 — Z-suffix mismatch.** polars drops a trailing UTC `Z` from both operands when a Z-suffixed literal meets a String column (restoring equality and preserving lexicographic order); pandas routes `IN [temporal]` through the same temporal `=` with Cypher 3-valued OR. Pinned per engine, **not by parity** — B-4 was wrong on both engines, so parity testing could never have caught it.

**A-1/A-2/A-3 — UNION.** Empty branches no longer contribute a type (which alone fixes `['7','8','9']` → `[7,8,9]`); NaN dedups against None; BOOLEAN keys apart from INTEGER, with concat no longer upcasting `True` to `1` before dedup runs. Genuinely unrepresentable mixed-type branches decline typed on polars (a polars column cannot hold both `'7'` and `7`) — pinned in both directions.

83 new pins. **Zero pygraphistry test changes** — nothing asserted the old behavior. Gates: full-tree failure set byte-identical (96 local GPU-less failures both sides), TCK 0 failed with 2 promotions cascaded, typecheck 329, ruff + guards clean.

**Honest residuals:** B-5 untouched (fixing it would contradict the text-path convention that naive text denotes UTC, which B-1b and B-3 both rely on). B-7's tz-aware-vs-naive column case still raises a raw pandas TypeError — it hits a pushdown path *before* the row evaluator. And a **new cuDF divergence to verify on GPU**: the mixed-comparison early-return at `row/pipeline.py:608-611` yields all-`False` for cuDF to avoid a host round-trip, so pandas now returns null where cuDF still returns false — deliberately left rather than blind-editing GPU code that could not be executed locally.


lmeyerov:
**GPU sweep @211c2885b: 9430 passed, 6 failed** — with the parent `c6fc1fcda` measured in the same session at **9347 passed, 6 failed**. Delta **+83 passed, 0 new failures**, matching the 83 new pins exactly. The 6 failures are byte-identical on both commits and are **not from this PR**: they are a test-helper bug in the parent's newly-GPU-promoted #1913 pins (`AttributeError: 'Series' object has no attribute 'tolist'`). Fixed on that branch as `be80784b4` — the helper assumed pandas-vs-polars, but a cuDF/polars-gpu *request* can be served back as a polars Series, and cuDF's Series raises from **both** `to_list` and `tolist` by design, so attribute probing can't discriminate; it now sniffs the module like `dtypes.py` does and uses `to_arrow().to_pylist()`.

cuDF lane genuinely executed: `-k "cudf or gpu" -rs` → 1344 passed, single skip is a missing igraph, **zero GPU-availability skips**.

**The flagged cuDF divergence is CONFIRMED and newly introduced by this PR.** `row/pipeline.py:607-620` early-returns `False` for any cuDF operand while the pandas path now correctly appends `pd.NA`. At the parent both engines agreed (both wrong); this fix moved pandas to the openCypher answer and left cuDF behind:

| shape | oracle | pandas | cuDF |
|---|---|---|---|
| `RETURN '1.0' < 1.0` | null | None ✓ | **False** |
| `RETURN NOT ('1.0' < 1.0)` | null | None ✓ | **True** |
| `WHERE NOT (n.str_col < 1.0)` | `[]` | `[]` ✓ | **`['a','b']`** |
| `WHERE n.str_col < 1.0` (bare) | `[]` | `[]` ✓ | `[]` ✓ masked |

Only genuinely-incomparable ordering comparisons, and only when the boolean is *observed* — via `NOT` or by projecting it. A bare positive `WHERE` masks it because dropping-on-null and dropping-on-false coincide. The new pin file is pandas/polars only, so nothing catches this; a follow-up will either extend the cuDF path or pin the divergence explicitly.

**Good news on the severe finding:** B-1 is fixed on **both** engines — the const-fold and ordering work is engine-agnostic — and at the parent those same cuDF cases raised `NotImplementedError: Binary operations with timezone aware operands is not supported`, so this PR also *repaired* cuDF there.

lmeyerov:

Audited head `42566a644` against merge base `d05073934`; the base branch moved to `8cd3650d3` (#1914's audit push) and is folded forward here (clean merge). The cascade's +336 tests had never had a mutation round — they are the audit target below, plus one category-1 fix.

### Verdict: BLOCKING FINDING, found and fixed in this push — SAFE TO LAND at `6ac9c22d3`

**Category 1 (introduced): duration scaling/rendering hoisted the time group into days, silently changing date arithmetic.** Hand-computed oracle: a Duration keeps its month/day/second groups separate, and a Date consumes the DAY group while dropping the SECONDS group. Repro at `42566a644` (both engines):
```
RETURN date('2020-01-01') + duration('PT18H') * 2   -- head: '2020-01-02'  openCypher: '2020-01-01'
RETURN date('2020-01-01') + duration('PT36H')       -- head: '2020-01-01'  (correct -- the same value!)
```
The head answered the SAME instant two different ways: `_scale_duration` merged days into a single nanosecond total and `format_duration_calendar_components` re-rendered any >=24h time group as days (`PT36H` -> `P1DT12H` shape), so the scaled form advanced dates the plain form (correctly) left alone. Reachable through `duration ± duration` too (`PT30H + PT7H`). The merge base emitted visible garbage (`'2020-01-01PT18HPT18H'` string concat); the new fold lane replaced that with a plausible wrong DATE — the worse failure mode. Two of the PR's own unit pins codified the regrouping (`(0,0,25h) -> "P1DT1H"` and `test_format_normalizes_an_overlong_seconds_group_into_days`); rewritten.
**Fix (this push):** groups scale and render in place — only a fractional DAY result spills into time (`P1D / 2` -> `PT12H`, matching the PR's own `P1D * 0.5` pin); `PT18H * 2` -> `PT36H`; `P3D * 1.5` -> `P4DT12H`. Bonus in the same lane: `duration('P1D') * 0` was declined into Python's `'P1D' * 0` == **empty string** (silent; same at base via the older literal fold) — now served as `PT0S`; division by zero still declines typed (pinned both ways).

### Mutation audit: 17 mutants, 17 caught
All with pytest summary lines. fold-off 81F; date-consumes-seconds-group 11F; month-clamp-off 9F; time-wrap-off 4F; scaler-merges-days 12F; formatter-regroups 12F (both killed by the new B5 cells — they were the live defect); times-zero-declines 3F; polars-union-guard-off 2F; pandas-bool-widen-off 1F; NaN-dedup-key-off 1F; bool-dedup-key-off 1F; empty-branch-drop-off 1F; assume-ns-unit 14F (the pandas-3 us-resolution trap, killed by the resolution files); native-dtype-detection-off 15F; incomparable-to-False 1F (killed at the PR's exact changed site — an earlier variant of this mutant hit a different pre-existing NA site and was superseded); duration-scalar-probe-off 2F; temporal-IN-off 2F.

### New cells + anti-vacuity
24 new cells: 18 B5 group-preservation cells (9 shapes x 2 engines — scaled/summed seconds stay seconds, date consistency both directions, day-fraction spill, times-zero, datetime consumes the seconds group), division-by-zero typed error, UNION ALL cross-branch multiplicity, int/float dedup (`1` vs `1.0` -> 1 row), null-vs-false stays 2 rows, and a cuDF UNION parity cell. At the merge base: the B5 cells fail (concat garbage), the two cypher/ unit files fail wholesale (new-module imports), and the PR's semantics file fails 75/110 (35 base-passing regression fences, all declared). The three rewritten pins fail at the unfixed head (12F each via the scaler/formatter mutants).

### Category 2 (pre-existing / by-design, listed)
- `duration('P1M') / 2` declines typed (Neo4j serves it via the 30.4375-day average month); conservative NIE, pinned.
- cuDF UNION collapses `true` next to `1` (upcast) where fixed pandas keeps both and polars declines typed — pinned as a NAMED divergence cell to flip when the widening is ported (`test_a4_cudf_union_parity_or_named_divergence`).
- polars UNION with genuinely mixed value types is a typed NIE by design (pinned by the PR).

### Engines
pandas + polars on every temporal/union group; cuDF on the union parity cell (dataframe ops only — cupy is broken here, `libnvrtc.so.12`; the 38 `[cudf]` index-lane failures inherited from the base are that env, and polars-gpu is out of scope).

### Gates
ruff clean; all three guards rc=0 from inside the worktree; mypy error set identical to base (4 pre-existing transitive); `test_polars_lane_completeness` green (the PR registered its polars-relevant file; the two cypher/ unit files are engine-free); coverage floors hold — `durations.py` 99% / `folding.py` 99% re-measured vs floors 98.95/98.9. Full `graphistry/tests/compute`: `failure SET identical to base (105 = 105, both diffs empty; 102 known env + the 3 GPU cells #1914 itself added)`.

Issue #1923, opened 2026-08-15
GFQL DAG: ASTCall consumes the previous binding, nondeterministic default output, root-namespace hijack (round-012)


lmeyerov:
Round-012 probe of the `let()`/`ref()`/`call()` DAG surface — never probed before this campaign. ~75 case-shapes × both engines, oracles written to `ORACLES.md` before execution, every finding verified twice on two fixtures. **4 correctness findings (3 silent-wrong), 4 typed-error gaps.** The Cypher `GRAPH x = … / USE x` surface came back exemplary and is the fix precedent for several of these.

## F1 — SILENT-WRONG: a `let()` DAG's default output is NONDETERMINISTIC across processes

`topological_sort` iterates `dependents.get(current, set())` — a **set of strings** (`chain_let.py:157`). Python randomizes string hashing per process, so sibling ordering, and therefore which binding is "last" when `output=` is omitted, changes run to run.

```python
b = {'root': n({}), 'q1': ref('root',[n({'k':'A'})]),
     'q2': ref('root',[n({'k':'B'})]), 'q3': ref('root',[n({'k':'C'})])}
g.gfql(let(b))   # no output=
```
Across 14 process runs (identical on both engines):

| PYTHONHASHSEED | 0 | 1 | 2 | 3 | 4 | 5 | 10 | 11 | 13 | 14 |
|---|---|---|---|---|---|---|---|---|---|---|
| returned | `[]` | `a` | `b` | `a` | `d` | `b` | `w,x` | `y,z` | `u,v` | `w,x` |

Deterministic *within* a process — which is exactly what `test_chain_let.py::test_execution_order_deterministic` checks, so the existing pin structurally cannot see it. Individual binding values are correct; only the choice of returned binding is wrong. Fix: sort the dependents iteration.

## F2 — SILENT-WRONG: an `ASTCall` binding silently consumes the previous binding's output

`execute_node` passes `__original_graph__` to ASTNode/ASTEdge/Chain/nested-ASTLet but passes `accumulated_result` to `ASTCall` (`chain_let.py:352-355` vs `:275-284`, `:364`). Independent siblings leak into each other, and **reordering two unrelated dict keys changes the answer** — violating the contract pinned by `test_let_matchers.py::test_matchers_operate_on_root_graph`.

```python
P = let({'flt': n({'type':'person'}), 'deg': call('get_degrees')})
Q = let({'deg': call('get_degrees'), 'flt': n({'type':'person'})})
```
[...]

PR #1927, opened 2026-08-15
fix(gfql): DAG bindings consume the root graph, deterministic output order (#1923)

lmeyerov:
Independent of the GFQL stack — based on master, merges on its own.

Three silent-wrongs on the `let()`/`ref()`/`call()` surface, all found by round-012 probing, **all cases where both engines agree and both are wrong** (differential testing could not have found them).

**F2 — an `ASTCall` binding consumed the previous binding's output.** Every binding kind except `ASTCall` re-fetched the root back out of the namespace; `ASTCall` used the threaded `accumulated_result` as-is. So **reordering two dict keys changed the answer** — `get_degrees` reported every person as degree 0 when a filter binding preceded it. Fixed structurally rather than by patching the call site: `execute_node`'s `g` parameter now **is** the root, and all four namespace lookups are deleted.

**F1 — the default output was nondeterministic across processes.** `topological_sort` iterated a `Set[str]`; with per-process hash randomization, master produces **three distinct answers** across seeds. Ties now break by declaration order, and cycle reporting sorts its neighbours so error paths are stable too.

**F3 — a user binding named `__original_graph__` hijacked the root.** Fell out of F2: the root is no longer stored in the user namespace, so that name is an ordinary binding and `output='__original_graph__'` is not addressable. The `startswith('__')` filter in the output error is also gone — it was hiding user bindings.

**F4** replaces a bare empty-message `assert` with a typed `NotImplementedError` naming unbound edge endpoints, matching how the chain surface already declines the same shape (it also covers an edgeless graph, which died on that assert). **F5–F8**: nested-`let` dependencies now resolve via `free_variables()`; the single-binding validation shortcut is gone; `GFQLValidationError` propagates unwrapped instead of being downgraded to `RuntimeError`; and structural errors use the `E151`/`E153` codes the Cypher parser already uses instead of bare `ValueError`.

## Two places where fixing one exposed another

**F7 broke lexical read-through, and F5 explained why.** Removing the single-binding shortcut made a nested `let` validate its own bindings, so an outer `ref` read as undefined — the shortcut had been *masking* that, and a two-binding inner let was already broken at master. Fixed by threading enclosing scope through validation and intersecting deps with the local scope before scheduling.

**F2 exposed a test that pinned the bug.** `test_call_operations.py::TestCallInDAG::test_call_in_dag` asserted `len(result._nodes) == 3  # Only 'user' nodes` — i.e. that a call binding inherits the sibling's filter, precisely the behavior #1923 calls silent-wrong and that `test_let_matchers.py::test_matchers_operate_on_root_graph` contradicts. Rewritten as two named contract pins with a hand-computed degree oracle.

## Anti-vacuity: 38 of 48 new cells fail at master

`test_let_binding_contracts.py`. The 10 that pass at master are reported honestly: two hash seeds happen to give the right answer, and the sibling-isolation cells were never wrong. F1's pin runs the DAG in **subprocesses across `PYTHONHASHSEED` 0/1/2/3/7/13** — the existing `test_execution_order_deterministic` cannot see the bug because it never leaves the process.

## Gates
`graphistry/tests/compute` master `102 failed, 9340 passed` → branch `102 failed, 9391 passed`, **failure sets byte-identical**. Let/scoping suites 140 passed; `test_call_operations.py` 27 passed; polars chain/hop/row-pipeline + new file 1588 passed. ruff, mypy (328 files), both guards clean.

Two earlier baseline runs were **discarded as contaminated** (one ran inside the worktree being edited; one had the new test file temporarily present in the master worktree, tripping the polars lane-completeness lock) and re-run in a dedicated clean worktree.

Self-review gate: zero added comments in `graphistry/compute/**`, zero `Any`/`type: ignore`/`hygiene-ok`/`cast(` — it also *removed* a pre-existing `cast(Plottable, ...)` on a touched line in favour of an `is not None` narrowing.

**NIE-tier deliberately left** (reported, not fixed): undeclared schema effects, `collapse` leaking an internal column, docstrings advertising an idiom pinned to return empty, `circle_layout`'s raw `KeyError`.

### DIFF
--- a/graphistry/compute/chain_let.py
+++ b/graphistry/compute/chain_let.py
@@ -5,6 +5,7 @@ from graphistry.Engine import Engine, EngineAbstract, resolve_engine
 from graphistry.Plottable import Plottable
 from graphistry.util import setup_logger
 from .ast import ASTObject, ASTLet, ASTRef, ASTRemoteGraph, ASTNode, ASTEdge, ASTCall
+from .exceptions import ErrorCode, GFQLValidationError
 from .execution_context import ExecutionContext
 from .engine_coercion import ensure_engine_match
 from .gfql.policy import PolicyContext, PolicyException
@@ -35,10 +36,8 @@ def extract_dependencies(ast_obj: Union[ASTObject, 'Chain', 'Plottable']) -> Set
             deps.update(extract_dependencies(op))
     
     elif isinstance(ast_obj, ASTLet):
-        # Nested let is an opaque execution unit — no external dependencies.
-        # Inner bindings are resolved in the inner DAG's own scope.
-        pass
-    
+        deps.update(free_variables(ast_obj))
+
     elif isinstance(ast_obj, Chain):
         # Chain may contain ASTRef operations
         for op in ast_obj.chain:
@@ -51,6 +50,19 @@ def extract_dependencies(ast_obj: Union[ASTObject, 'Chain', 'Plottable']) -> Set
     
     # Other AST types (ASTCall, ASTRemoteGraph) have no dependencies
     return deps
+
+
+def free_variables(dag: ASTLet) -> Set[str]:
+    """Names a let reads from its enclosing scope: inner references minus inner bindings
+
+    :param dag: Let whose bindings are scanned
+    :returns: Set of binding names the let expects its parent scope to provide
+    :rtype: Set[str]
+    """
+    referenced: Set[str] = set()
+    for inner_obj in dag.bindings.values():
+        referenced.update(extract_dependencies(inner_obj))
+    return referenced - set(dag.bindings.keys())
 
 
 def build_dependency_graph(bindings: Dict[str, Union[ASTObject, 'Chain', 'Plottable']]) -> Tuple[Dict[str, Set[str]], Dict[str, Set[str]]]:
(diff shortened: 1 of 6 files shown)


graphistry/compute/chain_let.py
```diff
@@ -114,7 +134,7 @@ def dfs(node: str, path: List[str]) -> Optional[List[str]]:
         color[node] = GRAY
         path.append(node)
         
-        for neighbor in dependencies.get(node, set()):
+        for neighbor in sorted(dependencies.get(node, set())):
```
lmeyerov: Vectorize ?

graphistry/compute/chain_let.py
```diff
@@ -140,73 +160,95 @@ def dfs(node: str, path: List[str]) -> Optional[List[str]]:
...
+
     while queue:
         # Process node with no remaining dependencies
         current = queue.pop(0)
         result.append(current)
-        
+
         # Update dependents
-        for dependent in dependents.get(current, set()):
+        for dependent in sorted(dependents.get(current, set()), key=declaration_order.__getitem__):
```
lmeyerov: Vectorize, incl use native structure to the df engine ?

PR #1928, opened 2026-08-15
perf(gfql): ride the #1888 endpoint-closure gate on work the traversal already does

lmeyerov:
Stacked on #1895 (`fix/gfql-1888-endpoint-closure`). Merge that first.

#1895 makes the endpoint-closure contract true on every surface, but pays for it with a standalone O(E) pass in front of every kernel: on this box a seeded 1-hop costs ~2.6x what it did before the gate. This PR makes the check ride passes the traversal already performs. **The contract is unchanged and nothing is skipped on a heuristic** — the check either moves onto work that already happens, or is skipped only after being *proven* unnecessary.

Product diff is 3 files, +89/−32.

## What the fusion does, per site

**`graphistry/compute/hop.py` (pandas + cuDF) — `closure_rides_traversal`.**
For a *seeded, single, bounded* hop the seeds are intersected with the resolvable ids on the source side, and each hop's already-frontier-narrowed edge frame is filtered on its destination side. The work then follows the frontier instead of the edge table. Unseeded and revisiting traversals (multi-hop, `to_fixed_point`) keep the one-pass edge filter — the unseeded variant was measured on the fused form and *regressed*, so it is deliberately not taken.

**`.../gfql/lazy/engine/polars/hop_eager.py`.**
The two endpoint gates become semi-joins inside the single-hop lazy plan, which already semi-joins pairs against the frontier: one `collect_all`, no extra materialized edge frame. An unseeded frontier needs no source-side join, because `all_nodes` is a subset of the resolvable ids by construction. The single-hop predicate is lifted to `serves_single_bounded_hop` so the gate and the plan cannot disagree about which shape they are serving.

**`.../gfql/lazy/engine/polars/chain.py` — n/e/n fast path.**
The endpoint stack this path already builds to pick output nodes also carries the *proof*: an empty anti-join of those endpoints against the node ids means no edge can need dropping, so the endpoint semi-joins run only on a graph that actually has a dangling endpoint. A side carrying a node filter needs no closure join at all — its ids were drawn *from* the node table (`sides_not_closed_by_a_filter`).

## Measured

dgx GB10 (`dgx-guard/safe_run.sh`, which enforces `--gpus all`), frozen sha-verified `git archive` builds mounted `:ro`, `--network none`, `--binding cold`, `--warmup 2 --runs 17`, **18 interleaved slots per arm** (6-way rotating order), 100k nodes / 2.4M edges.

- **A** = pre-gate `0064072f7` · **B** = this PR's base `a7c9d6fd5` (the gate) · **C** = this PR `0fc3e12db`
- One slot median per cell; CI is a 20k-resample bootstrap on the ratio of medians; p is Mann-Whitney over the 18 slot medians.
- `dirty` = the same shape with ~2% of endpoint ids naming rows the node table does not have (93,060 of 2,400,000 edges — 3.88% — carry at least one dangling endpoint).

### Clean graph, node table bound — where the gate's cost was pure overhead

| engine | shape | A pre | B gate | C fix | **C vs B** | 95% CI | p | C vs A |
|---|---|---:|---:|---:|---:|---|---:|---:|
| polars | seeded hop 1k | 12.85 | 33.82 | 17.02 | **−49.7%** | [−56.9%, −46.9%] | 3.2e-07 | +32.5% |
| polars | chain n/e/n | 11.03 | 26.44 | 15.75 | **−40.4%** | [−47.6%, −36.1%] | 3.2e-07 | +42.8% |
| polars | unseeded hop | 74.40 | 100.54 | 93.84 | **−6.7%** | [−10.7%, −2.3%] | 1.5e-03 | +26.1% |
| pandas | seeded hop 1k | 56.90 | 75.52 | 57.94 | **−23.3%** | [−25.7%, −21.9%] | 3.2e-07 | +1.8% |
| pandas | chain n/e/n | 52.18 | 52.37 | 52.29 | −0.1% | [−2.6%, +3.1%] | 0.74 | +0.2% |
| pandas | unseeded hop | 217.37 | 216.27 | 210.63 | −2.6% | [−5.4%, +1.8%] | 0.048 | −3.1% |

pandas/cuDF is returned to pre-gate cost outright. The polars residual vs pre-gate is the endpoint-membership work itself; removing it needs a persisted closure fact invalidated on rebind (#1913), not a cheaper predicate.

### Dirty graph (~2% phantom ids) — where the gate does real work

| engine | shape | A pre | B gate | C fix | **C vs B** | 95% CI | p |
|---|---|---:|---:|---:|---:|---|---:|
| polars | seeded hop 1k | 12.45 | 37.20 | 16.14 | **−56.6%** | [−61.8%, −53.0%] | 3.2e-07 |
| polars | unseeded hop | 73.27 | 101.77 | 89.51 | **−12.0%** | [−16.3%, −4.4%] | 2.1e-05 |
| pandas | seeded hop 1k | 56.99 | 85.30 | 58.82 | **−31.0%** | [−32.9%, −27.8%] | 3.2e-07 |
| polars | chain n/e/n | 10.53 | 25.79 | 42.44 | **+64.6%** | [+48.5%, +81.0%] | 3.2e-07 |
| pandas | chain n/e/n | 46.25 | 45.88 | 46.40 | +1.1% | [−1.4%, +5.2%] | 0.32 |
| pandas | unseeded hop | 240.49 | 224.12 | 223.03 | −0.5% | [−4.1%, +2.5%] | 0.62 |

On this graph the **A** column is not a perf comparison: pre-gate returns a *different, wrong* answer (2,400,000 edges out of the polars chain where the closed answer is 2,306,940). It is shown only to locate the arms.

### Unbound control (no node table bound — the gate is vacuous)

Six cells, all between **−0.5% and +4.9%**, none significant (C vs B, min p = 0.16). The fusion does not disturb the path it does not apply to.

### Reproducibility

The whole board was run twice on separate 18-slot campaigns (the first against the base at `8a7591b41`, before it absorbed master). **Every one of the 10 non-neutral C-vs-B effects from the pre-merge campaign falls inside this run's 95% CI**, as does every cell of the intermediate run. The larger-looking movement in the *C-vs-A* column (polars chain +35.7% pre-merge → +42.8% here) is in the **A** arm, not this change: B-vs-A moved by a comparable amount on the same cell while C-vs-B stayed put.

## Value identity

- **B and C returned identical `(n_nodes, n_edges)` on all 18 benchmark cells**, clean and dirty, in both campaigns. Every differing cell is A-vs-{B,C} — i.e. the #1888 bug the gate fixes.
- **6,480-case differential fuzz, 0 diffs vs the base.** Graphs *with* dangling endpoints; both engines × bound/unbound × hop/chain × direction × seeds-outside-the-node-table × `return_as_wave_front` × labels × `min_hops` × self-loops/duplicate edges. Frames compared by row count plus a content hash of the id columns.
- `graphistry/tests/compute` failure set is **byte-identical to the base**: 102 failed on both, all pre-existing (predominantly `gfql/index` cuDF cases). Passing count 10208 vs the base's 10190 — the +18 are this PR's new pins.
- `test_endpoint_closure_matrix.py`: 280 passed / 87 skipped vs the base's 262 / 81. The delta is exactly this PR's new `test_hop_seeded_on_an_id_with_no_node_row` (6 seed/direction params × 4 engine lanes; the 6 skips are `polars-gpu`). It pins the *source* side of the rule — a seed id with no node row is itself an unresolvable endpoint, which a destination-only reading of the contract would let through. The cuDF lane really executes (cuDF 25.10).

## Two owner decisions, deliberately not resolved here

**1. The polars hop uses the unconditional semi-join form, not the proof-based form.**
A proof-then-fallback hop (anti-join first, gate only if it finds something) measured ~8ms faster on a *clean* unseeded hop — 81 vs 89ms — but **+86% on a dirty graph**. This PR chose predictability: the unconditional form is never slower than the gate on any graph, at the cost of leaving that 8ms on the table for clean unseeded hops. Reversible in one place if the opposite trade is preferred. (Those two proof-form figures are from the pre-merge campaign at the older base and were not re-measured here; everything else above is post-merge.)

**2. The polars chain n/e/n *is* the proof-based form, and it costs +64.6% vs the gate on dirty graphs.**
That is the price of the proof that makes the clean case 40.4% faster: a failed proof means a second, gated pass. Dirty graphs were exactly the silently-wrong inputs before #1888, so no user has a standing perf expectation there — but it is a real cost on a real shape, and it is the one cell in this PR that moves the wrong way. Symmetry with decision 1 would mean making the chain unconditional too, trading the clean-case win for dirty-graph predictability.

Benchmark code and raw slot data live in pyg-bench; numbers stay out of source comments.

## CI note

`tck-gfql` resolves its companion suite by branch name, falling back to `main`. The base's tck branch `fix/gfql-1888-endpoint-closure` promotes 4 OPTIONAL MATCH scenarios that #1891/#1895 now serve correctly; tck `main` still expects them to xfail, so the first push here failed `test_direct_cypher_xfail_contract[match-where6-2]` purely from the fallback. A `perf/gfql-1888-closure-gate` ref pointing at the base's tck sha (`48ced5951`, unchanged content) was added to tck-gfql so this PR tests the same contract as its base. It becomes redundant once those promotions reach tck `main`.

Refs #1888. Follow-up for the polars residual: #1913.


graphistry/compute/gfql_unified.py
```diff
@@ -1032,6 +1032,31 @@ def _run_logical_pass_pipeline(logical_plan: LogicalPlan, ctx: PlanContext) -> L
     return PassManager(DEFAULT_LOGICAL_PASSES, DEFAULT_TIER2_PASSES).run(logical_plan, ctx).plan
 
 
+if TYPE_CHECKING:
+    from graphistry.compute.gfql.lazy import ExecutionTarget
+
+
+def _policied_auto_serves_via_pandas_until_the_polars_route_emits_hooks(
```
lmeyerov, line 1039: Cudf ?

Issue #1960, opened 2026-08-19
GFQL remote plumbing batch: NaN/inf leaks requests.InvalidJSONError; output= silently dropped for non-Let; shape variant has no params; node_col_subset leaves a dangling _node binding


lmeyerov:
**Category: bare-crash (raw leak) + silent parameter drops + invalid returned Plottable.** All repro-backed on master (`84be35fb3`) with a mocked transport. Low/medium severity batch from the 2026-08 remote-surface amplification probe; filed together because they all live in `chain_remote_generic`'s request/response plumbing.

## 1. NaN / inf in a filter value leaks a raw `requests.exceptions.InvalidJSONError`

```python
g.gfql_remote([n({'x': float('nan')})], api_token='tok', dataset_id='ds-1', format='json')
# requests.exceptions.InvalidJSONError: Out of range float values are not JSON compliant: nan
```

Same for `float('inf')` and for a NaN inside a predicate (`n({'x': gt(float('nan'))})`). The local engine accepts these values and returns a definite answer; the remote path dies with a `requests` exception the caller never imported. Note the contrast: every other non-JSON-able filter value (`np.int64`, `pd.Timestamp`, `datetime`, `pd.NA`, `pd.NaT`, `bytes`, `np.bool_`) is correctly declined with a typed `GFQLTypeError` from `_filter_dict_to_json`'s validator — NaN/inf slip through that check and blow up in `requests`.

Expected: extend the same typed check to non-finite floats (a NaN literal is not valid JSON, so this can never go on the wire), or serialize to a defined null/predicate.

## 2. `output=` is silently ignored for non-Let queries

`chain_remote.py:230-231` only attaches `gfql_output` inside the `if is_let:` branch:

```python
g.gfql_remote([n()], output='foo')
# body -> {"gfql_operations": [...], "gfql_query": {...}, "format": "json", "engine": "pandas"}   # no gfql_output
g.gfql_remote({'type':'Let', 'bindings': {...}}, output='foo')
# body -> {..., "gfql_output": "foo", ...}
```

A user who names an output on a chain (or on a Cypher string that happens to compile to a flat chain — the same source text can compile either way depending on whether it needs bindings) gets no error and no effect. Expected: honor it, or raise a typed error saying `output` requires a Let/DAG query.

## 3. `gfql_remote_shape()` cannot take `params` (or `output`) at all

[...]

Issue #1955, opened 2026-08-19
GFQL remote: Let/DAG queries ship an empty gfql_operations array — an unsupporting server silently returns the whole graph


lmeyerov:
**Category: silent-wrong (category 1), by construction.** Repro-backed for the client half (wire body verified); the server half is INDICATIVE (no live server available to the probe).

## Summary

When `gfql_remote()` sends a Let/DAG query it puts an **empty** `gfql_operations` array next to the real `gfql_query` payload, and documents in a `UserWarning` that an older server will therefore "return the original graph unchanged":

`graphistry/compute/chain_remote.py:216-231`

```python
if is_let:
    warnings.warn(
        "gfql_remote() is sending a Let/DAG query. Servers that do not support "
        "the gfql_query field will receive an empty gfql_operations array and "
        "return the original graph unchanged. ...", UserWarning, stacklevel=2)
    request_body = {"gfql_operations": [], "gfql_query": chain_json, "format": format}
```

An empty operation list is a *valid* GFQL program whose result is the identity. So on an unsupporting server the failure mode is **a plausible, wrong answer** (the whole graph) rather than an error, and the client has no way to tell the two apart: there is no capability negotiation, no server-version check, and no post-hoc assertion on the response. The only signal is a `UserWarning`, which is invisible in notebooks with warnings filtered, in library code, and in any log-scraping pipeline.

A related passthrough: an empty dict is accepted with no validation at all and produces the same identity request.

## Repro (client side, no server)

```python
from unittest.mock import MagicMock, patch
import pandas as pd, graphistry, json
from graphistry.compute.ast import n, ASTLet
from graphistry.compute.chain import Chain

g = graphistry.edges(pd.DataFrame({'s':[0],'d':[1]}),'s','d').nodes(pd.DataFrame({'id':[0,1]}),'id')
g.session.api_token='tok'; g.session.certificate_validation=True
OK = MagicMock(ok=True, json=lambda: {"nodes": [], "edges": []}, headers={"content-type":"application/json"})

for q in [ASTLet({'a': Chain([n({'x':'a'})])}), {}]:
    with patch("graphistry.compute.chain_remote.requests.post") as p, \
         patch.object(type(g),'base_url_server', lambda self:'https://t'):
        p.return_value = OK
        g.gfql_remote(q, api_token='tok', dataset_id='ds-1', format='json')
        print(json.dumps(p.call_args.kwargs['json'])[:160])
```

```
[...]

Issue #1664, opened 2026-06-30
GFQL: conform to openCypher/SQL semantics — track, fix & document violations (3VL nulls, …)


lmeyerov:
## Summary

GFQL is a Cypher-flavored query language, but in several places it diverges from **openCypher / SQL semantics** — most importantly **three-valued (Kleene) NULL logic**. This is an umbrella issue to **track, fix, and document** those violations systematically.

Key principle: **conformance is to the openCypher/SQL spec, not merely to the pandas engine.** GFQL uses pandas as the differential-parity oracle for the other engines (cuDF, polars, polars-gpu), but the pandas engine itself can be non-conformant — so "all engines match pandas" is necessary but **not sufficient**. Where the spec and pandas disagree, the spec wins (we fix pandas), or we document an explicit, justified deviation.

## Reference example (already fixed — the pattern to generalize)

`ne()` / `<>` on NULL: `n({"col": ne(x)})` and cypher `WHERE n.col <> x` over a NULL cell used to **keep** the null row on pandas (`NaN != x` → `True`), diverging from cuDF and polars (both drop it) and from pandas' *own* `WHERE NOT n.col = x` path. Per openCypher/SQL 3VL, `null <> x` is `null` → not a match → row excluded. Fixed in `NE.__call__` (mask nulls); now all four engines agree. (commit `39e278a5`)

A 4-engine null-audit harness (`pandas`/`cudf`/`polars`/`polars-gpu` × `filter_dict` + cypher `WHERE`, with a NULL-bearing fixture) pinned the entire divergence to exactly that one predicate — that audit pattern is how we should sweep the rest.

## Areas to audit / track (each: confirm vs openCypher, fix or document, lock with a test)

- [ ] **Comparison predicates on NULL** — `eq`/`gt`/`lt`/`ge`/`le`/`between`/`is_in` (audit shows these already drop nulls = 3VL-correct; lock with explicit cross-engine tests so they can't regress).
- [ ] **Negation / `NOT` on NULL** — `NOT IN`, `NOT contains`, `NOT startswith/endswith`, double-negation — these are the high-risk cases (negating an `unknown` must stay `unknown` → excluded), since `ne` was the one that was wrong.
- [ ] **String predicates on NULL** — `contains`/`startswith`/`endswith`/`match` over a null cell.
[...]

PR #1952, opened 2026-08-19
fix(gfql): retain min_hops branches that end below max_hops (pandas+polars paired)

lmeyerov:
Fixes #1944

## The bug

`min_hops`/`max_hops` are documented as INCLUSIVE traversal bounds and the prune removes only "dead-end branches that do not reach `min_hops`". But the backward retention walk seeded `current_targets` with the goal nodes at the TOP hop level only, then narrowed targets to the previous level's sources. A goal reached at a level strictly BELOW `max_edge_hop` was therefore never a target when its own level was processed, so its terminating edge — and everything feeding it exclusively — was silently pruned.

Repro at master (`0c3f3a1fa`), graph `0->1, 1->2, 2->3, 3->4, 0->2, 2->0`, seed `{2}`, `min_hops=2, max_hops=3, forward, wavefront`. The tail walk `2->3->4` ends at hop 2, inside `[2, 3]`, so it qualifies:

| surface | master | oracle / this PR |
|---|---|---|
| `hop()` pandas | N `[0,1,2,4]`, E `[(0,1),(1,2),(2,0)]` | N `[0,1,2,3,4]`, E all 6 |
| `hop()` cuDF | N `[0,1,2,4]`, E `[(0,1),(1,2),(2,0)]` | N `[0,1,2,3,4]`, E all 6 |
| `chain()` pandas | N `[0,1,2]`, E `[(0,1),(1,2),(2,0)]` | N `[0,1,2,3,4]`, E all 6 |
| `chain()` polars | N `[0,1,2]`, E `[(0,1),(1,2),(2,0)]` | N `[0,1,2,3,4]`, E all 6 |

Note the direct-hop rows: master kept node `4` while dropping both edges that reach it — an incoherent node/edge frame, not merely a small answer.

## The fix

An edge traversed at a level `>= min_hops` ends a qualifying walk ITSELF, so it is retained outright; only sub-`min_hops` levels still have to feed a retained longer walk.

Landed **paired** because the polars chain mirror (`hop_eager.py`, the layered backward-tree walk) reproduces the identical under-retention. Two product hunks, one on each engine:

- `graphistry/compute/hop.py` — pandas kernel, shared by cuDF; covers forward, reverse and undirected arms.
- `graphistry/compute/gfql/lazy/engine/polars/hop_eager.py` — the polars chain mirror.

## Mutation check (why pairing is the point)

Reverting either engine's half **alone** turns the same **31** cross-engine parity cells red — an identical failure SET both ways:

[...]

PR #1969, opened 2026-08-19
fix(gfql): keep None filter values in to_json (#1954)

### LINKED ISSUE #1954: GFQL wire: filter_dict None values are silently DROPPED by to_json — 'match nothing' becomes 'match everything' over the wire
**Category: silent-wrong (category 1).** Repro-backed, runs on master (`84be35fb3`), pandas, no server needed.

## Summary

`_filter_dict_to_json()` drops every filter entry whose value is `None`. A filter that the local engine evaluates as "matches nothing" therefore serializes to *no filter at all*, i.e. "matches everything". Any surface that goes through `to_json()` — `gfql_remote()` / `chain_remote()` (mandatory), `Chain.to_json()`/`from_json()` round-trips, saved query JSON, the wire protocol consumed by the server — silently returns the whole graph where the in-process call returns the empty graph.

`graphistry/compute/ast.py:123-128`:

```python
def _filter_dict_to_json(filter_dict: Dict[str, Any]) -> Dict[str, Any]:
    return {
        k: v.to_json() if isinstance(v, ASTPredicate) else v
        for k, v in filter_dict.items()
        if v is not None            # <-- entry vanishes, no error, no warning
    }
```

Used by `ASTNode.to_json` (`ast.py:238`) and `ASTEdge.to_json` for `source_node_match` / `edge_match` / `destination_node_match` (`ast.py:578-580`).

## Repro

```python
import pandas as pd, graphistry, json
from graphistry.compute.ast import n, e
from graphistry.compute.chain import Chain

nodes = pd.DataFrame({'id': [0, 1, 2], 'x': [None, 'a', 'b']})
edges = pd.DataFrame({'s': [0, 1], 'd': [1, 2], 'w': [None, 'k']})
g = graphistry.edges(edges, 's', 'd').nodes(nodes, 'id')

for label, q in [('node',        [n({'x': None})]),
                 ('edge_match',  [e(edge_match={'w': None})]),
                 ('src_match',   [e(source_node_match={'x': None})])]:
    j = Chain(q).to_json()
    loc = g.gfql(q)
    rt  = g.gfql(Chain.from_json(j))
    print(label, 'local(n,e)=', (len(loc._nodes), len(loc._edges)),
                 'wire(n,e)=',  (len(rt._nodes),  len(rt._edges)),
                 json.dumps(j))
```

Observed:

```
node       local(n,e)= (0, 0)  wire(n,e)= (3, 0)  {"type": "Chain", "chain": [{"type": "Node", "filter_dict": {}}]}
edge_match local(n,e)= (0, 0)  wire(n,e)= (3, 2)  {... "edge_match": {}}
src_match  local(n,e)= (0, 0)  wire(n,e)= (3, 2)  {... "source_node_match": {}}
```

`g.gfql_remote([n({'x': None})], ...)` puts exactly that `filter_dict: {}` on the wire (verified by intercepting `requests.post`), so the server is asked for **the entire graph** while the same query in-process yields **zero rows**.

## Expected

One of, in preference order:

1. Serialize the `None` as a JSON `null` and give it a defined meaning (whatever `filter_dict={'x': None}` means locally — today it is "match nothing"; per #1664 the 3VL-correct meaning may be `IS NULL`), so local and wire agree; or
2. Reject `None` filter values at construction time with a typed `GFQLTypeError` pointing at `is_null()`/`notnull()`.

Silently widening a predicate to *true* is the one outcome that must not happen.

## Notes

- The local answer itself (`0 rows`) may separately be wrong under openCypher 3VL — see #1664. This issue is about the **drop**, which is wrong under either reading.
- Found by the 2026-08 remote/over-the-wire amplification probe. Related, not duplicate: #1916 (local vs remote schema-strictness divergence).

lmeyerov:
Fixes #1954

## The defect

`_filter_dict_to_json` (`graphistry/compute/ast.py`) dropped every filter entry whose value was `None`:

```python
return {
    k: v.to_json() if isinstance(v, ASTPredicate) else v
    for k, v in filter_dict.items()
    if v is not None          # <-- entry vanishes, no error, no warning
}
```

A filter the local engine evaluates as "matches nothing" therefore serialized to *no filter at all* — "matches everything". `n({'x': None}).to_json()` produced `{'type': 'Node', 'filter_dict': {}}`.

## Root cause: incidental, not intentional

The `if v is not None` guard looks deliberate, so its origin was established before removing it:

- It dates to the **original** serialization commit `ece0924ba` (`feat(gfql): chain serialization`, 2023), and was mechanically lifted into the `_filter_dict_to_json` helper by `14e1ad20e` (#1603, `chore(gfql): shrink predicate ast helpers`) — a pure refactor with no behavior change, no test, and no comment.
- **No caller depends on the drop.** Nothing in the tree builds a filter dict with `None` placeholders it expects to be ignored. Searched every producer: the DSL aliases `n`/`e`/`e_forward`/`e_reverse`/`e_undirected` pass user dicts through verbatim; `hop.py` and `gfql/index/api.py` only ever test whether the *whole* match dict is `None`, never its values.
- **No test asserted the drop.** Zero references to `_filter_dict_to_json` outside `ast.py`; every existing `'filter_dict': {}` assertion is for a genuinely empty `n()`.
- **No wire/validation layer rejects a JSON `null`.** `JSONVal` admits `None`, `is_json_serializable(None)` is `True`, `_validate_filter_schema` skips plain values, and there is no jsonschema anywhere in the tree.

So: **intent (2), incidental JSON-cleanliness habit.** Conversely, the Cypher lowering path *does* produce `None` filter values today and always has — `MATCH (a {x: null})`, `MATCH (a) WHERE a.x = null`, and `MATCH (a {x: $p})` with `params={'p': None}` all lower to `filter_dict={'x': None}` — so the drop was actively destroying real user queries.

## Blast radius is wider than the issue title: this is wrong LOCALLY too

The issue reports the remote path. But `serialize_binding_ops` (`ast.py`) is an **in-process** `to_json`/`from_json` launder used by the Cypher connected-pattern and cartesian-MATCH lowering, so the filter evaporated with no server involved. Measured on master (`42b4f7f7d`), 3 nodes with `x=[None,'a','b']`:

| query | master | correct | this PR |
|---|---|---|---|
| `MATCH (a {x: null}) RETURN a.id` | 0 | 0 | 0 |
| `MATCH (a) WHERE a.x = null RETURN a.id` | 0 | 0 | 0 |
| `MATCH (a {x: null})-[r]->(b) RETURN a.id` | **2** | 0 | 0 |
| `MATCH (a {x: null}), (b) RETURN a.id` | **9** | 0 | 0 |

## The fix

Drop the `if v is not None` clause: one line in one function. `None` now serializes as JSON `null`. `maybe_filter_dict_from_json` already preserved `null` verbatim (it only special-cases `dict` values for predicate revival), so **no `from_json` change is needed** — the asymmetry was entirely on the write side.

Also refreshed one now-stale docstring bullet in `cypher/lowering.py` that cited the drop as the reason `_connected_join_pushable_value` refuses to push a `None`. The guard itself is **unchanged and still correct** — pushing `= null` as a `filter_dict` equality is not the same thing as openCypher's three-valued comparison, and that meaning belongs to the residual (#1664 territory, out of scope here).

## Pins

New `graphistry/tests/compute/test_filter_dict_none_serialization.py`, covering the whole family:

- `to_json` preserves `None` for node `filter_dict`, `edge_match`, `source_node_match`, `destination_node_match`, and for a mixed `{'x': None, 'y': 'a'}` dict.
- Full `json.dumps`/`json.loads`/`from_json` text round trip for each.
- Round-trip *equivalence*: `chain(ops)` and `chain(from_json(to_json(ops)))` return the same node ids and edge pairs on a graph with nulls, plus an explicit "did not widen to the whole graph" assertion.
- The Cypher null-parameter lowering, and the four Cypher null-pattern shapes above.

Assertions are on id lists and row counts rather than `to_dict('records')`, so the py3.13 `None`/`nan` cell-rendering split cannot make them vacuous.

## Verification

- **Anti-vacuity**: the pins were run in a throwaway worktree at the merge-base `42b4f7f7d`. See the numbers in the PR discussion.
- **Mutation check**: restoring the `if v is not None` clause turns the pins red again.
- **Failure-set comparison** against a master baseline over `graphistry/tests/compute` + `graphistry/tests/test_compute*.py`, in both directions.
- `bin/ci_comment_density_guard.py`, `./bin/lint.sh`, `./bin/typecheck.sh` all rc=0, run from inside the worktree.

## Out of scope

- Whether `filter_dict={'x': None}` *should* mean `IS NULL` rather than "never matches" is #1664, not this issue. This PR makes the wire form agree with whatever the local engine already does; it does not change the local answer.
- #1955–#1960 (the other remote-probe findings) are untouched.
- No GPU box available, so no cuDF arm was run. `filter_by_dict`'s `== None` path is engine-shared; pandas and polars were both verified to agree at 0 rows.


lmeyerov:
## Verification receipts

Merge-base: `42b4f7f7d` (ghhttps/master at branch time). All runs local, py3.12.3 / pandas / polars 1.42.0, no GPU box available.

### Anti-vacuity — pins are RED at the merge-base
Fresh throwaway worktree at `42b4f7f7d`, pin file copied in, product code untouched:

```
21 failed, 2 passed in 3.07s   (rc=1)
```

The 2 that pass at master are `MATCH (a {x: null}) RETURN a.id` and `MATCH (a) WHERE a.x = null RETURN a.id` — those two shapes never route through binding-op serialization, so they were already correct. They are kept so the family is pinned as a whole rather than only at the broken shapes.

With the fix: `23 passed` (rc=0).

### Mutation check — per site
There is exactly **one** fix site: the `if v is not None` clause in `_filter_dict_to_json` (`graphistry/compute/ast.py`). Applying **only** that one-line change on top of the untouched merge-base, with nothing else from this PR (no CHANGELOG, no lowering.py docstring):

```
23 passed in 2.64s   (rc=0)
```

So that single site accounts for **21 / 21** of the red. The other two files in the diff are non-behavioral, confirmed by the fact that the pins are fully green without them.

### Failure-set comparison — both directions
Suite: `graphistry/tests/compute` + `graphistry/tests/test_compute.py` + `test_compute_chain.py` + `test_compute_hops.py` + `test_compute_filter_by_dict.py`, run at the merge-base and at this branch.

| | failed | passed | skipped | xfailed |
|---|---|---|---|---|
| merge-base `42b4f7f7d` | 107 | 12343 | 980 | 46 |
| this branch | 107 | 12362 | 980 | 46 |

`comm` on the sorted FAILED node-id sets:

```
=== ONLY IN FIX (new failures) ===
(empty)
=== ONLY IN BASE (newly fixed) ===
(empty)
```

Identical sets, not just identical counts. The +19 passed is this PR's pin file as collected by that run. The 107 pre-existing failures are environment-bound and untouched by this PR: 86 are cudf/GPU/cugraph, the rest are polars-version drift in the conformance matrix (`ceil`/`floor`/`round`) and missing umap/dask optional deps.

### Gates — all run from **inside** the worktree
```
python bin/ci_comment_density_guard.py   -> rc=0  (1202 grandfathered, no growth)
./bin/lint.sh                            -> rc=0
./bin/typecheck.sh                       -> rc=0  (mypy 2.3.1, 331 source files)
```

### Polars lane
The pin file is not registered in `bin/test-polars.sh` and does not need to be: it never mentions polars, so `test_polars_lane_completeness.py` does not apply to it. The contract it pins is serialization, which is engine-independent. Both engines were nonetheless checked by hand and agree — pandas and polars each return 0 rows for every `None`-filter shape, before and after the round trip.

### cuDF
No GPU box was available for this lane, so no cuDF arm was run. No cuDF-only code arm is introduced by this diff — the changed line is engine-independent dict serialization — so there is nothing to mark `# pragma: no cover`. `filter_by_dict`'s `== None` execution path is shared across engines and unchanged.

### Not changed, on purpose
- **`_connected_join_pushable_value` still refuses to push a `None`** (`cypher/lowering.py`). Its stated reason was the drop, which is now gone, but the guard remains correct for a different reason: a pushed `= null` becomes a `filter_dict` equality, which is not openCypher's three-valued comparison. Only the docstring bullet was corrected. Un-blocking that pushdown is an optimization for #1664, not this issue.
- **`_predicate_value` still lowers `= null` to a raw `None`** rather than an explicit `never_match()`. That would be a more self-describing encoding and more robust to a server that reads JSON `null` differently, but it changes the AST shape, not the serialization defect, and belongs with the 3VL work.
- **#1955–#1960 untouched.** This diff does not incidentally repair any of them; the two Cypher wrong answers it does fix (connected-pattern and cartesian `MATCH` with a null property) are unfiled consequences of #1954 itself, found while root-causing it, not separate issues.

lmeyerov:
## Re-verified after rebase

Master moved while this was being prepared (#1952 landed), which made the PR CONFLICTING on the CHANGELOG `### Fixed` header — and a CONFLICTING PR fires **zero** CI, so it was rebased onto `0aaa86222` and every receipt above was re-measured against that merge-base. The code diff is byte-identical; only the CHANGELOG entry moved (both entries kept, #1954 above #1944).

**Anti-vacuity at `0aaa86222`** — pins copied into a clean worktree, product code untouched:
```
21 failed, 2 passed in 2.67s   (rc=1)
```
Same 21/23 as before the rebase. With the fix: `23 passed` (rc=0).

**Failure-set comparison, re-run at `0aaa86222`:**

| | failed | passed | skipped | xfailed |
|---|---|---|---|---|
| merge-base `0aaa86222` | 107 | 12353 | 980 | 46 |
| this branch `c51a5f266` | 107 | 12376 | 980 | 46 |

```
=== ONLY IN FIX (new failures) ===
(empty)
=== ONLY IN BASE (newly fixed) ===
(empty)
```

Identical sets in both directions. +23 passed is exactly this PR's pin file.

**Gates re-run from inside the rebased worktree**: comment-density guard, `./bin/lint.sh`, `./bin/typecheck.sh` — all rc=0.

**PR state**: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (awaiting review approval, per repo convention — no auto-merge). Check runs are live and real, not skipped: 23 pass / 6 pending / 1 skipping / 0 fail at the time of writing, so the lint lane is green and the test lanes are genuinely executing.

Issue #1957, opened 2026-08-19
GFQL remote: engine='auto' is resolved from the CLIENT's installed libs (frameless graph sends engine=cudf); polars graphs POST then die on an untyped ValueError


lmeyerov:
**Category: surface-divergence / engine dishonesty + bare-crash-after-side-effect.** Repro-backed on master (`84be35fb3`) with a mocked transport.

Related but distinct from #1824 (`polars-gpu` silently serving CPU *locally*): this is the remote path's analogous honesty problem.

## 1. `engine='auto'` on the remote path resolves against the *client's* installed libraries

`chain_remote_generic` resolves the engine locally and ships the resolved string to the server (`chain_remote.py:144-149, 245`; identical in `python_remote.py:129-142`). `resolve_engine('auto', g)` falls through to "whatever this machine can import" when there is no local frame to inspect (`Engine.py:120-123`):

```python
has_cudf_dependancy_, _, _ = lazy_cudf_import()
if has_cudf_dependancy_:
    return Engine.CUDF
return Engine.PANDAS
```

The documented remote flow is precisely the frameless one — bind a `dataset_id` and query server-side data with no local copy. Repro:

```python
g = graphistry.bind(); g.session.api_token='tok'; g.session.certificate_validation=True
# resolve_engine('auto', g) -> Engine.CUDF   (on a client where cudf is importable)
# request body -> {"...": ..., "engine": "cudf"}
```

Observed on this probe box (cudf importable, no GPU): `engine="cudf"` is sent. On a client without cudf the *same query against the same server dataset* sends `engine="pandas"`.

So the server's execution engine — with all the engine-dependent semantics the 2026-08 campaign has been cataloguing, plus GPU cost/availability — is chosen by an accident of the *client's* pip environment, silently, with the docstring claiming only that auto "auto-detects based on DataFrame type". There is no DataFrame to detect.

Expected: for the remote path, `engine='auto'` should mean "let the server choose" (omit the field, or send `auto`), or be resolved from the *server's* advertised capability. A client-side cudf install must not select a GPU run on the server, and its absence must not veto one.

## 2. polars-backed graph: request is POSTed, then the client throws the answer away with an untyped `ValueError`

[...]

PR #1973, opened 2026-08-19
fix(gfql): typed remote error surface — no raw requests/zipfile/KeyError/IndexError leaks (#1956, #1960)

### LINKED ISSUE #1956: GFQL remote: client error surfacing leaks raw requests exceptions, destroys server error messages, and bare-crashes on error/odd responses (6 repros)
**Category: bare-crash / raw-exception-leak / surface-divergence.** All six repro-backed on master (`84be35fb3`) with a mocked `requests` transport; no live server needed.

Precedent for "raw leak is a filed defect": #1880, #1915 B-7.

## Repro harness

```python
import json, io, zipfile
from unittest.mock import patch
import pandas as pd, graphistry, requests
from graphistry.compute.ast import n
from graphistry.compute.chain_remote import chain_remote_generic
from graphistry.compute.python_remote import python_remote_generic

def resp(status, content, ctype):
    r = requests.models.Response(); r.status_code = status
    r._content = content; r.headers['content-type'] = ctype; r.url = 'https://t/x'
    return r
# ... a Plottable `g` with pandas frames, api_token + dataset_id passed explicitly ...
```

## 1. Raw `requests.exceptions.JSONDecodeError` leaks (both modules)

Server (or, far more commonly, a proxy/LB/gateway) returns an error status with `content-type: application/json` but a non-JSON body:

```python
r = resp(500, b'<html>gateway error</html>', 'application/json')
```

- `chain_remote` → `requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
- `python_remote` → same

Cause: `json.JSONDecodeError` **subclasses `ValueError`**, so the `except (ValueError,) as ve: raise ve` arm (`chain_remote.py:275-277`) and the `except ValueError: raise` arm (`python_remote.py:167-169`) — both written to re-raise the module's own `ValueError` — instead swallow the fall-through and re-raise the decoder's exception. The `except Exception: response.raise_for_status()` fallback is unreachable for this case.

## 2. `python_remote` leaks a raw `requests.exceptions.HTTPError`; `chain_remote` does not (divergence)

```python
r = resp(502, b'<html>bad gateway</html>', 'text/html')
```

- `python_remote` → `requests.exceptions.HTTPError: 502 Server Error: None for url: ...` (raw, body discarded)
- `chain_remote` → `ValueError: GFQL remote operation failed: <html>bad gateway</html> (HTTP 502)`

`python_remote.py:161-173` has no non-JSON branch at all, so it always ends at `response.raise_for_status()`.

## 3. Both zip error handlers destroy the server's error message

Server answers 200 with a JSON error body while the client expects a zip:

```python
r = resp(200, json.dumps({"error": "GFQL validation failed: unknown column 'foo'"}).encode(), 'application/json')
chain_remote_generic(g, [n()], format='parquet', output_type='all')
```

Observed: `ValueError: GFQL remote operation failed - server response is not a valid zip file. This usually indicates a server validation error. Response status: 200`
Expected: the message the handler had already built — `... unknown column 'foo'`.

Cause: the informative `ValueError` is raised *inside* a `try` whose own `except Exception:` catches it and replaces it (`chain_remote.py:374-385`, `python_remote.py:219-234`). The good message is only reachable as `__context__` in the traceback. `python_remote` loses `"python task raised ZeroDivisionError"` the same way.

## 4. `KeyError: 'edges'` on a 200 JSON error body (`format='json'`)

```python
r = resp(200, json.dumps({"error": "boom"}).encode(), 'application/json')
chain_remote_generic(g, [n()], format='json', output_type='all')
```
Observed: `KeyError: 'edges'` (`chain_remote.py:404` indexes `o['edges']`/`o['nodes']` unguarded). The JSON path has no equivalent of the zip path's error sniffing.

## 5. Missing zip member → `IndexError: list index out of range`

```python
buf = io.BytesIO()
with zipfile.ZipFile(buf,'w') as z: z.writestr('edges.parquet', b'')
```
Observed: `IndexError: list index out of range` from `chain_remote.py:314-315` / `python_remote.py:200-201`.

## 6. Zip member selection is substring/order dependent — silently mis-assigns frames

`nodes_file = [f for f in zip_ref.namelist() if "nodes" in f][0]` also matches any member whose name merely *contains* `nodes`. A zip containing `nodes_and_edges.parquet` + `nodes.parquet` yields:

```
nodes: [{'s': 0, 'd': 1}]   # <-- the edge table, silently, bound as nodes
edges: [{'s': 0, 'd': 1}]
```

No error. Any future extra member (`nodes_meta.json`, `edges_stats.parquet`, …) either swaps frames or feeds a non-parquet blob to `read_parquet`.

## Expected

- Error surfacing should be a single shared helper used by both modules, raising a typed GFQL error (e.g. `GFQLRemoteError`/`GFQLValidationError`) that carries HTTP status + server message; no `requests` exception class should reach a user of the public API.
- Catch `json.JSONDecodeError` (or use a sentinel/`else` clause) before the generic `ValueError` arm, so the fallback is reachable.
- Do not wrap the informative raise in a `try` that catches it — build the message, then raise outside.
- Guard `o['nodes']`/`o['edges']` and sniff an `error` key on 200 responses in the JSON path.
- Select zip members by exact name (or a manifest), not substring; raise a typed error naming the missing member.

Found by the 2026-08 remote/over-the-wire amplification probe (first pass over this surface). Cross-ref #1916.

### LINKED ISSUE #1960: GFQL remote plumbing batch: NaN/inf leaks requests.InvalidJSONError; output= silently dropped for non-Let; shape variant has no params; node_col_subset leaves a dangling _node binding

lmeyerov:
## What reproduced

All ten sub-items across #1956 and #1960 reproduced at master `401f54f2f` with a mocked transport, and **all ten are fixed**. The harness patches `requests.sessions.Session.send`, not `requests.post`, so requests' real body preparation (`json.dumps(..., allow_nan=False)`) and real `Response` decoding both still run — patching `requests.post` hides the #1960-1 defect entirely.

| # | claim | reproduced at master | after |
|---|---|---|---|
| 1956-1 | `requests.exceptions.JSONDecodeError` leaks from **both** modules | yes, both | typed `GFQLRemoteError`, server body preserved |
| 1956-2 | `python_remote` leaks raw `HTTPError`, `chain_remote` wraps it | yes (the divergence is exactly as filed) | both typed, both keep the body |
| 1956-3 | zip handlers' `except Exception` eats their own informative message | yes, both | server's real text survives |
| 1956-4 | `KeyError: 'edges'` on a 200 JSON error document | yes — and **also in `python_remote`**, not just `chain_remote` | typed, names the server message |
| 1956-5 | `IndexError` on a zip missing a member | yes, both | typed, names the missing member |
| 1956-6 | substring member selection binds the EDGE table as nodes | yes, both, **silently** | exact-stem selection; ambiguity is an error |
| 1960-1 | NaN/inf leak `requests.InvalidJSONError` | yes (nan, inf, and nan inside a predicate) | typed `GFQLTypeError`, before the POST |
| 1960-2 | `output=` silently dropped for non-Let | yes (`gfql_output` absent from the body) | typed decline |
| 1960-3 | `gfql_remote_shape` has no `params` | yes (`TypeError: unexpected keyword argument`) | `params` and `output` threaded through |
| 1960-4 | `node_col_subset` strands `_node` | yes — and **`edge_col_subset` strands `_destination` the same way** | typed `GFQLSchemaError` |

Two findings beyond the filed text: #1956-4 affects `python_remote` too, and #1960-4 has an edge-side twin. Both are covered.

## Fix decisions

A single new `graphistry/compute/remote_response.py` backs both modules, so `chain_remote` and `python_remote` can no longer drift apart the way #1956-2 documents.

**Where I chose a typed decline over new behavior:**

- **`output=` on a flat chain (#1960-2) — declined, not honored.** `gfql_output` selects which *binding* of a Let/DAG to return. A flat chain has no bindings, so there is nothing for the server to select; forwarding it would just move the failure to the server (or be ignored there, reproducing the original silent drop). The issue lists "raise a typed error saying `output` requires a Let/DAG query" as an acceptable outcome, and that is what ships: `GFQLSyntaxError` E109, raised before the request. Let/DAG queries are untouched and still send `gfql_output`.
- **`node_col_subset`/`edge_col_subset` stranding a binding (#1960-4) — declined, not silently rebound.** Dropping the stale binding would hand back a graph that is quietly missing its id column; naming the column is the honest answer. Scoped deliberately: the check fires **only when the caller asked for a column subset**, and it runs on the *final* result, after server `metadata.bindings` hydration — so a server that authoritatively rebinds the result (the case `test_metadata_overrides_existing_bindings` pins) is unaffected. That scoping is load-bearing: an unscoped version broke that existing test.
- **Non-finite floats (#1960-1) — gated on the wire body, not in `ASTNode.validate()`.** Putting it next to the existing `is_json_serializable` check would (a) change local-execution behavior, which is out of scope and which the issue explicitly contrasts as working, and (b) **miss `n({'x': gt(float('nan'))})` entirely**, since predicate values short-circuit that check. Scanning the assembled request body catches all three filed shapes and raises the same `GFQLTypeError` class the other non-JSONable values already get, naming the JSON path (`gfql_operations[0].filter_dict.x.val`).

**Where new behavior was the right answer:**

- `select_zip_member` matches on the member's *stem* (`nodes.parquet` → `nodes`), falls back to a looser match only when it is **unambiguous**, and raises typed when there is no match or several. The fallback keeps compatibility with a server naming members differently; the ambiguity error is what stops #1956-6 from ever silently picking `[0]` again.
- `GFQLRemoteError` subclasses `GFQLValidationError` **and `ValueError`**, so code written against the previous untyped `ValueError` keeps working. New codes E401 `remote-request-failed` / E402 `remote-response-malformed`; the HTTP status is both in the message and on `.context['status_code']`.
- `python_remote(output_type='json')` uses a decode that does **not** sniff `error` keys: a task's own return value is the result there, and a task may legitimately return an error-shaped dict. Only the graph/table paths refuse error documents.

## Receipts (merge-base `401f54f2f`)

**Red at merge-base.** Run as-is at master the pin file is 1 collection error, not meaningful red — `GFQLRemoteError` does not exist there. So it was also run at master with a one-line shim aliasing `GFQLRemoteError = ValueError` (the loosest class master could raise) and **every other assertion byte-identical**: **24 failed, 7 passed**. The 7 passers are the controls that must stay green in both directions — `test_non_json_error_status[gfql]` (chain_remote already wrapped it; #1956-2 is python-only), `test_finite_filter_value_still_goes_on_the_wire`, `test_output_on_let_still_reaches_the_wire`, `test_col_subset_keeping_the_bound_columns_is_accepted`, `test_no_col_subset_leaves_server_supplied_bindings_alone`, and both `test_well_formed_zip_still_round_trips`.

**Per-site mutation** (each site reverted individually on the branch, pins re-run; 31 green when whole):

| site | file | pins red |
|---|---|---|
| A non-2xx handler | `chain_remote.py` | 2 |
| B non-2xx handler | `python_remote.py` | 2 |
| C `BadZipFile` message preservation | `chain_remote.py` | 1 |
| D `BadZipFile` message preservation | `python_remote.py` | 1 |
| E zip member selection | `chain_remote.py` | 3 |
| F zip member selection | `python_remote.py` | 3 |
| G json result guards | `chain_remote.py` | 2 |
| H json result guards | `python_remote.py` | 2 |
| I non-finite wire gate | `chain_remote.py` | 4 |
| J output-requires-Let gate | `chain_remote.py` | 1 |
| K shape `params` threading | `chain_remote.py` | 1 |
| L binding check, node branch | `remote_response.py` | 2 |
| L2 binding check, edge branch | `remote_response.py` | 1 |
| L3 binding check, whole | `remote_response.py` | 3 |

No site mutates to zero, so no pin is vacuous. Restoring every site returns 31/31.

**Failure-set comparison.** Full `graphistry/tests`, `TEST_CUDF=1`, cudf 25.10 + cupy 13.6 on the local RTX 3080 Ti with `LD_LIBRARY_PATH` pointed at the pip `nvidia/*/lib` dirs and GPU JIT verified by running an actual kernel (not just an allocation). See the receipts comment below for the two FAILED sets.

`python bin/ci_comment_density_guard.py`, `./bin/lint.sh`, `./bin/typecheck.sh` — all rc=0 from inside the worktree, each captured explicitly rather than through a pipe.

## Deliberately not changed

- **#1955** (Let ships an empty `gfql_operations`) and **#1957** (`engine='auto'` resolved from the client env) — out of scope by instruction; the adjacent code is untouched and both behaviors are byte-identical.
- **#1971's `df_import_args` / csv decline and #1972's `normalize_task_code()`** — both preserved; `test_remote_csv_fidelity.py` and `test_python_remote_code_normalization.py` stay green.
- `is_json_serializable()` still accepts NaN. Tightening it would change local execution and `is_in` semantics, which neither issue asks for; the wire gate is where the defect lives.
- No server-side change (an `error` key on a 200 is still how the server signals validation failures on the zip path).

Fixes #1956
Fixes #1960


lmeyerov:
## Failure-set receipts (merge-base `401f54f2f`)

Full `graphistry/tests`, `TEST_CUDF=1`, deterministic order (`-p no:randomly`), same box, cudf 25.10 + cupy 13.6 on an RTX 3080 Ti with `LD_LIBRARY_PATH` pointed at the pip `nvidia/*/lib` dirs (GPU JIT verified by running a real kernel, not just an allocation).

| | failed | passed |
|---|---|---|
| merge-base `401f54f2f` | 18 | 13755 |
| this branch | 18 | 13786 |

**+31 passed = exactly the 31 new pins. The two FAILED sets are identical in both directions** — zero new failures, zero incidental fixes. The 18 are pre-existing local-env failures (dask, sentence-transformers / text-search, a few GPU-specific), untouched by this diff.

Measured twice on the branch: once before and once after the follow-up that put the HTTP status into the error message text (and pinned it). Both runs give the same 18/13786 and the same identical set — the second run is the one at the pushed HEAD.

**CI at the pushed head:** 73 success, 3 skipped, 0 failed. The three skips are `cypher-frontend-ci-gates`, `cypher-frontend-strict-typing (py3.12)`, and `cypher-frontend-surface-guard` — path-filtered lanes, not lint-induced skips; every lint/typing lane in the 73 is green.

lmeyerov:
## Follow-up: #1956-6 was only half-fixed — now complete (`cf0137ef5`)

Confirmed the report at the previous head. Exact-stem precedence fixed the decoy-alongside-correct-member case, but the loose fallback reintroduced the defect whenever the exact member was absent:

```
select_zip_member(['nodes_and_edges.parquet', 'edges.parquet'], 'nodes') -> 'nodes_and_edges.parquet'
select_zip_member(['nodes_and_edges.parquet'],                  'nodes') -> 'nodes_and_edges.parquet'
select_zip_member(['nodes_and_edges.parquet'],                  'edges') -> 'nodes_and_edges.parquet'
```

One compound member bound as **both** tables at once. That is the filed silent-wrong, still live.

### Fix

A loose match is accepted only when the name mentions one kind and **not the other**. Compat intent is preserved and the compound case becomes a typed decline that names the ambiguity rather than guessing which table the file holds.

Verified at the new head, both directions:

| names | `'nodes'` | `'edges'` |
|---|---|---|
| `['nodes_and_edges.parquet']` | decline | decline |
| `['nodes_and_edges.parquet','edges.parquet']` | decline | `edges.parquet` |
| `['nodes_and_edges.parquet','nodes.parquet','edges.parquet']` | `nodes.parquet` | `edges.parquet` |
| `['graph_nodes.parquet','graph_edges.parquet']` | `graph_nodes.parquet` | `graph_edges.parquet` |
| `['out/result_nodes.csv','out/result_edges.csv']` | `out/result_nodes.csv` | — |

The docstring claimed *"never by substring"* while the fallback did exactly that. It now states the real policy: exact stem first, then an unambiguous loose match that cannot match the other kind, else decline.

### Receipts

**New pins (8 added, 31 -> 39), red before the change at the previous PR head:** `test_compound_member_is_never_bound_to_either_table`, `test_compound_member_does_not_stand_in_for_a_missing_member`, and `test_compound_only_zip_declines_instead_of_binding_one_table_twice` (gfql + pyrem) — **4 failed / 35 passed**. The three compat/control pins (`test_prefixed_member_names_still_resolve`, `test_exact_member_still_wins_over_a_compound_decoy`, `test_prefixed_member_zip_still_round_trips`) were already green at that head and stay green — that is the point of them.

**Red at merge-base** (`401f54f2f`, master has not moved), same `GFQLRemoteError = ValueError` shim, all other assertions byte-identical: **30 failed / 9 passed** (was 24/7). Honest caveat: **4 of those 30 are `ModuleNotFoundError` on the new `remote_response` module**, not meaningful red — the four function-level pins that call `select_zip_member` directly. Those are established by mutation instead (site M below). The two end-to-end pins that reach the defect through the public API fail at master for the right reason. The 9 passers are the both-directions controls.

**Per-site mutation**, new site:

| site | mutation | pins red |
|---|---|---|
| M loose-match other-kind exclusion | `candidates = list(mentions)` | 4 |

Sites E and F (zip member selection in each module) also went 3 -> 4 red, since the new end-to-end pin covers them too. No site mutates to zero.

**Failure-set comparison**, re-measured at the new head against the same merge-base baseline:

| | failed | passed |
|---|---|---|
| merge-base `401f54f2f` | 18 | 13755 |
| this branch | 18 | 13794 |

**+39 = exactly the 39 pins; the two FAILED sets are identical in both directions.**

`ci_comment_density_guard.py`, `lint.sh`, `typecheck.sh` — all rc=0. **CI at `cf0137ef5`: 73 success, 3 path-filtered skips, 0 failed. MERGEABLE.**

Also kept the dead `elif output_type == 'json': return o` removal, per your call.

PR #1991, opened 2026-08-20
fix(gfql): seeded typed 1-hop and leading OPTIONAL MATCH keep bag multiplicity

lmeyerov:
## Why

Two `xfail(strict=True)` pins recorded a runtime **wrong answer**, not a missing feature. Both shapes ran, emitted zero warnings, and returned a plausible deduplicated node set where openCypher returns a bag — while the *sibling* spelling of each pattern was already correct, so the engine disagreed with itself:

| query | before | openCypher | sibling that was already right |
|---|---|---|---|
| `MATCH (a {name:'Ann'})-->(b) RETURN b.id` (two parallel `Ann->Bob` edges) | `[2]` | `[2, 2]` | unseeded `MATCH (a)-->(b) RETURN b.id` → `[2,2,3,4]` |
| `OPTIONAL MATCH (a)-->(b) RETURN a.id` | `[1, 2, 3]` | `[1, 1, 2, 3]` | `MATCH (a)-->(b) RETURN a.id` → `[1,1,2,3]` |

## Root cause — two compile-time carve-outs in the binding-row gate

`_forces_relationship_multiplicity_projection_bindings` decides whether a projection runs on binding rows (bag) or on the per-alias node table (set). Two of its early exits were unsound:

1. `_seeded_typed_hop_reduction_is_value_correct` claimed a selectively-seeded single hop projecting only destination properties was "already answered without binding rows" by the seeded typed-hop fast path. That fast path answers with a destination-node **SET** (`drop_duplicates(subset=[node])` / `.unique(subset=[node])`), so the claim is false exactly when parallel edges exist — the shape the pin names. **Deleted.**

2. The OPTIONAL veto fired for *any* query containing an OPTIONAL clause. That is much broader than the null-extension it protects: a **sole leading** `OPTIONAL MATCH` binds nothing before it, so no row can go unmatched — it is a plain MATCH for row purposes, and the zero-match case is served by the empty-result-row null extension, which never consults binding rows. **Narrowed** to `_sole_leading_optional_match` (one match clause, optional, no reentry/WITH/UNWIND/CALL). The `test_optional_match_no_match_null_extension_preserved` control that the veto existed for stays green.

## Keeping the fast lane

Deleting carve-out (1) alone moves the LDBC IS5 shape off the fast path. Measured on 200k nodes / 1M edges, `MATCH (m {id: 7})-[{type:'KNOWS'}]->(p) RETURN p.a, p.b`, pandas, median of 20:

| | ms |
|---|---|
| master (fast path, wrong bag) | 7.4 |
| carve-out removed, no fast-path change | **124.5** |
| this PR (fast path + bag expansion) | 7.6 |

So the fast path now recognizes the multiplicity-preserving `rows(binding_ops=[n0,e1,n2])` lowering of the same seeded shape — the binding ops are `ops[:3]` verbatim, so the seed-first reduction is the identical query — and re-expands one destination row per matched edge. The expansion short-circuits when the two sides have equal height (the de-dup removed nothing), which is the common no-parallel-edge case, so the served shape stays at fast-path cost.

## Deliberate, named consequence

Shapes where the **Cypher** fast path declines (a datetime property; a requested-vs-actual engine mismatch) no longer reach the **native chain** fast path either, because the lowering they see changed. On pandas they now show the same `int64 -> float64` rows-pivot upcast that every other pandas lane already shows — including the shapes the Cypher fast path *serves*, which cast to it deliberately. That artifact is pre-existing and still tracked; what changed is that the declined and served shapes now agree instead of disagreeing. Two pins were updated to assert plain fast-vs-full frame equality (which is what they asserted before the int64 claim was added), and the now-unreachable `_assert_values_equal_conformant_dtypes` helper was deleted rather than left dead.

## Pins

- `test_row_multiplicity_semantics.py::test_seeded_parallel_edge_multiplicity` — xfail removed (pandas + polars)
- `test_path_trail_semantics.py::test_seeded_parallel_edge_multiplicity` — xfail removed (pandas + polars)
- `test_row_multiplicity_semantics.py::test_leading_optional_match_keeps_multiplicity` — xfail removed (pandas + polars)
- **new** `test_seeded_typed_hop_fastpath.py::test_fast_path_engages_on_the_bag_lowering_of_a_property_return` — anti-vacuity for the fast-path extension: deleting the `binding_ops` branch keeps every value test green and turns this red, which is the only thing standing between IS5 and the 124ms lane
- **new** `test_seeded_typed_hop_fastpath.py::test_parallel_edges_keep_their_row_through_the_fast_path` (pandas/polars/cudf) — the bag through the fast path itself

Local: `graphistry/tests/compute` full run, `bin/lint.sh`, `bin/typecheck.sh`, `bin/ci_comment_density_guard.py` all rc=0. GPU lanes exercised (cuDF 25.10).


lmeyerov:
## CI red → root-caused and fixed in `fe496e538`

`test_indexed_bindings::test_destination_property_projection_dtype_parity[pandas]` — the seeded fast path returned `float64`/`object` where the query's own generic answer was `int64`/`bool`.

**It is not a null promotion.** There is no null round-trip on this path; the promotion was an *explicit* `astype("float64")` in the fast path, written to mimic the pandas rows-**pivot**. Measured lane matrix, same query, `MATCH (source {public:100})-[:A]->(destination) RETURN destination.rank AS r, destination.flag AS f, destination.id AS i`:

| | generic | fast path |
|---|---|---|
| pandas, un-indexed | `r=float64 f=object` | `r=float64 f=object` |
| pandas, indexed | `r=int64 f=bool` | `r=float64 f=object` ← **the bug** |
| polars, indexed | `r=int64 f=bool` | `r=int64 f=bool` |
| cuDF, indexed | `r=int64 f=bool` | `r=int64 f=bool` |

**Which side changed:** neither *rows* side. Row counts are 3 on every cell above — the bag, correct, unchanged by this. What changed is *which lane serves*: once the bag lowering (`rows(binding_ops=...)`) replaced `rows(table=nodes, source=...)` for this shape, the **indexed connected-bindings kernel** became eligible, and that kernel never pivots. So the generic answer moved to `int64`/`bool` while the fast path kept casting for a pivot that no longer runs. The fast path was the lone divergent lane, on the lone divergent engine.

**Dtype decision:** the cast now keys on `index_ctx`, which the function already computes — the same signal that decides whether the indexed kernel can serve. pandas therefore matches *its own generic answer in both index states*, and matches polars/cuDF on the indexed one. No nullable-integer dtype is needed (nothing here is null), and no engine gains a dtype it did not already return — pandas loses one divergence. This moves toward the #1982 cross-engine dtype contract, not away from it.

**Cross-engine check (run):** polars and cuDF were already `int64`/`bool` on both lanes and are untouched by the change (their branches never applied the pivot casts). Verified on cuDF 25.10 and polars 1.42 locally.

**No pin was relaxed.** The parity assertion did its job and is unchanged — it is the thing that caught this. Both directions of the new rule are killed by *existing* pins (measured): forcing "always keep source dtypes" reddens `test_seeded_typed_hop_fastpath::test_pandas_int_bool_dtype_parity`; dropping the rule reddens `test_indexed_bindings::test_destination_property_projection_dtype_parity`.

**Multiplicity oracles re-confirmed red at merge-base** (mutation-run, all killed):
- restore `_seeded_typed_hop_reduction_is_value_correct` → `test_seeded_parallel_edge_multiplicity` red in both files
- restore the blanket OPTIONAL veto → `test_leading_optional_match_keeps_multiplicity` red (pandas + polars)
- make the bag expansion a no-op → 9 red, incl. all three engines of `test_parallel_edges_keep_their_row_through_the_fast_path`
- delete the `binding_ops` fast-path branch → 10 red engagement pins (values stay correct; this is the perf guard)

**Failure-set comparison, `graphistry/tests/compute`, both directions:** master 2 failed / 12727 passed / 43 xfailed; this branch 2 failed / 12746 passed / 37 xfailed. The same two failures on both (`test_categorical_searchany_decline_or_correct`, `TestChainCoercion::test_chain_dask_edges` — pre-existing, unrelated). Nothing went fail→pass unexpectedly. The 6 removed xfails are exactly the three un-xfailed pins × 2 engines.

Scoped to this PR; #1992 and #1993 are untouched.

PR #2001, opened 2026-08-20
fix(gfql): treat binding seed ids as identities

### LINKED ISSUE #1996: GFQL: polars-gpu answers a different count() than every other engine on duplicate node-id rows (fast-path vs generic-route divergence)
Found on dgx-spark (GB10, cudf 26.02.01, cudf-polars 26.02.01) during the 2026-08 GPU validation. **Root-caused as engine-independent** — polars-gpu is the surface that currently exposes it, not the cause.

## Repro

```
MATCH (p {kind:'P'})-[]->(c {kind:'C'}) RETURN c.city AS city, count(*) AS n ORDER BY n DESC, city ASC
```

on a fixture with **duplicate node-id rows** (`_dup_start_node_rows_data`):

| engine | result |
|---|---|
| pandas | LA=6, NY=2, SF=1 |
| polars | LA=6, NY=2, SF=1 |
| cudf | LA=6, NY=2, SF=1 |
| **polars-gpu** | **LA=10, NY=3, SF=1** |

Repro script: `dgx-spark:/home/lmeyerov/gpuval-repro/probe_dup.py`.

## Why polars-gpu differs — and why that is not the real bug

The OLAP fused fast path is currently **inert on polars-gpu**: cudf_polars 26.02 cannot execute its plan (`NotImplementedError: Unhandled map function unnest`, `No GPU support for DataFrameScan(... 'node_type': Null ...)`, and a `type_dispatcher` `Invalid type_id`). The raise escapes `_execute_single_hop_grouped_aggregate_fast_path` **before its eager twin can serve**, so polars-gpu lands on the generic route while every other engine lands on the fast path.

**The two routes disagree.** Confirmed by stubbing the fast path to `None` (`probe_generic.py`): with the fast path disabled, **pandas, polars and cuDF all answer 10/3 too**. So this is a pre-existing **fast-path-vs-generic-route multiplicity divergence** on duplicate node-id rows, and polars-gpu merely makes it user-visible today.

## Which answer is right

Under openCypher bag semantics, duplicate node-id rows in the source table should each contribute, so the generic route's higher count may be the correct one and the fast path may be the side that silently dedupes. **That needs deciding** — closely related to #1994 (whole-entity projection multiplicity), which is the same "the fast path answers a set where a bag is required" family.

## Expected

The fast path and the generic route must agree, whichever answer is correct. Today a user gets a different count purely from which engine they selected, with no warning.

## Note on discovery

This is only observable with a working GPU. There is **no CI lane that runs cuDF or polars-gpu** (`bin/ci_gpu_gate_audit.py` reports "37 gates across 23 test files / workflows setting TEST_CUDF: NONE"), so nothing in CI can catch this class.

lmeyerov:
## Summary

- treat node ids, not duplicate physical node-table rows, as identities when generic binding traversal starts
- preserve first-encounter order in pandas/cuDF and Polars seed state
- keep relationship rows as the source of MATCH bag multiplicity, including genuine parallel edges

## Correctness oracle

The fixture duplicates storage rows for node ids 1 and 2 but adds no graph identities or relationships. Its hand-derived grouped counts are `LA=6, NY=2, SF=1`; adding one parallel `1 -> 5` relationship changes only LA to 7. Before this fix the generic route returned `10/3/1` and `12/3/1` by multiplying paths from duplicate seed rows.

## Semantics boundary

GQL-conformant Cypher preserves binding-row duplicates for plain `RETURN` / `RETURN ALL`; only `DISTINCT` deduplicates ([Neo4j Cypher RETURN](https://neo4j.com/docs/cypher-manual/current/clauses/return/)). This change follows that boundary:

- duplicate physical node-table rows with the same node id collapse only at the initial identity wavefront
- relationship match rows are never collapsed by this change
- a genuine parallel edge therefore remains an additional match row and changes LA from 6 to 7

## Validation

- exact reviewed head: `a888c33f2d42f626bb0754fb1186b2d1c1bf6561`
- topology: four single-parent commits directly on `master@d509e428b0b63a7dd6d92ee5271e75afd15ca2bf`; zero merge commits
- incremental diff: four intended files, +94/-6; whitespace clean
- prohibited added-line audit: zero `Any`, `object`, `cast`, `getattr`, `setattr`, `hasattr`, or unqualified `List`/`Dict`; finite engine values use `Literal`
- focused safe CPU regression: 4 passed / 4 GPU cells deliberately deselected
- `./bin/lint.sh`: syntax, Ruff, type-hygiene, comment-density, and relative-import guards clean
- `./bin/typecheck.sh`: mypy 2.3.1, 336 source files, zero issues
- no Docker, local GPU, or broad local suite; the current GitHub Checks panel is authoritative for the full matrix

## Review resolutions

- **Stacking:** the stale predecessor merges were removed. The branch is now four linear commits directly on landed #2000 master.
- **Positive/negative boundary:** duplicate physical seed rows are the positive case that must collapse; a real parallel relationship row is the negative control that must remain multiplicative. Both cases run across the engine matrix.
- **Anti-vacuity without dynamic patching:** the regression uses semantically neutral `SKIP 0` to select the generic route and `assert_fast_path(..., served=False)` to prove the grouped fast path was consulted and declined.

Closes #1996

## Reviewer handoff

Please review the incremental four-file diff for:

1. seed-state deduplication is by node id and preserves first encounter order
2. relationship rows, including parallel edges, still determine MATCH bag multiplicity
3. the regression proves the generic route without `monkeypatch.setattr` or another dynamic-attribute escape

## Stack / landing

- predecessor #2000 is landed at `d509e428b0b63a7dd6d92ee5271e75afd15ca2bf`
- this PR is based directly on `master`
- successor #2002 must be rebased onto the resulting master after this PR lands
- land only the exact green head using GitHub **Create a merge commit**; do not delete the branch while successors are being repaired

### DIFF
--- a/graphistry/tests/compute/gfql/cypher/test_binding_seed_identity.py
+++ b/graphistry/tests/compute/gfql/cypher/test_binding_seed_identity.py
@@ -1,0 +1,89 @@
+"""Generic binding rows use node ids as identities.
+
+Duplicate physical node rows do not add nodes or relationships. Match rows come
+from relationship rows, including parallel relationships.
+"""
+from typing import Dict
+
+import pandas as pd
+import pytest
+
+import graphistry
+import graphistry.compute.gfql_unified as gfql_unified
+
+
+try:
+    import cudf
+
+    HAS_CUDF = True
+except ImportError:
+    HAS_CUDF = False
+
+
+try:
+    import polars as pl
+
+    HAS_POLARS = True
+except ImportError:
+    HAS_POLARS = False
+
+
+ENGINES = [
+    "pandas",
+    pytest.param("cudf", marks=pytest.mark.skipif(not HAS_CUDF, reason="cudf not installed")),
+    pytest.param("polars", marks=pytest.mark.skipif(not HAS_POLARS, reason="polars not installed")),
+]
+
+
+def _run_generic(engine: str, *, parallel_edge: bool) -> Dict[str, int]:
+    nodes = pd.DataFrame({
+        "id": [1, 2, 3, 4, 5, 6, 7, 8, 1, 2],
+        "kind": ["P", "P", "P", "P", "C", "C", "C", "C", "P", "P"],
+        "city": [None, None, None, None, "LA", "NY", "SF", "LA", None, None],
+    })
+    edges = pd.DataFrame({
+        "s": [1, 2, 3, 4, 1, 2, 3, 4, 1],
+        "d": [5, 5, 6, 7, 8, 6, 8, 8, 5],
+    })
+    if parallel_edge:
+        edges = pd.concat(
+            [edges, pd.DataFrame({"s": [1], "d": [5]})],
+            ignore_index=True,
+        )
+
+    if engine == "cudf":
+        graph = graphistry.nodes(cudf.from_pandas(nodes), "id").edges(
+            cudf.from_pandas(edges), "s", "d"
+        )
+    elif engine == "polars":
+        graph = graphistry.nodes(pl.from_pandas(nodes), "id").edges(
+            pl.from_pandas(edges), "s", "d"
+        )
+    else:
+        graph = graphistry.nodes(nodes, "id").edges(edges, "s", "d")
+
+    result = graph.gfql(
+        "MATCH (p {kind:'P'})-->(c {kind:'C'}) "
+        "RETURN c.city AS city, count(*) AS n ORDER BY city ASC",
+        engine=engine,
+    )._nodes
+    if hasattr(result, "to_pandas"):
+        result = result.to_pandas()
+    return {str(row.city): int(row.n) for row in result.itertuples(index=False)}
+
+
+@pytest.mark.parametrize("engine", ENGINES)
+@pytest.mark.parametrize(
+    "parallel_edge,expected",
+    [
+        (False, {"LA": 6, "NY": 2, "SF": 1}),
+        (True, {"LA": 7, "NY": 2, "SF": 1}),
+    ],
+)
+def test_generic_binding_seed_ids_are_identities(
+    engine: str, parallel_edge: bool, expected: Dict[str, int], monkeypatch: pytest.MonkeyPatch
+) -> None:
+    monkeypatch.setattr(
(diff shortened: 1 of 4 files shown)


lmeyerov:
Is incorrectly stacked?

graphistry/tests/compute/gfql/cypher/test_binding_seed_identity.py
```diff
@@ -0,0 +1,94 @@
...
+    chain: Chain,
+    *,
+    engine: EngineAbstractType,
+    reentry_start_nodes: typing.Optional[DataFrameT] = None,
+) -> typing.Optional[Plottable]:
+    _ = base_graph, chain, engine, reentry_start_nodes
+    return None
+
+
+def _run_generic(engine: _GFQLEngine, *, parallel_edge: bool) -> typing.Mapping[str, int]:
```
lmeyerov, line 40: Is this sufficient, positive and negative testing on either side of the boundary?

lmeyerov:
Resolved. The stack was stale after #2000 was rewritten and landed. #2001 is now rebased directly onto `master@d509e428b0b63a7dd6d92ee5271e75afd15ca2bf` at exact head `a888c33f2d42f626bb0754fb1186b2d1c1bf6561`: four single-parent commits, zero merges, and only the four #2001 files in the incremental diff. Successor #2002 will be rebased only after this exact head lands.

graphistry/tests/compute/gfql/cypher/test_binding_seed_identity.py
lmeyerov: Yes—the regression now pins both sides of the boundary. Duplicate physical node-table rows for ids 1/2 are the positive case: the initial traversal wavefront must collapse them as the same graph identities (`LA=6, NY=2, SF=1`). A genuine parallel `1 -> 5` relationship is the negative control: relationship match rows must not be deduplicated, so only LA increases to 7. Both cases run across the engine matrix. I also removed `monkeypatch.setattr`: neutral `SKIP 0` selects the generic route, and `assert_fast_path(..., served=False)` proves the grouped fast path was consulted and declined, so the value assertion cannot pass vacuously.

Issue #1937, opened 2026-08-18
GFQL decline-ergonomics batch: P1M/2 avg-month, cross-kind rebind at compile time, edge-alias-vs-source-column attribution, all-NULL identity projection ruling


lmeyerov:
Typed-decline improvements recorded by the 2026-08 amplification audits, none silent-wrong, batched:

1. `duration('P1M') / 2` declines typed; Neo4j serves via the 30.4375-day average month (#1919 category-2). Decide: conform or keep the stricter decline.
2. Cross-kind rebinds (`WITH a AS r`, node→edge) surface as an execution-time `GFQLTypeError`; #1912's audit showed widening the guard to compile time passes every pin ("conservative-improving"). Do the widening.
3. An edge alias named after the edge SOURCE column destroys that column and errors with confusing attribution ("column s does not exist") — typed but misleading (#1912 residual pin).
4. **Owner ruling requested** (#1897 round 2 flagged explicitly): all-NULL identity projection (`RETURN p.nul AS g, b.id`) now declines where the pre-#1897 code answered correctly by accident; the openCypher answer is recorded in the pin's docstring. Serve it or bless the decline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_01AjbKuKheqDu78oapRT5AYm

lmeyerov:
## Re-scoping: only item 1 needs an owner decision

This batch reads as four questions for the owner. It is not. Verified against master `af494becc`:

### Item 1 — `duration('P1M') / 2` — **LIVE, needs the owner**

```
MATCH (n) RETURN duration('P1M') / 2 AS half   ->  GFQLTypeError (declines)
MATCH (n) RETURN duration('P2D') / 2 AS half   ->  'P1D'          (works)
```

So duration division works for day-based durations and declines only for month-based ones, because a month has no fixed length. Neo4j serves it using a documented 30.4375-day average month.

**The decision, in one line: conform to Neo4j's average-month constant, or keep the decline?** Genuinely two-sided — half of an average month is a quietly lossy answer few users would predict, and a typed decline is defensible for an ambiguous unit. This is the lowest-stakes item in the whole pre-release set.

### Item 4 — all-NULL identity projection — **appears already RESOLVED, no decision needed**

The shape recorded in this issue answers correctly on master:

```
[...]

PR #2004, opened 2026-08-20
fix(gfql): reject cross-kind WITH rebinds early

lmeyerov:
## Scope

Addresses graphistry/pygraphistry#1937 item 2 only; this PR does not close the issue.

The local Cypher compiler now rejects a `WITH` projection that rebinds a bare MATCH-bound whole-entity alias onto another live MATCH-bound entity alias, including node-to-edge and edge-to-node cases. These shapes previously escaped the same-kind guard and could resolve rows and properties against different bindings. The decline uses the existing typed `E108` contract.

## Final stack

- Base: `master`, exact prepared base `222799c6e2890536080f68cb0e6e7f957124d77c`
- Head: `dd044ea4d68c6f9e529b5de693d31b341d2d28ed`
- Linear delta: four one-parent commits, zero merge commits
  - `9b5ba26d9` — reject cross-kind WITH rebinds early
  - `36090a8c5` — tighten scalar rebind error contracts
  - `914e351c3` — type the cross-kind boundary tests
  - `dd044ea4d` — add the release changelog entry
- Incremental scope: three files, +36/-19
  - `CHANGELOG.md`
  - `graphistry/compute/gfql/cypher/lowering.py`
  - `graphistry/tests/compute/gfql/test_alias_scoping_semantics.py`

## Boundary contract

Error side, asserted directly at compile time:

- node alias onto a live edge alias → exact `E108`, value `a AS r`
- edge alias onto a live node alias → exact `E108`, value `r AS b`
- existing node-to-node and edge-to-edge live-alias rebinds remain exact typed declines

Safe and adjacent controls remain unchanged:

- self-renames and plain carries
- fresh whole-entity targets
- property/scalar projections that shadow a live alias name
- terminal `RETURN a AS b`
- `WITH` output consumed by a reentry `MATCH`
- earlier, more specific UNWIND-collision validation
- the existing scalar-then-property typed error contract

## Review resolution

The sole review question asked whether both sides of the error boundary were sufficiently tested. The final test matrix covers both cross-kind error directions plus same-kind errors, scalar/property vs entity sources, live vs fresh/reentry targets, carries/self-renames, terminal RETURN, and validation precedence.

## Validation

CPU/code gates only; no Docker, cuDF/GPU, DGX, or broad local suite:

- focused boundary tests: 40 passed / 54 deselected
- strict added-line audit: zero `Any`, `object`, `cast`, `getattr`, `setattr`, `hasattr`, unqualified `List`/`Dict`, or broad `str` annotations
- `./bin/lint.sh`: pass, including Ruff, type-hygiene, comment-density, and relative-import guards
- `./bin/typecheck.sh`: pass, mypy 2.3.1, 336 source files
- `git diff --check`: pass
- changelog: added under `[Development]` → `Fixed`

## Landing

Maintainer-authorized merge-commit landing with admin override only after this exact head is fully green. Retain the source branch.


graphistry/compute/gfql/cypher/lowering.py
```diff
@@ -6432,14 +6432,12 @@ def _reject_with_rebind_onto_live_alias(query: CypherQuery) -> None:
             continue
         for item in clause.items:
             source = item.expression.text.strip()
-            source_kind = pattern_aliases.get(source)
-            if source_kind is None:
+            if source not in pattern_aliases:
```
lmeyerov, line 6435: Is this sufficient positive negative testing on either side of the error boundary ?

lmeyerov: Yes—after the follow-up test cleanup, this is sufficient for the item-2 error boundary.

The error side is pinned directly at compile time in both cross-kind directions:

- node onto live edge alias: `a AS r` -> exact `E108`
- edge onto live node alias: `r AS b` -> exact `E108`

The adjacent side is discriminating: self-renames and plain carries execute; scalar/property projections may shadow a live alias; terminal RETURN renames and WITH-to-MATCH reentry execute; fresh-name whole-entity renames retain the existing typed unknown-alias behavior; UNWIND collision keeps its earlier specific validation; and scalar-then-property misuse retains its existing typed error.

That covers the actual predicate boundary: bare entity vs scalar/property source, live MATCH-bound vs fresh/reentry target, both node/edge directions, and validation precedence. The cross-kind tests are explicit functions, the finite engine parameter is a `Literal`, and the incremental scan adds zero broad `str` annotations or banned typing/dynamic helpers.

Focused CPU result: 40 passed / 54 deselected. Lint/hygiene and mypy (336 files) are green. No Docker or GPU was used.

PR #2005, opened 2026-08-20
fix(gfql): preflight remote engine requests

### LINKED ISSUE #1957: GFQL remote: engine='auto' is resolved from the CLIENT's installed libs (frameless graph sends engine=cudf); polars graphs POST then die on an untyped ValueError

lmeyerov:
## Summary

- share remote-engine preflight between `gfql_remote()` and `python_remote()`
- preserve explicit pandas/cuDF requests unchanged in the request body
- reject unsupported explicit engines with typed `E405` before auth refresh, upload, or POST, with an accurate engine-specific diagnostic

This completes the remaining #1957 explicit-engine honesty and side-effect contract. The Polars work in this release is the local GFQL `polars`/`polars-gpu` engine; the public remote contract remains pandas/cuDF. Unsupported Polars-backed remote frames already decline before transport, and this PR makes explicit unsupported engine requests decline consistently. `engine="auto"` remains unchanged.

## Exact state

- base: `master` at `a2b1b44662161ed886665ddf92f7b14fd53b1bda`
- head: `55161a11d57207a66dd977279c5cccabddfb84e0`
- scope: 8 files, +205/-36
- GitHub: `MERGEABLE`; `BLOCKED` reflects required review only
- all 3 review threads are resolved and current

## Validation

- hosted exact-head rollup: 61 passed, 7 conditional skips, 0 failed/cancelled, 0 pending
- Read the Docs and downstream changed-line coverage passed
- bounded local focused suite: 72 passed, 2 expected optional skips
- Ruff passed all 7 changed Python files
- touched-file mypy passed all 7 changed Python files
- type-hygiene and comment-density ratchets show no growth
- complete added-line prohibited-pattern scan has zero hits
- `git diff --check` is clean
- no live remote service, Docker, local GPU, DGX, or broad suite was used

## Reviewer decisions

1. Confirm both public remote APIs resolve unsupported explicit engines before auth refresh, upload, or POST.
2. Confirm explicit pandas and cuDF values reach the request body unchanged.
3. Confirm AUTO behavior is intentionally unchanged.
4. Confirm E405 carries accurate API, field, value, and suggestion context.
5. Confirm local Polars support does not imply a remote Polars execution protocol in this release.

## Landing

PR #2004 is already merged and this PR is rebased directly onto current `master`; there is no dependent successor. The branch-development agent will not merge this PR without separate owner authorization.

Closes #1957


graphistry/compute/chain_remote.py
```diff
@@ -150,20 +153,10 @@ def chain_remote_generic(
 
     strict_level = resolve_strict_level(self, strict=strict)
 
-    if not api_token:
-        self._pygraphistry.refresh()
-        api_token = self.session.api_token
-
```
lmeyerov, line 156: Why is it ok to drop these ?

graphistry/compute/python_remote.py
```diff
@@ -150,16 +156,6 @@ def task(g: Plottable) -> Dict[str, Any]:
     
     if not dataset_id:
         raise ValueError("Missing dataset_id; either pass in, or call on g2=g1.plot(render='g') in api=3 mode ahead of time")
-    
-    assert format in ["json", "csv", "parquet"], f"format should be 'json', 'csv', or 'parquet', got: {format}"
```
lmeyerov, line 154: Some of these still seem valid

graphistry/compute/remote_df_io.py
```diff
@@ -46,8 +53,36 @@ def _is_pandas_like(df: Optional[DataFrameT]) -> bool:
...
+    """Resolve a supported remote engine before auth, upload, or transport.
+
+    :param engine: Requested engine or ``auto``.
+    :param graph: Graph used to resolve ``auto``.
+    :param api_name: Public entry point named in the error message.
+    :return: A pandas or cudf engine.
+    :raises GFQLRemoteError: When the resolved engine is not supported remotely.
+    """
+    resolved = resolve_input_engine(engine, graph)
+    if resolved in (Engine.PANDAS, Engine.CUDF):
```
lmeyerov, line 70: I thought we are adding polars in this release

lmeyerov (commented):
We should enable polars, polars-gpu engines on backend

This pr feels a bit bigger than expected...

graphistry/compute/chain_remote.py
lmeyerov: These were moved, not dropped. Credential refresh now occurs after output/engine/frame/query/schema preflight so an invalid explicit engine cannot refresh credentials, upload, or POST. `test_explicit_unsupported_engine_declines_before_side_effects` asserts refresh/upload/POST are untouched, while the supported-engine test still reaches the mocked POST with pandas/cuDF unchanged.

graphistry/compute/python_remote.py
lmeyerov: Agreed that these checks remain valid; they are retained at earlier/shared seams. The format assertion moved before code validation, the dataset-id/upload checks remain below auth refresh, and only the duplicated engine block was replaced by `resolve_remote_engine()` before side effects. Focused tests pin both supported wire preservation and unsupported pre-side-effect decline.

graphistry/compute/remote_df_io.py
lmeyerov: The Polars work in this release is the local GFQL `polars`/`polars-gpu` engine. The public remote guide/docstrings still specify pandas/cuDF, and master already declines Polars-backed remote frames before POST with E404 (`5301009c5`). Adding a remote Polars execution protocol would require separate server and result-frame support. This PR therefore makes explicit `polars`/`polars-gpu` decline typed E405 before side effects, keeps AUTO unchanged, and now states that boundary in CHANGELOG.

PR #2030, opened 2026-09-04
perf(gfql): memoize node dtype reads per frame (#2029)

### LINKED ISSUE #2029: perf(gfql): Cypher compile scans every object column of the node table on each call (pandas point lookups 100+ ms on SNB)
On the LDBC SNB SF0.1 graph as pyg-bench binds it (327,588 nodes × 32 columns, 27 of them `object`, non-unique ids, boolean `label__*` columns), every `g.gfql("MATCH ...")` call on the pandas engine costs 105–126 ms regardless of shape — even when `gfql_explain` reports the seeded typed-hop fast path as served. The native `g.gfql([n({"id": x})])` is 0.7 ms on the same graph.

cProfile of `MATCH (m:Message {id: X})-[:HAS_CREATOR]->(p:Person) RETURN p.id AS personId, p.firstName AS firstName, p.lastName AS lastName` (3 calls, 0.319 s):

- `_compile_string_query` → `_node_dtypes_cache_key` → `filter_by_dict._read_node_dtypes` → `_object_column_holds_non_strings` × 96 calls: **0.305 s** (`pd.api.types.infer_dtype(frame[column], skipna=True)` over the full column, for all 27 object columns, on every compile);
- the fast path itself (`_execute_seeded_typed_hop_fast_path` → `_seeded_typed_return_dst_pandas_cudf`): 0.012 s.

So the compile-cache *key* is O(rows × object columns) per call, and it dominates any point query on a wide pandas node table. On a 3-column synthetic graph the same scan is ~20 ms (the number the docs' SNB page attributed to "per-call compile and row pipeline").

Proposed: memoize `_read_node_dtypes` per node-frame identity (the same identity + fingerprint contract the resident indexes use), so the string-content gate is computed once per frame and the compile key becomes O(columns); alternatively key the compile cache on `(id(nodes), shape, dtype names)` and defer the content check to the pushdown site that needs it. Expected: SNB pandas point lookups drop from ~110 ms to the fast path's ~5 ms; the polars engine is unaffected (no object-dtype inference).

Evidence: pyg-bench `results/snb-aligned-indexed-20260904` (pandas SF0.1 seed lookup 30.1 ms, SF1 359.5 ms) and the profile above (local, pygraphistry 3fb216dd).

lmeyerov:
Fixes #2029.

`_compile_string_query` keys the compile cache on the node dtypes, which forces `_LazyNodeDtypes` to materialize on every string query; `_read_node_dtypes` then runs `pd.api.types.infer_dtype` over the full values of every `object` column (the string-content gate for predicate pushdown). On the LDBC SNB SF0.1 node table as pyg-bench binds it (327,588 rows × 32 columns, 27 object), that scan was 305 ms of a 319 ms 3-call profile; the seeded typed-hop fast path it gates ran in 12 ms.

Change: `_read_node_dtypes` memoizes per node frame — keyed on `(id(frame), engine)`, valid while a weakref still resolves to the same object and the (length, columns) fingerprint matches (the resident indexes' identity contract), bounded to 32 entries, registered with `cache_registry` so `gfql_clear_caches()` empties it. No semantic change: the gate still scans on the first read of any frame.

Measured on the real SF0.1 graph (pandas, index resident, warm median of 5):

| query | before | after |
|---|---|---|
| official IS5 (`MATCH (m:Message {id})-[:HAS_CREATOR]->(p:Person) RETURN p.id, p.firstName, p.lastName`) | 105.9 ms | **1.95 ms** |
| official IS1 (two-alias, 8-column projection; no fast path) | 126.3 ms | 22.4 ms |
| `MATCH (p:Person {id}) RETURN p.firstName, p.lastName` | 115.6 ms | 13.7 ms |

polars is unaffected (no object-dtype inference). Tests: memo hit does not rescan, rebound/grown frame rescans, per-engine entries, `gfql_clear_caches()` empties it, end-to-end string queries. Lint/type-hygiene/comment guards pass.

https://claude.ai/code/session_01WwMmVFo44ADiRRj5cxh7i1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,14 @@ All notable changes to the PyGraphistry are documented in this file. The PyGraph
 The changelog format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and all PyGraphistry-specific breaking changes are explictly noted here.
 
-## [0.59.0 - 2026-08-31]
+## [Development]
 <!-- Do Not Erase This Section - Used for tracking unreleased changes -->
+
+### Fixed
+
+* GFQL: every Cypher string query re-read the node table's dtypes to build its compile-cache key, and that read scans the values of every `object` column (the string-content gate for predicate pushdown). On a wide pandas node table this cost more than the query it keyed: an SNB SF0.1 seeded lookup on 327k nodes × 27 object columns spent 105 ms of 106 ms there. The read is now memoized per node frame (identity plus a length/columns fingerprint, the resident indexes' contract) and cleared by `gfql_clear_caches()`; the same lookup now takes 2 ms (#2029).
+
+## [0.59.0 - 2026-08-31]
 
 ### Breaking
 
--- a/bin/test-polars.sh
+++ b/bin/test-polars.sh
@@ -79,6 +79,7 @@ POLARS_TEST_FILES=(
     # polars params (bound validation, hops=None run-to-closure, edges-only node output)
     # only ever run in this lane
     graphistry/tests/compute/gfql/test_hop_semantics_1918.py
+    graphistry/tests/compute/gfql/test_node_dtypes_memo_2029.py
     # #1882/#1913-f4/#1879 crash-family pins: the polars params (filter helpers on polars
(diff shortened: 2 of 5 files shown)


lmeyerov:
looks fine, bigger thing is locking in expected rough latency (pyg-bench?) to avoid overall latency regressions again, at least on a few common basic query shapes of varying complexity like as happened here

Issue #2034, opened 2026-09-04
gfql: rows(table=nodes, source=alias) multiplies rows for duplicate node ids and joins null ids to each other


lmeyerov:
Found while pinning the seeded node-lookup fast path (parity harness on the branch for the seeded fast paths, 2026-09-04).

1. Duplicate node ids: with two node rows carrying `id = 7`, `MATCH (p {id: 7}) RETURN p.age AS a` on the full path returns 8 rows on pandas and cuDF (2 duplicate rows → 2^3 through the chain's forward/reverse/combine self-joins); polars returns 2. The expected answer is one row per matching node row (2).
2. Null ids: with two rows whose `id` is null, `MATCH (p {name: 'n3'}) RETURN p.id, p.age` returns both null-id rows on the full path (NaN-keyed merge matches NaN to NaN) instead of the one row that matches the predicate. A seeded typed hop whose seed row has a null id raises `GFQLTypeError` on the full path (merge object vs float64) instead of returning no rows.

The seeded fast paths (`_execute_seeded_node_lookup_fast_path`, `_execute_seeded_typed_hop_fast_path`) return the intended answer for these inputs; `graphistry/tests/compute/gfql/test_seeded_node_lookup_fastpath.py::test_node_lookup_returns_each_duplicate_id_row_once` pins the fast answer on all three engines. The full path should agree; until it does, the differential parity tests exclude duplicate and null ids.

Where: `graphistry/compute/chain.py` `_chain_impl` combine step (self-join on the node id) and `graphistry/compute/gfql/row/frame_ops.py` `rows(source=...)`; the merges should key on row position for node-frame lookups, and null ids must never link.

lmeyerov:
Disposition (0.60 landing phase): deferred behind the stacks. The seeded fast paths (#2035/#2037/#2038) return the intended one-row-per-node answer for duplicate and null ids and pin it; the full-path chain combine still self-joins on the id column. This is the same chain-combine seam as the cuDF 26.2 duplicate-id divergence in #2043, so both get fixed together right after the stacks land (positional node-frame lookups; null ids never link). Not release-blocking on its own: the shapes that reach the full path with duplicate ids are the ones the fast paths already decline, and the differential harness excludes duplicate/null ids until then.

lmeyerov:
[...]

Issue #2033, opened 2026-09-04
perf(gfql): polars native chain for a seeded typed 1-hop does not take the seeded fast path (3-4x slower than the same Cypher)


lmeyerov:
`g.gfql([n({"id": X}), e_forward({"type": "HAS_CREATOR"}), n({"type": "Person"})], engine="polars")` runs the full lazy forward/backward/combine traversal, while the identical Cypher `MATCH (m {id: X})-[:HAS_CREATOR]->(p:Person) RETURN p` is served by the seeded typed-hop fast path.

Measured locally on a 300k-node table with 27 object columns (RTX 3080 Ti box, best of 5 warm, not a published number): polars native chain 9.5 ms vs Cypher 2.7 ms; pandas native chain 1.8 ms (pandas takes `chain._try_chain_fast_path`).

Where it declines: `graphistry/compute/gfql/lazy/engine/polars/chain.py`, the index consult (~line 982) and the eager single-hop shortcut (~line 1006) both require `_plain_edge`, i.e. `edge_match is None`, and an unfiltered destination node. A typed edge (`edge_match={"type": ...}`) or a destination filter therefore falls through to the full traversal.

Fix shape: relax `_plain_edge` to allow scalar `edge_match` and a scalar destination `filter_dict`, or call `chain_fast_paths._seeded_typed_return_dst_polars` there and rebuild `(nodes, edges)` plus alias tags the way `_tag_fast_path_aliases` does on pandas. Keep `test_seeded_typed_hop_fastpath.py::TestPolarsFastPathGates` (LazyFrame declines, mixed-engine frames decline, varlen declines).

Found while writing `graphistry/tests/compute/gfql/test_gfql_latency_contract.py` (#2031), which for this reason pins Cypher against plain frame ops rather than the native chain.

lmeyerov:
Fix: #2038 (native Polars chains resolve seeds through resident indexes; index-gated, engagement recorded in gfql_explain) — 75/75 green at 4a80847bf, GPU receipt vs master in the same image, round-003 differential sweep parity on every served polars shape. Stacked on #2037; closes on merge.

Issue #245, opened 2021-07-13
[FEA] gremlin should take ResultSets


lmeyerov:
**Is your feature request related to a problem? Please describe.**

Currently, taking an out-of-band gremlin result requires `g2 = g.resultset_to_g(rs)`. That's not super discoverable nor memorable.

**Describe the solution you'd like**

`g2 = g.gremlin(rs)`

**Describe alternatives you've considered**

`g2 = g.resultset(rs)` and current `g2 = g.resultset_to_g(rs)`

**Additional context**

Do before or as part of https://github.com/graphistry/pygraphistry/issues/244

PR #2024, opened 2026-09-04
perf(gfql): numpy form of the undirected seed-rediscovery rule (#2023)

lmeyerov:
Fixes #2023.

## What

`undirected_rediscovered_seed_ids` (the #1918 wavefront rule: a seed survives iff another seed shares its component, or it lies on a cycle — self-loops plus the multigraph 2-core) was a pure-Python loop over every traversed edge, run on every undirected hop step by both the pandas/cuDF and the polars paths. On LiveJournal (34.7M edges), an undirected 2-hop from 50 seeds spent 108 s of 117 s in it under cProfile; the June tree ran the same query in 1.9 s.

Same rule, now as frame operations inside the caller's own engine (review round 1 asked for engine-generic code in its own module, cross-engine tests, no product perf pin, and a changelog entry):

- `graphistry/compute/gfql/seed_rediscovery.py` — pandas and cuDF over `DataFrameT` (merge, isin, groupby via the existing `dataframe_utils` helpers); cuDF frames never leave the GPU.
- `graphistry/compute/gfql/lazy/engine/polars/seed_rediscovery.py` — the polars twin (joins, anti-joins, group_by).
- Rule A = multi-source label propagation from the seeds, then a pointer-jumping union over seed ranks for the labels that meet across an edge. Rule B = leaf peeling to the 2-core plus self-loops. Both iterate to a fixed point.
- `hop.py` shrinks by ~100 lines to a five-line call site; `hop_eager.py` calls the polars twin.

## Tests (`graphistry/tests/compute/gfql/test_seed_rediscovery_2023.py`, parametrized over pandas / cuDF / polars)

Each boundary pins a kept seed and a dropped seed for the same reason: acyclic path (alone vs two seeds), triangle vs pendant off it, parallel edges vs a single edge, self-loop vs its neighbour, star hub vs hub+leaf, absent seed alone vs beside a kept one, NULL endpoint; plus string ids, a narrower seed dtype, empty inputs, a 300-edge pendant path, and random multigraphs equal to the pure-Python oracle per engine. cuDF ran locally on the GPU box. `test_hop_semantics_1918`, `test_hop`, `test_chain`, `test_varlen_bounded_engine_parity_1787`: 400 passed / 73 skipped with cuDF enabled. Lint, type-hygiene and comment guards pass; changelog under `## [Development]`.

## Measured

Rule alone, random graphs, local box (3-run min):

| edges / seeds | pandas | polars | cuDF |
|---|---|---|---|
[...]

PR #2035, opened 2026-09-04
perf(gfql): serve seeded node lookups and two-alias projections on the fast paths

lmeyerov:
Stacked on #2031 (contract pins) → #2030 (dtype memo). Closes the two fast-path coverage gaps behind the SNB point-lookup regression (plan T2.10.2).

**What changes**
- New `seeded_node_lookup` fast path: `MATCH (p {id}) RETURN p` / `RETURN p.a AS x, p.b` resolves the seed through the resident node-id index, a resident node-property index (`gfql_index_node_props`), or one scalar scan, and projects the matched rows directly. Before, this shape ran the full two-pass chain over the node table with no index consulted (the alias on the node op disqualified the native single-op shortcut).
- Seeded typed hop: a bag RETURN may project from both pattern aliases (`RETURN m.id, p.firstName`, the LDBC IS1 form), one row per matched edge, built from the seed and destination rows the reduction already holds. Before, it fell to the row pipeline's connected-bindings frame builder.
- Shared seed resolution (`chain_fast_paths._seed_node_rows`): node-id index → property index → scan, with the ORIGINAL filter dict re-applied through the canonical filter (`filter_by_dict` / `filter_by_dict_polars`) so typed errors, temporal and mixed-type comparisons match the full path exactly. The seeded hop now also benefits from a property-indexed seed.
- Dtype parity rule for the lean projections: the pandas rows-pivot upcast (int/float → float64, bool → object) is applied only where the canonical path pivots — destination alias only, empty results always, and never when the indexed connected-bindings kernel would have served (`_indexed_kernel_admits` mirrors its seed admission and frontier/gather cost gates).

**Evidence** (SNB SF0.1 fixture, local box, not published; pyg-bench sentinel #245): pandas IS1 official 19.1 → 4.0 ms, node-only lookup 13.7 → 1.2 ms; polars IS1 7.9 → 5.7 ms; cuDF IS1 78 → 39 ms; every basic shape now reports `fast_path served` on pandas, polars and cuDF. Contract pins in `test_gfql_latency_contract.py` flip from strict-xfail to pass.

**Tests**: `test_seeded_node_lookup_fastpath.py` (lookup shapes × scan/indexed × 3 engines, independent oracle, property-index engagement with the `node_lookup` trace seam, out-of-shape declines, policy decline, duplicate-id rows, two-alias shapes incl. int/bool from both aliases, no-match dtypes, non-binding indexed seeds, bag multiplicity, multi-seed, hub seed over the frontier gate); cross-alias pins in `test_seeded_typed_hop_fastpath.py` flip from decline to engage-with-parity; engagement map extended; `test_use_policy_sparse_serves_dense_declines` moved to a 2-hop shape. Full `graphistry/tests/compute/gfql` sweep on pandas/polars/cuDF: 11,896 passed, 5 failures = pre-existing local `libcublas.so.12` import errors (identical on master). Guards (ruff, mypy, comment density, type hygiene, Cypher surface) green.

Found on the way: #2034 (full path multiplies rows for duplicate node ids and joins null ids), #2033 (polars native chain skips the seeded fast path).

https://claude.ai/code/session_01WwMmVFo44ADiRRj5cxh1i1


lmeyerov:
Ready for review at f8f2e7355 (rebased on master after #2024/#2028; 77/77 checks green incl. tck-gfql, RTD green).

Hardening per the review bar (audit table in the private plan, `reviews/pygraphistry-2035-hardening/audit.md`):
- Every admission check in `_execute_seeded_node_lookup_fast_path` and the two-alias projection now has a pin on both sides: whole-row RETURN with a suffix, expression items, mixed whole-row + property items, LazyFrame nodes, requested engine ≠ frames' engine, and a destination-side extension dtype all decline to the full path with parity; seed-side extension and datetime columns keep parity; the scan reason (`index_missing` / `served` / `index_stale`) is visible in `gfql_explain`.
- One finding fixed here (`explain.py`): under `index_policy="off"` a served Cypher fast path was reported as `used_index: True` / `index_selected`; `gfql_explain` now derives `used_index` and the decision from index steps only and reports `policy_off`. Pinned on pandas / polars / cuDF.
- Cross-platform: dgx-spark, RAPIDS 26.02 image (cudf 26.02.01), lookup + typed-hop + engagement + latency-contract + index + indexed-bindings + polars-gpu suites: 727 passed / 0 failed / 1 opt-in skip. mypy clean on the changed file; no new Any/ignore/cast.
- pyg-bench lock-ins: the SNB point-latency sentinel (is1/is5/seeded-hop/node-only on pandas + polars, < 10 ms absolute, drift-gated vs the committed baseline that contains this PR) and the docs-numbers acceptance gate.

Suggested order after this: #2036 next (green; it is what keeps the #2028 scaling pin stable — a loaded runner just measured 30.6x vs the 30x bound on a branch without it), then #2037 → #2038 → #2040.

lmeyerov:
READY (f8f2e7355): 77/77 checks + RTD; boundary audit above; real-GPU 727/0/1 opt-in skip; mypy clean; sentinel + acceptance gate lock-ins.

lmeyerov:
Fair challenge on "two projections". In the code, "two" was never a projection count: it meant the two node endpoints of the single hop, and the path already took any number of properties from either endpoint (an eight-property RETURN is served). The benchmark-shaped gap was real though: an edge-alias property (`RETURN r.since`) declined to the full path, and that is the natural n. Pushed in e2db9cb60:

- the seeded typed-hop fast path now projects any number of properties from any alias of the hop (seed node, edge, destination node), one row per matched edge; edge properties keep their source dtypes like the full path;
- a projection that touches any non-destination alias takes the per-edge frame, so a no-match seed with an edge property projects an empty frame instead of reading node columns (a bug the new pins caught);
- pins: all three aliases, edge-only including endpoints, twelve properties across three aliases, parallel edges, no match; two hops and an absent edge property decline with parity;
- CHANGELOG wording now states the general contract rather than the IS1 shape.

What still declines by construction: anything beyond one hop (a different fast path's job) and anything with side channels (policy, WHERE, OPTIONAL). Local: fast-path + index + latency + lowering suites 548 + 1765 passed (the two cuDF t6 lowering failures are the dev-box libcublas environment, identical on master); ruff/guards/mypy clean, no new Any/ignore/cast. Real-GPU receipt for the new shapes follows on this PR. #2037/#2038/#2040 are restacked on this head.

lmeyerov:
Scenario review (include/exclude, per your note; matrix in the private plan `reviews/pygraphistry-2035-hardening/scenarios.md`): 16 shapes probed for served-or-declined plus fast-vs-full parity.

- EXCLUDE found and landed (7e3f98347): an edge alias named like an edge column (`-[type:HAS_CREATOR]->` on a frame with a `type` column), or a node alias named like a node column, now declines on every RETURN form. Before this, the projection's marker insert raised on pandas/cuDF where the full path serves; the polars full path has its own collision residual (#2039/#2046). Pins on all three engines for filtered / projected / not-projected / whole-row.
- INCLUDE candidates, deferred as #2048 (not needed by the SNB shapes, each reopens admission logic): single-alias WHERE on the seed or destination alias, and LIMIT-only whole-row RETURN. Both decline with parity today.
- Everything else (reverse direction, undirected, OPTIONAL, label-only seed, IN seeds, count, DISTINCT, string/bool/two-property seeds, null-valued seed, absent property, two hops) is served with parity or declines with parity.

Real-GPU at 37ada01bf: 766 passed / 0 failed; rerun on 7e3f98347 in flight. #2037/#2038/#2040 restacked on this head.

lmeyerov:
Real-GPU receipt at 7e3f98347 (dgx-spark, RAPIDS 26.02 image, cudf 26.02.01 / polars 1.35.2): lookup + typed-hop + engagement + latency-contract + index + indexed-bindings + polars-gpu suites = 784 passed, 0 failed, 1 opt-in skip. CI on this head and the restacked #2037/#2038/#2040 is running.

lmeyerov:
READY (7e3f98347): 77/77 checks + RTD; real-GPU 784 passed / 0 failed; n-alias generalization, alias/column collision declines, scenario matrix (#2048 for the two include candidates). #2037 e08a05938 / #2038 88ff6257b / #2040 c4d2e9aa2 restacked on this head, 75/75 each.

lmeyerov:
Quality pass per the review checklist, landed as aad106cc9:

- Error amplification: a random-shape differential (`test_random_projection_shapes_match_the_full_path`: random aliases, property subsets from seed/edge/destination, seed hit or miss, ORDER BY/LIMIT suffixes, indexed or not, 12 seeds × 3 engines) found one more real divergence: projecting an edge's endpoint binding (`r.src`) served the ids where the full path resolves an absent property (null). The fast path now declines those items (pinned). On polars, shapes the full path declares unsupported (`NotImplementedError` for the row-op `select`) are held to parity-or-error, and served shapes are checked against the pandas full path as the oracle.
- pyg-bench / contract: the latency contract gains the all-aliases shape (`m.lastName, r.type, p.firstName`) against a faithful two-lookup plain-frame floor: pandas 1.1×, polars 2.5×, cuDF 4.4× locally. The sentinel's IS1/IS5/hop/node-only routes are unchanged by the later commits, so the cdbcc640a baseline stands; the master re-measure covers the cells.
- DRY: the select-item admission loop is duplicated between the two fast paths; noted on #2048 to fold with the WHERE-pushdown work rather than churn this PR again.

Suites: lookup + typed-hop + indexed-bindings + engagement + latency-contract + index + alias-scoping = 825 passed locally; guards/mypy clean. #2037 ab3a32ef3 / #2038 6312e1853 / #2040 ab74105b4 restacked. Real-GPU run of aad106cc9 in flight; CI running.

lmeyerov:
Real-GPU receipt at aad106cc9 (dgx-spark, RAPIDS 26.02 image, cudf 26.02.01 / polars 1.35.2): lookup + typed-hop + engagement + latency-contract + index + indexed-bindings + polars-gpu suites = 829 passed, 0 failed, 1 opt-in skip.

lmeyerov:
READY (aad106cc9): 77/77 checks + RTD; real-GPU 829 passed / 0 failed; random-shape differential + curated boundary pins on 3 engines; latency-contract pin for the all-aliases shape; #2037 ab3a32ef3 / #2038 6312e1853 / #2040 ab74105b4 restacked, 75/75 each.

lmeyerov:
READY + MERGING (7db7968f6) — owner-authorized after the round-002 test-amplification pass.

**Round-002 (timeboxed 1–1.5 h, hypotheses not pre-constrained):** a 90-cell differential matrix (30 shapes × pandas/polars/cuDF, default route vs fast paths disabled vs pandas oracle) over 13 hypothesis families (categorical / string / nullable dtypes, stale indexes, `$param` seeds, naming collisions, WITH/UNION/OPTIONAL/count/DISTINCT composition, parallel-edge multiplicity, policy force/decline, whole-row variants, seed value types, empty results). Found 2 new classes, both fixed here with pins:
- an empty projection kept the node id column at int64 where the full path widens it to float64 (no-match seed, destination label mismatch; pandas + cuDF)
- a whole-row RETURN on pandas next to a categorical column kept neighbouring bool/int columns narrow; the typed-hop lane now declines pandas extension-dtype frames (polars/cuDF keep parity and stay served)

Post-fix matrix: 0 DIFF / 81 parity / 9 both-raise (same exception class on both routes). Artifacts and probe outputs in the plan's `rounds/round-002/`.

**Receipts at 7db7968f6**
- CI: 77/77 check-runs success, incl. tck-gfql, changed-line coverage, polars-lane completeness and the three hygiene guards.
- Real GPU (dgx-spark, `graphistry/test-rapids-official:26.02-gfql-polars`, cudf 26.02.01 / cupy 13.6.0 / polars 1.35.2, safe_run): 865 passed, 1 skipped (opt-in wall-clock gate), 2 xfailed (recorded gaps) — +36 cells over the previous head's 829 with the same skip/xfail set.
- Local broad GFQL suite (`graphistry/tests/compute/gfql` + chain/hop): 12152 passed, 0 failed.
- mypy on the changed product files: no new errors vs master; no new `Any` / `type: ignore` / typing `cast` in product code.
- pyg-bench sentinel (`snb_point_latency`, SF0.1, dgx-spark release runner, pyg-bench main 5cf81ae, run `snb-sentinel-7db7968f6-r1`): every Cypher point served and within the gate vs the 2026-09-04 baseline (pandas IS1 1.97 ms vs 1.97; IS5 1.07 vs 1.05; seeded-hop props 1.03 vs 1.02; entity 3.03 vs 3.01; node-only props 0.59 vs 0.58; entity 2.47 vs 2.49; polars 1.7–5.3 ms, all ≤ baseline except node-only-entity 2.74 vs 2.26, inside the 1.5× gate). The three *native op-list* points read as regressions only because the baseline was measured at the #2038 head, which stacks #2037/#2038 on this PR; those lanes do not exist at this head and land next, after which the sentinel is re-run on master.

Restacked on this head with identical patch-ids: #2037 ff818e147, #2038 0664c3a19 (also drops a duplicated #2033 CHANGELOG bullet a union rebase had left), #2040 682c072aa.

PR #2038, opened 2026-09-05
perf(gfql): use resident seed indexes in native Polars chains

### LINKED ISSUE #2033: perf(gfql): polars native chain for a seeded typed 1-hop does not take the seeded fast path (3-4x slower than the same Cypher)

lmeyerov:
Native Polars op-list queries with named nodes, scalar edge filters, or a filtered destination fell through to the full traversal even when resident indexes could resolve the seed. Single-node native queries also scanned their seed predicates.

Reuse the shared seeded reduction for these directed one-hop shapes and the shared node resolver for single-node queries. The seeded lane serves only when the resident node-id and adjacency indexes cover the hop (validated once, passed into the reduction), runs after the polars index-policy consult so `index_policy` off/force/use and the engine-mismatch diagnostics are unchanged, and records an `indexed_traversal` step so `gfql_explain` reports the engagement. Indexed gathers preserve input table order; native aliases match the full Polars path. Unsupported shapes, alias/property collisions, duplicate node ids, and graphs without a usable index keep the existing full-path behavior; scalar filters raise the same structured schema errors as the full path.

Stacked on #2037. Fixes #2033.

Validation (head cdbcc640a): polars lane `bin/test-polars.sh` 8,238 passed (4 failures are the local libcublas/dask environment, identical on the base); index contract + narrow-combine + seed-resolution + seeded-fast-path suites 436 passed; 33 pins in the new test file (explain/index_trace engagement, policy-off and no-index declines with full-path parity, duplicate-id decline, schema-error parity). Ruff, changed-file mypy, comment-density, type-hygiene, Cypher surface guard and diff checks pass. DGX measurements follow from a dedicated frozen checkout. Found while amplifying: #2039 (polars full path raises on an edge alias colliding with its filtered edge column).


lmeyerov:
READY (479a42194): 75/75 + RTD incl. tck-gfql; index-gated native polars seeded lane with engagement trace; start_nodes negative pin; real-GPU polars-gpu suites green. Stacked on #2037.

lmeyerov:
READY (4a80847bf) — restacked on the new #2037 head after #2035 landed (patch-id identical to the previous head). This restack also drops a duplicated #2033 CHANGELOG bullet that an earlier union rebase had left (one bullet now).

**Round-003 amplification** (owner ask: reflect on the review flow before this lands) ran at this head: 51 shapes × pandas/polars/cuDF, default route vs the polars seed lane and the chain fast path forced to decline, plus the pandas full path as oracle. Polars cells: every served shape (`_try_seeded_chain_polars` engaged and the `native_seeded_hop` trace recorded) is value- and dtype-identical to the polars full path, including seed forms (`None`, absent, float, bool, label-only, two scalars, absent column), NULL-endpoint and dangling edges, string keys, categorical props, stale index, empty edge table, and declines for undirected / multi-hop / fixed-point / edge-first shapes. The only polars-vs-pandas differences (alias/column collisions rendered with `_right` suffixes; repeated node rows answered once per row) are identical on master and are not this lane.

**Cost receipt** (local box, medians of 7, ms; polars seeded typed hop `[n({"id": k}, name="p"), e_forward({"type": ...}, name="e"), n(name="q")]`): indexed 4.00 / 6.24 / 14.22 on master at 10k / 100k / 1M rows vs 1.88 / 1.65 / 1.74 here (flat in table size); unindexed 4.80 / 7.81 / 29.87 vs 4.88 / 8.14 / 26.26 (unchanged). Node lookup unchanged (~0.3 ms).

**Receipts at 4a80847bf**
- Real GPU (dgx-spark, RAPIDS 26.02 image, 13 suites incl. `test_chain.py`): 1331 passed, 3 failed, 1 skipped (opt-in wall-clock gate), 2 xfailed. The 3 failures (`test_chain.py:730` ×2, `:1277`) are identical on master in the same image: the cudf-26.02-only divergences tracked as #2043, not this PR.
- CI: 75/75 check-runs success (incl. tck-gfql, changed-line coverage, polars-lane completeness, the three hygiene guards).
- pyg-bench: the sentinel's polars native points are this PR's lock-in; re-run on master after landing.

lmeyerov:
CI at the rebased head a3991bbf2 (on master 4c5e25281 after #2037 landed): 77/77 check-runs success (tck-gfql, changed-line coverage, polars lane, hygiene guards included). Patch-id identical to the previous head 4a80847bf, so the GPU receipt (1331 passed, 3 failures identical to master, #2043) and the round-003 parity sweep carry over. Ready to merge; #2040 auto-retargets to master afterwards and I rebase it.

Issue #2039, opened 2026-09-05
gfql polars: an edge alias that collides with the edge column its own filter uses raises incompatible-column-type (pandas serves it)


lmeyerov:
## Summary
On the polars engine, a native op-list chain whose edge alias has the same name as an edge column that the same edge step filters on raises `GFQLSchemaError [incompatible-column-type]`. The pandas engine executes the same chain (alias marker authoritative over the colliding column, as `combine_steps` documents).

Same on 0.59.0 (`3fb216dd1`) with and without resident indexes, so it is independent of the fast paths and of the #2037/#2038 stack.

## Repro
```python
import pandas as pd, polars as pl, graphistry
from graphistry.compute.ast import n, e_forward

nodes = pd.DataFrame({"key": [1, 2, 3], "id": [10, 20, 30], "type": ["p", "p", "m"]})
edges = pd.DataFrame({"s": [3, 3], "d": [1, 2], "type": ["HAS_CREATOR", "OTHER"]})
ops = [n({"id": 30}, name="m"), e_forward({"type": "HAS_CREATOR"}, name="type"), n(name="p")]

g_pd = graphistry.nodes(nodes, "key").edges(edges, "s", "d")
g_pl = graphistry.nodes(pl.from_pandas(nodes), "key").edges(pl.from_pandas(edges), "s", "d")
print(g_pd.gfql(ops, engine="pandas")._edges.columns)   # ['s', 'd', 'type']  (marker replaced the column)
g_pl.gfql(ops, engine="polars")                          # GFQLSchemaError: column "type" is numeric but filter value is string
```
A node alias colliding with a node property (`n({"id": 30}, name="id")`) is served by both engines.

## Expected
Both engines agree: either both serve with the documented marker-authoritative collision contract, or both reject the collision with a structured error before execution. The polars full path appears to tag the alias marker onto the edge frame before applying the step's own `edge_match`, so the filter then sees the boolean marker.

## Context
Found while adding native op-list shapes to the pyg-bench SNB point-latency sentinel (the SNB frames carry `id`/`type` columns, so alias/column collisions are realistic). The native polars seeded fast path (#2038) declines alias collisions, so this is the full path. Related: #2034 (rows() duplicate/null ids).

lmeyerov:
[...]

PR #2045, opened 2026-09-05
fix(gfql): apply single-alias predicate pushdown masks by position (#2020)

lmeyerov:
`MATCH (a)-[e]->(t) WHERE a.type IN [...] AND e.e_type IN [...] AND t.type IN [...] RETURN t.id` raised `GFQLTypeError [invalid-node-reference] ... Unalignable boolean Series provided as indexer` on pandas (and cuDF) while polars answered. The single-alias predicate pushdown filtered the alias frame with `frame.loc[mask]`; once an earlier pushdown had narrowed that frame, its labels were no longer a RangeIndex, while the mask evaluated on the renamed view came back with a fresh positional index, so label alignment failed. A mask computed on the same rows is positional by contract, so both pushdown sites (expression and searchAny) now keep rows by position through one helper typed with the repo's engine-agnostic aliases.

Pins (`test_alias_prefilter_alignment_2020.py`): the issue's query returns `['tx1']` on pandas, polars, cuDF; equals the scalar `=` form; and survives a node frame whose index is neither ranged nor ordered. Row-pipeline, lowering, alias-scoping, optional-match and multiplicity suites pass locally (1885; the two cuDF t6 lowering failures are the dev-box libcublas environment, identical on master). Real-GPU run to follow on dgx.

Fixes #2020.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QztW7jYsDd66e8rb8pJNQA

Issue #2020, opened 2026-09-04
gfql cypher pandas: three IN predicates across (a)-[e]->(t) crash with 'Unalignable boolean Series' (polars OK)


lmeyerov:
## Repro (master @ 3fb216dd)

```python
import pandas as pd, graphistry
nodes_df = pd.DataFrame({'id': ['a','b','c','tx1','tx2'],'type': ['person','person','company','transaction','transaction']})
edges_df = pd.DataFrame({'src': ['a','b','a','tx1','tx2'],'dst': ['b','c','tx1','tx2','c'],'e_type': ['knows','works_at','sent','transfer','received']})
g = graphistry.edges(edges_df,'src','dst').nodes(nodes_df,'id')
q = "MATCH (a)-[e]->(t) WHERE a.type IN ['person','company'] AND e.e_type IN ['sent','transfer'] AND t.type IN ['transaction','account'] RETURN t.id AS id"
g.gfql(q, engine='polars')   # OK -> ['tx1']
g.gfql(q, engine='pandas')   # GFQLTypeError [invalid-node-reference] Error executing 'rows': Unalignable boolean Series provided as indexer
```

Replacing the third predicate with `t.type = 'transaction'` or a map `(t {type: 'transaction'})` works on pandas. Engine parity gap: same query, Polars answers, pandas raises. Found while adding tested Cypher twins to `gfql/about.rst` examples (#2017 follow-up).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WwMmVFo44ADiRRj5cxh7i1

lmeyerov:
Fix: #2045 (apply single-alias predicate pushdown masks by position) — 77/77 green incl. the polars lane, MERGEABLE on master, awaiting merge; the three-IN shape is pinned there on pandas/polars/cuDF. Closes on merge.

PR #2046, opened 2026-09-05
fix(gfql): polars chain re-executes steps on the graph's original edge columns (#2039)

lmeyerov:
`[n({"id": 30}, name="m"), e_forward({"type": "HAS_CREATOR"}, name="type"), n(name="p")]` raised `GFQLSchemaError [incompatible-column-type]` on polars while pandas served it: the forward step stamps the alias marker `type` onto its edge frame, and the backward pass (`g_step_full`) and the pruned re-execution (`g_sub`) fed that stamped frame back into the hop, whose `edge_match` then compared the boolean marker to `'HAS_CREATOR'`. Both sites now take the step's edge rows joined back to the graph's original edge columns (semi-join on the chain's edge id), so re-execution sees the user values; the combine's marker placement is unchanged.

Pins: `test_polars_alias_column_collision_2039.py` (edge-alias collision, node+edge collision, node-only collision: same rows on pandas and polars under policy off and use; a 2-hop with the colliding alias is served), and the #1911 alias-scoping suite unchanged except `test_rows_route_edge_alias_colliding_with_its_own_type_filter`, which keeps pinning polars' rows-route decline on that shape (now a typed `GFQLValidationError` from the rows route rather than the chain) — that residual stays open on #2039. Local: polars chain/hop/narrow-combine/conformance/alias-scoping suites 2031 passed. Real-GPU (polars-gpu) run to follow on dgx.

Fixes the native op-list half of #2039.


lmeyerov:
Real-GPU receipt (dgx-spark, RAPIDS 26.02 image, cudf 26.02.01 / polars 1.35.2): `test_polars_alias_column_collision_2039.py` + `test_engine_polars_gpu.py` + polars chain/hop + alias-scoping suites at 662e557e7 = 1508 passed, 0 failed, 0 skipped.

lmeyerov:
READY (662e557e7): 77/77; polars re-executes steps on the graph's original edge columns (#2039); real-GPU 1508 passed / 0 skipped; the rows-route residual stays pinned and open on #2039.

lmeyerov:
Review follow-up (owner: "other specializations of the same bug" + mirror-path tests), head 0de93cef2.

**Probe.** 22 alias/column collision shapes × pandas/cuDF/polars × index policy off/use against the pandas full path (edge alias = filtered / unfiltered / src / dst / edge-id column; node alias = filtered column on the seed or the destination, node-id binding, an edge column name; hops=2; to_fixed_point; unseeded; step 2 of 2; all three colliding). Three more specializations surfaced:

1. **polars node side — fixed here.** A destination alias that names its own filtered column (`n({"type": "p"}, name="type")`) raised the same `incompatible-column-type`, and the final alias tagging joined the marker in as `type_right` with nulls where pandas replaces the column. The node step now filters the graph's own values for that column and the marker replaces a colliding column (marker-authoritative, as pandas' combine).
2. **multi-hop edge alias (every engine) — #2049.** `hops=2` / `to_fixed_point` with an edge alias = the filtered column: pandas and cuDF raise, polars emits `type_right`. Separate PR.
3. **alias = source / destination / edge-id binding column — #2050.** pandas/cuDF silently clobber the binding (the seed's edges vanish, or the edge id becomes `True`); polars raises a raw `SchemaError`. Only the node-id binding is validated today. Separate PR (validator).

**Pins.** Tests moved to the mirror path `graphistry/tests/compute/gfql/lazy/engine/polars/test_chain_alias_column_collision_2039.py`; new cross-engine matrix `graphistry/tests/compute/test_chain_alias_column_collision.py` (12 served shapes × 3 engines × scan/indexed = every single-hop collision matches the pandas full path; strict xfails for #2049 and #2050 that flip when they land; node-binding rejection pinned as the positive control). Local: 696 passed / 15 xfailed across the polars suites, cuDF 25.10 included; mypy unchanged; ruff clean. Still stacked on #2045; CI running.

lmeyerov:
READY (04ff8ed8e) — rebased onto master after #2045 landed; CI 77/77 check-runs success (the earlier red was the pre-move #2020 pin path the union rebase re-added to the polars lane list; dropped). Content unchanged since the review follow-up: polars node-side collision fix + cross-engine collision matrix + mirror-path tests.

lmeyerov:
note: #2049 goes into the eager case, and we'll separately handle alt specializations across recent fixes

PR #2054, opened 2026-09-05
refactor(gfql): route flow productized: chain_specializations layout, routes-off replay, shape registry + route harness, CI ledger

lmeyerov:
The route-centric test flow from the #2054 review, folded into one PR at the owner's request (formerly #2054 + #2061 + #2057). Stacked on #2056.

- **Layout**: `graphistry/compute/chain_specializations/{admission,hotpaths}.py` (pandas/cuDF) and `gfql/lazy/engine/polars/chain_specializations/{admission,hotpaths}.py`; each lane next to the admission predicate its dispatcher calls; `chain.py` and the polars chain only dispatch; `chain_fast_paths.py` keeps the shared seed/index helpers. Behavior-neutral. Tests mirror the paths.
- **Routes-off replay**: `GFQL_ROUTES_OFF=<route,...>` (tests/conftest.py) makes named hot paths decline so every existing test replays through the other routes; `@pytest.mark.route_engaged(...)` marks the tests that assert a route serves (skipped in that mode); `bin/test-routes-off.sh` writes the per-route ledger; the `gfql-routes-off` CI matrix publishes it non-blocking. Ledger table from the 7-mode replay: every remaining id was an engagement pin or a filed class (#2058 → #2062, #2059, #2034).
- **Registry + harness**: `tests/compute/gfql/routes/registry.py` (80 shapes registered by the specialization tables themselves) and `test_route_harness.py` (per admitted cell: served-by, same-engine general-path value parity, pandas general-path set oracle; declines recorded as the attenuation ledger → #2063).
- **CI path filter** (formerly #2057): the gfql filter now includes the chain engine files, so tck-gfql and the gfql lanes run on chain changes.
- tck-gfql's per-route ledger is graphistry/tck-gfql#196.

Receipts on this head below; earlier receipts on the closed #2061 and #2057.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QztW7jYsDd66e8rb8pJNQA

Issue #2043, opened 2026-09-05
gfql cuDF 26.02 (real GPU): chain default route and policy-forced full path disagree on prune_to_endpoints hops and duplicate node ids; tests use the removed cudf.DataFrame.from_pandas


lmeyerov:
Found by the cross-platform bar on dgx-spark (image `graphistry/test-rapids-official:26.02-gfql-polars`, cudf 26.02.01, cupy 13.6.0) while auditing stack B (#2035/#2037/#2038/#2040); reproduced on master bff7c323e in the same image. Local dev boxes on cudf 25.10 pass, so CI and local runs never see it.

1. `graphistry/tests/compute/test_chain.py::test_fast_path_differential_parity_vs_full_path[prune_endpoints_fwd-cudf]`, `[prune_endpoints_rev-cudf]` and `test_fast_path_dedups_duplicate_node_ids_on_hop[cudf]`: `g.gfql(ops)` (default route) and `g.gfql(ops, policy=...)` (policy-forced full BFS) return different node sets on cuDF — `[0,1,2,3,4]` vs `[1,2,3,4]` for the forward prune, `[0,1,2,3]` vs `[0,1,2,3,4]`-style for reverse, and `[0,0,1,2]` vs `[0,1,2]` for the duplicate-id hop. Same shapes agree on pandas and on cudf 25.10. Related: #2034 (duplicate/null ids on the full path).
2. 18 test sites call `cudf.DataFrame.from_pandas(...)`, which cuDF 26.2 removed (`cudf.from_pandas` is the surviving spelling): with a `-k` selection the fixtures raise `AttributeError` before any assertion, so most cuDF differential tests silently do not run on the current image.

Expected: one answer per shape on every engine; the fixture spelling accepted by both cuDF lines.

lmeyerov:
Product code has the same removed spelling at 4 sites (`graphistry/ai_utils.py:504,506`, `graphistry/umap_utils.py:619,1125`), so the AI/UMAP cuDF paths raise on cuDF 26.2 as well. `cudf.from_pandas` is accepted by both cuDF lines; PR to follow for the tests and these sites.

lmeyerov:
[...]

PR #2055, opened 2026-09-05
fix(gfql): one op-list alias-shadowing contract on every engine (#2049, #2039 polars edge side)

### LINKED ISSUE #2049: gfql: an edge alias that collides with the column its own filter uses breaks multi-hop steps (pandas/cuDF raise, polars mis-names)
## Summary
Follow-up to #2039 / #2046 (single-hop polars). The same marker-before-filter defect exists on the multi-hop paths, on every engine:

- **pandas and cuDF**: `e_forward({"type": "HAS_CREATOR"}, hops=2, name="type")` and the `to_fixed_point=True` form raise `GFQLSchemaError [incompatible-column-type]: column "type" is numeric but filter value is string`. The alias marker is written onto the edge frame after hop 1 and hop 2's `edge_match` then filters the boolean marker.
- **polars**: serves the same rows; the shadowed column keeps the user values and the marker lands beside it (`<name>_right`). That is the pinned polars alias-scoping contract (`test_alias_scoping_semantics.py`: a shadowed alias reads the user value), not a defect — the two engines differ by design in what the colliding column holds, and agree on rows.

Single-hop steps (`hops=1`, also multi-step chains where each step is single-hop) are correct on every engine after #2046.

## Repro
```python
import pandas as pd, graphistry
from graphistry.compute.ast import n, e_forward
nodes = pd.DataFrame({"key": [1, 2, 3, 4], "id": [10, 20, 30, 40], "type": ["p", "p", "m", "m"]})
edges = pd.DataFrame({"s": [3, 3, 4, 1], "d": [1, 2, 1, 4], "type": ["HAS_CREATOR", "OTHER", "HAS_CREATOR", "OTHER"]})
g = graphistry.nodes(nodes, "key").edges(edges, "s", "d")
g.gfql([n({"id": 30}, name="m"), e_forward({"type": "HAS_CREATOR"}, hops=2, name="type"), n(name="p")])  # raises
g.gfql([n({"id": 30}, name="m"), e_forward({"type": "HAS_CREATOR"}, hops=2, name="e"), n(name="p")])     # control: served
```
Same with `to_fixed_point=True`. Found by the alias/column collision probe run for the #2046 review (22 collision shapes × 3 engines × index policy).

## Expected
The eager engines serve the multi-hop step like the single-hop one: the step's own `edge_match` sees the graph's column values on the backward re-execution, and the rows match polars.

## Pins
Strict-xfail pins for these shapes live in `graphistry/tests/compute/test_chain_alias_column_collision.py` (every engine) and flip when this is fixed.

lmeyerov:
Closes #2049. Two commits.

**Contract (the surface decides, not the engine).** Op-list chains: an alias named like a column of the frame it marks becomes the boolean marker under that name, on pandas, cuDF and polars alike (shadowing). Cypher: the variable keeps reading the user value through its properties on every engine (`MATCH (a)-[type:K]->(b) RETURN type.type` answers the user values), because the shadowed values ride under the internal restore column that the row pipeline already resolves.

**Commit 1 (eager multi-hop, #2049).** A non-simple step (`hops=2`, `to_fixed_point`) whose edge alias names the column its own `edge_match` filters on had the chain's backward pass re-read the marker instead of the user values; the backward re-execution now runs on the graph's own edge rows for the step's ids (`_step_with_source_edge_columns`).

**Commit 2 (polars edge side).** The polars chain kept the user values under the alias and leaked the marker as a `type_right` join suffix (`_combine_edges` joined the marker frame onto the original edges). Now the marker takes the name and the user column moves to `shadow_restore_column(alias)`; the polars rows-route resolver and the binding-rows payload prefer that column for `alias.alias`, and the rows-route restore short-circuits when it is present. No `_right` columns.

**Pins.** `test_chain_alias_column_collision.py`: every served single-hop shape and both multi-hop shapes compare the marker values across pandas/cuDF/polars and reject join suffixes. `.../polars/test_chain_alias_column_collision_2039.py`: single-hop and two-hop edge alias marker equals pandas'. `test_alias_scoping_semantics.py`: the polars residual (`type.type` raised a typed decline) flips to parity with pandas; the other 14 scoping pins unchanged (Cypher reads user values on both engines).

**Visible change.** On polars, an op-list result whose alias shadowed an edge column now carries the marker under the alias and the user values under `__gfql_shadow_restore__<alias>__` instead of `<alias>` + `<alias>_right`. Stated in CHANGELOG.

Receipts: collision + 2039 + scoping + polars seed-resolution suites 222 pass / 9 xfail (cuDF on); guards + mypy green; broad suites running, dgx lane on the stack head to follow.


lmeyerov:
READY (3312f003f, on master b8f3e9965) — CI 69/69 non-skipped check-runs success; the 8 skipped lanes are the gfql path filter (chain.py is outside it; #2057 proposes the fix). tck-gfql run locally from its current main (66fee10, the ref CI would use) against this head: exit 0 — 3658 scenarios, 2964 supported, 1977 translated-supported, 689 xfail (baseline), direct-Cypher non-validation debt 0, no outcome drift. Local suites: 606 passed / 9 xfailed across collision matrix, chain, hop, alias-scoping, residual-polars, hop-scaling and native-seed with the cuDF lane; guards OK; mypy = master; ruff clean.

graphistry/tests/compute/test_chain_alias_column_collision.py
```diff
@@ -73,15 +73,14 @@ def test_single_hop_collisions_match_the_pandas_full_path(engine, indexed, shape
...
 def test_multi_hop_collisions_match_the_control_alias(engine, shape):
     ops = MULTI_HOP[shape]
     g = _graph(engine, False)
     expect = _sig(g.gfql([n({"id": 30}, name="m"), e_forward({"type": "HAS_CREATOR"}, hops=ops[1].hops, to_fixed_point=ops[1].to_fixed_point, name="e"), n(name="p")], engine=engine))
     res = g.gfql(ops, engine=engine)
     assert _sig(res) == expect
-    edges = res._edges.to_pandas() if hasattr(res._edges, "to_pandas") else res._edges
-    assert "type_right" not in edges.columns and bool(edges["type"].astype(bool).all())
+    # column contents after a collision are per-engine contracts (pandas: the marker replaces the
+    # column; polars: the shadowed column keeps the user values, test_alias_scoping_semantics.py)
```
lmeyerov: Per-engine shadowing semantics feels wrong

lmeyerov:
GPU receipt (dgx-spark, graphistry/test-rapids-official:26.02-gfql-polars, cudf 26.02.01 / polars 1.35.2, safe_run, TEST_CUDF=1; suites test_chain.py, test_hop.py, test_chain_alias_column_collision.py, tests/compute/gfql/**) on the restacked stack head a54ac4b35 (#2065, containing #2055 rework, #2056, #2054, #2061, #2062): 13386 passed / 95 failed / 38 xfailed / 4 xpassed. 93 of the failures are master 86de0f5d5's polars-gpu fused-lane set (#2064, fixed separately in #2066). The other 2 are strict XPASS of the #2043 prune_to_endpoints markers on cuDF 26.02: the prune shapes now agree fast-vs-full on this stack (the #2062 closure fix removed the divergence), so #2062 drops that marker in a follow-up commit (tree otherwise identical; #2065 restacked).

lmeyerov:
CI receipt at 510dcc63c (rework with the unified alias-shadowing contract): 72 check-runs success, 5 path-skipped as whole jobs (tck-gfql, test-polars ×6, test-gfql-core, gfql-benchmarks and the guards green). Broad local suites 12557 pass; GPU receipt on the stack head above.

Issue #2051, opened 2026-09-05
gfql polars: an unnamed untyped single-hop chain keeps duplicate node rows that pandas, cuDF and polars' own multi-hop collapse


lmeyerov:
## Summary
Sibling of #1993 (polars `hop()` de-dups its output node table like pandas). The polars **chain** path still returns a duplicated node row for the unnamed, untyped single-hop shape; every other shape — named steps, a typed edge, `hops=2`, undirected, `hop()` itself — collapses the duplicate, as pandas and cuDF do on every shape.

## Repro
```python
import pandas as pd, polars as pl, graphistry
from graphistry.compute.ast import n, e_forward
nodes = pd.DataFrame({"key": [1, 2, 3, 4, 5], "id": [10, 20, 30, 40, 50]})
nodes = pd.concat([nodes, nodes.iloc[[0]]], ignore_index=True)          # key 1 twice
edges = pd.DataFrame({"s": [1, 1, 2, 3, 3, 4], "d": [2, 3, 3, 1, 1, 5], "eid": range(6)})
g = graphistry.nodes(pl.from_pandas(nodes), "key").edges(pl.from_pandas(edges), "s", "d", "eid")
g.gfql([n({"key": 1}), e_forward(), n()], engine="polars")._nodes["key"].to_list()              # [1, 1, 2, 3]
g.gfql([n({"key": 1}, name="a"), e_forward(name="e"), n(name="b")], engine="polars")._nodes["key"].to_list()  # [1, 2, 3]
g.gfql([n({"key": 1}), e_forward({"type": "KNOWS"}), n()], engine="polars")   # [1, 2, 3] with a typed edge column
```
pandas gives `[1, 2, 3]` for all three. Same with `index_policy="off"` and with the seeded polars lane and the chain fast path forced to decline, so it is the general `_chain_traversal_polars` route for the simple single-hop shape.

Found by the sibling-specialization sweep run after #2046 (`plans/gfql-benchmark-numbers/spec-sweep/`).

## Expected
The polars chain collapses duplicate node rows on every shape, matching pandas' combine (the contract `test_fast_path_dedups_duplicate_node_ids_on_hop` states for pandas/cuDF).

## Pins
A strict-xfail pin lives in `graphistry/tests/compute/gfql/lazy/engine/polars/` next to the #2039 pins and flips when this is fixed.

Issue #2050, opened 2026-09-05
gfql: an alias named like the source, destination or edge-id binding column is not rejected (silent clobber on pandas/cuDF, raw polars SchemaError)


lmeyerov:
## Summary
A node alias equal to the node-id binding column is rejected up front with `GFQLValidationError` on every engine. The three other binding columns are not guarded:

| alias | pandas / cuDF | polars |
|---|---|---|
| edge alias = source column (`name="s"`) | served, but the marker overwrites the source column: result has **no edges** | served with an extra `s_right` column |
| edge alias = destination column (`name="d"`) | served, destination column replaced by `True` (`(3, True)` edge pairs) | served, extra `d_right` |
| edge alias = edge-id binding (`name="eid"`) | served, edge-id column replaced by `True` | raw `polars.exceptions.SchemaError: datatypes of join keys don't match` (not a GFQL error) |

## Repro
```python
import pandas as pd, graphistry
from graphistry.compute.ast import n, e_forward
nodes = pd.DataFrame({"key": [1, 2, 3], "id": [10, 20, 30]})
edges = pd.DataFrame({"s": [3, 3], "d": [1, 2], "type": ["HAS_CREATOR", "OTHER"], "eid": [100, 101]})
g = graphistry.nodes(nodes, "key").edges(edges, "s", "d", "eid")
ops = [n({"id": 30}, name="m"), e_forward({"type": "HAS_CREATOR"}, name="s"), n(name="p")]
g.gfql(ops)._edges          # empty on pandas; the seed's HAS_CREATOR edge is gone
```
Found by the alias/column collision probe run for the #2046 review.

## Expected
The chain validator rejects an alias that equals any binding column (source, destination, edge id) the same way it rejects the node-id binding, on every engine, before execution.

## Pins
Strict-xfail pins live in `graphistry/tests/compute/test_chain_alias_column_collision.py` (every engine) and flip when the validator covers these.

PR #2017, opened 2026-09-03
docs(gfql): publish final performance results and simplify engine guides

lmeyerov:
> **Status 2026-09-16.** Rebased onto current master **c9d7d5360** (head `b3dd07cbc`); 31 commits replayed with no conflicts. The vendored-number freshness contract is intentionally RED, and now on **two** checks rather than one: the vendored SNB runs are from `f7a7253bc` and the GraphBench key `graphbench.100k.q1.kuzu` is 17 compute commits behind, both past the 12-commit policy. Master moving 11 commits (#2085) pushed the second one over; the cause is identical and so is the fix. That fix is a single re-vendor once the open performance PRs (#2084, #2086, #2087) land, since those change the published numbers again; re-vendoring now would be thrown away and would churn the measured prose twice. Not ready to merge -- per the landing rule this goes last, after the benchmarks win.

GFQL documentation now gives readers a short start path, separate performance and language-reference hubs, and an engine guide tied to measured results. Existing page URLs are preserved. The performance hub leads with the speedup case study and separates examples and architecture from benchmark tables and methodology.

The benchmark pages use verified publication cells for values, ratios, query tallies, and generated charts. This refresh includes GraphBench Q1–Q9 at 20k/100k from PyGraphistry `d20c6ae1af73e64b05aa4b24459dfd3ba247edf9` and SNB SF0.1/SF1 from `f7a7253bc95d9cc0663bd130a53a03f2f355cbcd`. The source revisions remain distinct. The final artifact is byte-identical to merged [pyg-bench #267](https://github.com/graphistry/pyg-bench/pull/267); [#268](https://github.com/graphistry/pyg-bench/pull/268) records and enforces matching CPU placement. Historical database measurements keep their original dates and methods.

The engine guide explains input-based `auto` selection, frame conversion, unsupported-query behavior, and when Polars-GPU still executes CPU operations. Indexed message-content and creator lookups lead the measured four-engine SNB comparisons at both scales. The full tables retain every eligible result. GraphBench 20k q8 is explicitly identified as CPU-only under Polars-GPU. Instrumented GPU diagnostics are excluded from published timings.

The editorial pass addresses the recorded review items: shorter prose, defined lookup terms, consolidated engine decisions and memory guidance, simpler comparison tables, and one provenance section per page. The Neo4j/GDS and GraphFrames case studies retain their measured scope and limits. Benchmark contracts enforce source freshness, correct references, derived tallies, and chart synchronization.

Validation on the final data and prose:

[...]

PR #2056, opened 2026-09-05
fix(gfql): chain result contract: binding-column aliases decline, duplicate ids answer once, no internal columns (#2050, #2051, #2067)

lmeyerov:
Three related contract fixes on the chain result, folded into one PR at the owner's request (formerly #2056 + #2060 + #2067). Base master; first of the three-PR stack (this → #2054 → #2062), then #2017 last.

1. **Binding-column aliases are a typed decline on native chains too** (#2050): an edge alias named like `src`/`dst`/edge-id or a node alias named like the node id raises the same `GFQLValidationError` the Cypher route raises. TO-side `test_chain.py` pins re-pinned as typed declines (contract change stated here).
2. **Polars plain single-hop branch returns one row per node id** (#2051): the skip-combine branch deduplicates like the full path; strict xfails flipped. The cuDF 26.02 duplicate-id hop pin passes on this head on the real GPU (#2043 dedup half; marker dropped).
3. **The shadow-restore column exists only inside a Cypher pipeline**: `cypher_pipeline()` (context flag) marks a compiled Cypher execution; only then does the polars chain keep a shadowed user column under `__gfql_shadow_restore__<alias>__` for its row pipeline. On `gfql([...])` / `chain([...])` the marker shadows the column outright, like pandas and cuDF. Nothing is stripped from results (a blanket strip was tried and rejected: it broke pipelines composed from successive `gfql` calls and frame-returning ops).

Receipts on this head are posted below (CI, local, GPU on the stack top). Earlier receipts for each part are on the closed PRs #2060 and #2067.


lmeyerov:
READY (157a55cff, stacked on #2055) — CI 60/60 non-skipped check-runs success (7 path-skipped lanes, same filter gap as #2055). tck-gfql run locally from its current main (66fee10) against this head: exit 0 — 4143 passed / 6 skipped / 689 xfailed (baseline), 2964 supported scenarios, direct-Cypher non-validation debt 0. Local suites: 2372 passed / 24 skipped with the cuDF lane; guards OK; mypy = master; ruff clean. Contract change (TO-side binding aliases now a typed decline instead of served-with-overwrite) is stated in the description for your call.

lmeyerov:
GPU receipt (dgx-spark, graphistry/test-rapids-official:26.02-gfql-polars, cudf 26.02.01 / polars 1.35.2, safe_run, TEST_CUDF=1; suites test_chain.py, test_hop.py, test_chain_alias_column_collision.py, tests/compute/gfql/**) on the restacked stack head a54ac4b35 (#2065, containing #2055 rework, #2056, #2054, #2061, #2062): 13386 passed / 95 failed / 38 xfailed / 4 xpassed. 93 of the failures are master 86de0f5d5's polars-gpu fused-lane set (#2064, fixed separately in #2066). The other 2 are strict XPASS of the #2043 prune_to_endpoints markers on cuDF 26.02: the prune shapes now agree fast-vs-full on this stack (the #2062 closure fix removed the divergence), so #2062 drops that marker in a follow-up commit (tree otherwise identical; #2065 restacked).

lmeyerov:
CI receipt at f04fe117f (restacked on the #2055 rework): 60 check-runs success, 7 path-skipped as whole jobs; tck-gfql, test-polars ×6, test-gfql-core, gfql-benchmarks and the guards green. GPU receipt on the stack head above.

lmeyerov:
GPU receipt (dgx-spark, graphistry/test-rapids-official:26.02-gfql-polars, cudf 26.02.01 / polars 1.35.2, safe_run, TEST_CUDF=1; suites test_chain.py, test_hop.py, test_chain_alias_column_collision.py, tests/compute/gfql/**) on the folded stack top 500265e26 (#2062, containing #2056 and #2054): 13532 passed / 7 failed / 37 xfailed / 4 xpassed, versus master 86de0f5d5's 13227 / 93 in the same image. Every one of the 7 is inside master's set (5 polars-gpu strictness cases and the routing-contract / indexed-bypass pins tracked on #2064 and #1803); zero branch-specific failures, and the fused-lane fix folded from #2066 clears the other 86.

lmeyerov:
CI receipt at 5681875e1 (folded: #2050 decline + #2051 dedup + Cypher-only shadow restore; rebased onto master 60bb843c1): 77 check-runs, all success. Local on the stack top: broad CPU + cuDF 12939 pass / 0 fail; tck-gfql 4143 pass / 689 xfail (debt unchanged); GPU receipt above.

lmeyerov:
## Contract resolution (owner thread, 2026-09-06): op lists vs Cypher for an alias named like a binding column

**One rule on both surfaces, taken from openCypher: scope shadows, projection declines duplicate output names.** (`MATCH (src)-[e]->(dst)` is legal in Neo4j; `RETURN src, e.src AS src` is "multiple result columns with the same name".) Binding columns are structure, not data: the binding stays on the original column values everywhere.

- **Op lists (this PR):** projection is implicit and total, the structural columns are always projected, and `name=` is by definition an output column. An alias equal to a *same-frame* binding column (node alias == node-id; edge alias == source/destination/edge-id) is therefore a duplicate output column, and the typed decline (E108) is the Cypher-consistent answer. This PR declines only those same-frame cases; cross-frame aliases such as `n(name="src")` are unaffected and work.
  - Rejected: "structure wins, marker silently not emitted" (Cypher never drops a result column silently; downstream `.src` reads endpoints instead of a mask) and "marker wins, endpoints rebound to a generated name" (renames the user's binding unasked; the corruption shape master had).
- **Cypher:** `RETURN` projects entities, so `RETURN src` never collides. The existing Cypher-route E108 fires at the scoping layer, which is the wrong layer. That is fixed separately by a lowering rename: #2069. Not part of this PR.
- **Data columns** (e.g. a pipeline that recomputes `score` each phase) are untouched by the decline and shadow by design on both surfaces: op lists = dataframe assignment (later write wins on every engine, #2055), Cypher = `WITH` rebinding across scopes.

Dynamic pipelines that generate alias names can pre-check against `g._node`, `g._source`, `g._destination`, `g._edge`, or catch E108.

lmeyerov:
## Owner question: does the decline break TCK, q1–q9, or SNB? Checked, not reasoned.

- **TCK:** per-route ledger 4143/4143 on this head and on master. No change.
- **q1–q9:** GFQL runs bind `s`/`d` and `id`/`node_id`; aliases are computed names (`numPersons`, `outDegree`, …). The `(src:`/`(dst:` variables are in the Memgraph runner only. Unaffected.
- **SNB queries:** variables `p`, `m`, `x`, `n`, `f`. Unaffected.
- **SNB sentinel: one point would go red.** `native-hop-alias-collision` aliased the node op as `id` on a graph bound on `id`, which is exactly the same-frame binding collision this PR declines. Its stated intent is the alias-vs-data-column shadow shape, so the alias moves to `label__Message` in pyg-bench #250 (keys and bounds unchanged).

Three dgx sentinel runs with the corrected shape, gate vs the 86de0f5d5 baseline:

| run | timing bounds | collision point hash |
|---|---|---|
| this PR, 0cd9d65a6, r1 | polars/node-only-props 2.22 > 1.5×1.41; polars/node-only-entity 4.00 > 1.5×2.33 | differs (expected: shape changed) |
| this PR, 0cd9d65a6, r2 | all pass | differs (expected) |
| master 60bb843c1, control | polars/node-only-entity 3.85 > 1.5×2.33 | differs (expected) |

The polars node-only misses reproduce on master itself, so they are sentinel variance around a fast baseline run, not this PR. Baseline swap after this merges (pyg-bench #250).

## Gap found by the same repro, fixed here (2a645a3ab)

`g.chain()` (deprecated surface) bypassed the decline and died with a raw pandas `ValueError: The column label 'id' is not unique` / polars `DuplicateError`. The check now lives in chain.py and runs on both `gfql()` and `chain()`; pins are parametrized over the two surfaces plus a cross-frame negative pin (`n(name="s")` on the node frame: served, marker present). Targeted suites 212/0 incl. cuDF; lint, mypy (matches master's list), comment-density and type-hygiene guards clean; CI re-fired on 2a645a3ab.

lmeyerov:
## Receipts for head 5775f763d

- **Real-GPU lane (dgx-spark, RAPIDS 26.02 image, safe_run):** 13265 pass / 93 fail. The 93 failure ids are identical to master's set in the same image (the #2064 fused-lane set, fixed in #2062); zero branch-specific failures.
- **Broad CPU + cuDF suites (chain, hop, gfql, gfql/, collision, internal-columns, hypergraph):** 12681 pass / 1 fail. The one failure is the pandas undirected two-hop timing pin, run while another suite shared the box; it passes 3/3 in isolation.
- **Targeted:** 537/0 on chain, polars conformance, collision matrix and scoping semantics incl. cuDF; lint, mypy (same error list as master), comment-density and type-hygiene guards clean.
- **Sentinel:** three dgx runs in the comment above (pyg-bench #250).
- **CI:** running on 5775f763d; posted when it lands.

Stack restacked on this head: #2054 b5ba132fe (own product delta byte-identical before and after the rebase; helper kept out of the moved block) and #2062 c2443daa8 (own delta identical), both pushed with CI re-fired.

lmeyerov:
CI on 5775f763d: 77/77 green (check-run histogram via the API). All receipts for this head are now on the PR: CI, real-GPU lane, broad CPU+cuDF suites, targeted pins, sentinel runs (pyg-bench #250).

lmeyerov:
**Benchmarks pass with this PR, not after it.** pyg-bench #250 now carries the corrected sentinel shape, a new gate baseline taken at this head (5775f763d), and a per-engine drift tolerance derived from four back-to-back master runs. Gate at 5775f763d vs that baseline: ok, on all 18 points. Master fails that gate on exactly one field, the polars collision point's output hash, which is this PR removing the internal shadow-restore column from op-list output. TCK, q1–q9 and the SNB query arms are unaffected (checked above). #250 admin-merges on my side once its smoke CI is green.

lmeyerov:
pyg-bench #250 merged (main 56873798e): sentinel shape + baseline at this head + per-engine drift tolerance are on main. The benchmark side of this change is closed.

lmeyerov:
Docs-only commit 32ea472f0: the spec's Named Results section now states the marker rule in two sentences (the marker replaces a same-named column on the same frame; a binding-column name is rejected). No product or test change, so the GPU lane, broad-suite and sentinel receipts at 5775f763d stand; CI re-fired on 32ea472f0.

lmeyerov:
CI on 32ea472f0: 77/77 green. Head is final.

Issue #2064, opened 2026-09-06
gfql polars-gpu: the fused grouped-aggregate lane and the polars-gpu strictness tests fail on the RAPIDS 26.02 image (plan not GPU-executable on its cudf-polars)


lmeyerov:
Real-GPU lane on dgx-spark, image `graphistry/test-rapids-official:26.02-gfql-polars` (cudf 26.02.01, cupy 13.6.0, polars 1.35.2), `TEST_CUDF=1`, suites test_chain.py / test_hop.py / test_chain_alias_column_collision.py / tests/compute/gfql/**. At master 86de0f5d5: 13227 passed / 93 failed / 47 xfailed / 4 xpassed. The same 93 fail at the #2056 and #2060 heads (set difference vs master empty), so this is an image/runtime capability gap, not a branch effect:

- `tests/compute/gfql/cypher/test_grouped_aggregate_fused_polars.py` (matches_eager_twin_and_pandas ×71, serves_total_order_shapes ×8, serves_the_graph_benchmark_shapes ×2, places_nulls_the_opencypher_way ×2): on `engine='polars-gpu'` the fused lane raises `NotImplementedError: GFQL engine='polars-gpu': this query plan is not fully GPU-executable on the installed cudf-polars (NO-CHEATING: we raise rather than silently run it on CPU …)` where the tests expect the lane to serve.
- `tests/compute/gfql/test_strictness_levels.py` polars-gpu cases (absent_return_property_is_null ×2, warns_once, strict_absent_names_raise, quiet_emits_no_warning): same decline.

These pins never run in CI (#1890: polars-gpu executes in no CI lane), so the dgx receipt is the only place they show. Either the fused lane declines to the CPU polars plan on this cudf-polars (typed decline, tests adjusted to expect it on that runtime) or the plan is made GPU-executable; until then every real-GPU receipt carries these 93 as the known baseline.

lmeyerov:
[...]

PR #2062, opened 2026-09-06
fix(gfql): fixes found by the route sweeps: source dtypes on every route, polars prune decline, polars-gpu fused lane (#2058, #2053, #2064)

lmeyerov:
The three product fixes the route flow found, folded into one PR at the owner's request (formerly #2062 + #2065 + #2066). Stacked on #2054; top of the stack.

1. **Every route keeps source dtypes** (#2058): one stub-row widening artifact (id-only endpoint rows concatenated then deduplicated) mirrored on purpose at three other seams. The endpoint closure appends only missing endpoints; the bindings row frame restores an alias property's node-table dtype once its rows hold no null; the 0-row connected-join path no longer mirrors the widening; the Cypher seeded lanes keep source dtypes on every route. Contract change: pandas RETURN projections and 0-row aggregates of int/bool properties carry the source dtypes; four pins restated. The cuDF 26.02 prune_to_endpoints pins pass on the real GPU at this tree (#2043 prune half; marker dropped).
2. **Polars declines `prune_to_endpoints` on single-hop edges** (#2053): typed `NotImplementedError` and predicate decline; variable-length hops keep their native pruning.
3. **The fused grouped-aggregate lane keeps the group_by plan on the GPU target** (#2064): its value_counts alternative lowers to `unnest`, which cudf-polars has no implementation for; GPU receipt 607/5 vs master 520/91 on the two files.

Receipts on this head below; earlier receipts on the closed #2065 and #2066.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QztW7jYsDd66e8rb8pJNQA

PR #2066, opened 2026-09-06
fix(gfql): the fused grouped-aggregate lane keeps the group_by plan on the GPU target (#2064)

lmeyerov:
Closes the fused-lane half of #2064. Independent of the release stack (base master).

**What it was.** Not a dtype and not a GPU capability gap in our checks: the fused single-hop grouped-aggregate lane's low-cardinality `count(*)` alternative lowers to `col(key).value_counts()` + `unnest`, and cudf-polars 26.02 has no `unnest` map function (`NotImplementedError: Unhandled map function unnest` from `raise_on_fail=True`). `_run_fast_path_on_requested_target` converted that into "declined; plan not GPU-executable", so on `engine='polars-gpu'` every fused `count(*)` shape fell to the generic route; the values were right (the generic route answered), and the 71 + 8 + 2 + 2 fused pins failed on their served-by assertion. Found by dumping the failing frame's plan at the lane's collect on dgx.

**Fix.** `_low_cardinality_pure_count_plan` declines on the GPU target, so the lane keeps the `group_by(...).agg(pl.len())` formulation there, which cudf-polars executes (verified directly on the engine). CPU keeps the value_counts plan unchanged.

**Pin.** `test_low_cardinality_count_plan_declines_on_the_gpu_target`: under `target_mode(GPU)` the alternative returns None; under CPU it still yields the `UNNEST` plan. Runs without a GPU.

**Not in scope.** The `test_strictness_levels.py` polars-gpu cases fail on the CPU polars engine too (`select` rows op not natively supported on polars); and the absent-property placeholder in `gfql_unified.py:340` is an untyped `pl.lit(None)`, which cudf-polars rejects (`Invalid type_id`) once that path is reachable; both stay on #2064.

Receipts: local fused file 252 pass; guards + mypy green; GPU receipt below.


lmeyerov:
GPU receipt (dgx-spark, graphistry/test-rapids-official:26.02-gfql-polars, cudf 26.02.01 / polars 1.35.2, safe_run, TEST_CUDF=1) at a4b6b4b85 on `test_grouped_aggregate_fused_polars.py` + `test_strictness_levels.py`: 607 passed / 5 failed, versus 520 passed / 91 failed on master 86de0f5d5 for the same two files. Every fused grouped-aggregate pin now passes on polars-gpu; the 5 left are the strictness polars-gpu cases (`MATCH (n) RETURN n.nope_col`), which raise the polars engine's `select` rows-op decline on CPU polars too and stay on #2064.

lmeyerov:
CI receipt at a4b6b4b85: 65 check-runs success, 4 path-skipped (tck-gfql, test-polars ×6, test-gfql-core, gfql-benchmarks and the guards green). GPU receipt above (607 / 5 vs master 520 / 91 on the two files).

lmeyerov:
CI receipt at e1eb73795 (union-rebased onto master after #2055 landed; own delta unchanged): 65 check-runs success, 4 path-skipped. GPU receipt above stands (the change is in the lane's plan choice only).

lmeyerov:
Folded into #2062 (owner asked for a shallower stack); the commits, receipts and CHANGELOG entries live there unchanged.

PR #2067, opened 2026-09-06
fix(gfql): the shadow-restore column exists only inside a Cypher pipeline, never on op-list results

lmeyerov:
Follow-up to #2055's unified alias-shadowing contract. Base master, independent of the stack. Replaces the first cut of this PR (a blanket strip of `__gfql_*` columns at the op-list surface), which broke pipelines composed from successive `gfql([...])` calls: the tck-gfql phase-1 executor binds its own `__gfql_`-prefixed working columns and feeds one result into the next call, and any user who composes steps the same way would have lost them.

**Rule.** Internal columns are not created on the user surface in the first place. `cypher_pipeline()` (a context flag in `gfql/identifiers.py`) marks a compiled Cypher execution; only then does the polars chain keep a shadowed user column under `__gfql_shadow_restore__<alias>__` for the row pipeline to read back. On `gfql([...])` and `chain([...])` the marker shadows the column outright, exactly as pandas and cuDF do. Nothing is stripped from any result.

**Pins** (`tests/compute/test_gfql_op_list_hides_internal_columns.py`): no internal column on pandas, cuDF and polars for the shadow shape, marker and payload intact (fails on the previous tree for polars); Cypher `type.type` / `type.w` still read the user values on both engines; user-requested hop labels survive; a caller's own `__gfql_mine__` column survives; the deprecated `chain()` surface is clean.

Receipts: pins + hypergraph + scoping + collision matrix (cuDF on) 230 pass / 9 xfail; all 60 TCK precedence scenarios that failed on the first cut pass; guards + mypy green; broad suites and full local tck-gfql below.


lmeyerov:
Local receipt at fc6caa6ae (TEST_CUDF=1, cuDF 25.10): test_chain / test_hop / test_gfql / tests/compute/gfql / collision matrix / test_compute_chain / test_chain_let: 12784 passed, 3 failed. Two were the polars-lane completeness pins for the new test file (now registered in bin/test-polars.sh); the third, `test_compute_chain.py::TestChainBindingsTable::test_direct_rows_binding_ops_supports_zero_hop_empty_multihop_seed_row_on_cudf`, fails identically on plain master 60bb843c1 in this environment (pre-existing, not from this change; it raises inside `_gfql_connected_bindings_state` on cuDF).

lmeyerov:
Folded into #2056 (owner asked for a shallower stack); the commits, receipts and CHANGELOG entries live there unchanged.

PR #2070, opened 2026-09-07
perf(gfql): SNB point-query levers — index-hit verify, numpy hop tail, row-call trims, no edge-index scaffold for row calls

lmeyerov:
## Objective
Indexed seed lookup, message creator, and message content queries repeatedly filter small results and rebuild intermediate frames. This PR removes that overhead while preserving validation, fallback behavior, and supported query results.

## Changes
1. **An exact index hit answers a one-predicate seed without re-filtering**.
2. **Alias tagging on the seeded hop builds its flag columns in place on pandas and cuDF**.
3. **An index hit verifies its residual scalar predicates on the hit rows** instead of running the canonical filter (~0.3 ms of fixed overhead on a one-row frame); the typed E302 errors are kept.
4. **The seeded typed hop validates endpoints on its gathered rows with native NumPy or CuPy arrays** on pandas and cuDF numeric IDs. Null and incompatible ID cases retain the generic path.
5. **`rows()` / `select` on pandas skip the copies and null handling a one-row projection does not need** (bool marker mask short-circuit; no copy before a mask filter; the projection frame built directly; zero-row edge slice without copy).
6. **A chain of plain row-table calls skips the edge-index scaffolding** (an O(E) column write and drop per call). This also fixes a leaked `__gfql_edge_index__` column on an edge row table.
7. **The alias lookup frame for binding rows is built in three frame ops** rather than one per column (joined-rows route 15 → 13 ms locally); the no-scaffold result drops the row context.
8. **The polars single-node branch is a route** (admission predicate, lane body, routes-off target, harness parity). No behavior change; the harness then showed the polars *general* path dropping null-id rows and collapsing duplicate ids on node-only op lists where the lane and the pandas oracle keep them: filed as #2071 (with #2034), ledgered as known divergences of that route.

## Review fixes at e5f67a7ce
- Declared graph and AST state uses direct attributes. Optional endpoint annotations match the existing guards; backend capability checks remain where needed.
- cuDF alias insertion and endpoint filtering use native masks and arrays. Nullable residual predicates reject nulls correctly. Alias collisions and duplicate requests retain generic handling.
- GPU strictness and count corrections are included. Redundant comments were removed.
- Added 81 review-driven tests covering backend agreement, nulls, aliases, input preservation, mixed ID widths, empty/repeated edges, dangling endpoints, and both directions.

The [review response](https://github.com/graphistry/pygraphistry/pull/2070#issuecomment-5597770199) maps the changes to the inline and overall review comments.

## Historical benchmark measurements
These measurements accompanied the earlier implementation, before the final review revision. They are retained as historical evidence and are not exact-head e5f67a7ce measurements. DGX Spark, indexed pandas, median of three runs in milliseconds; competitor cells are from the published board.
| query | SF0.1 now (was) | SF1 now (was) | Kuzu SF0.1 / SF1 | Memgraph | Neo4j |
|---|---|---|---|---|---|
| seed_lookup | 3.06 (3.86) | 3.19 (4.14) | 2.33 / 2.52 | 0.59 / 0.52 | 3.50 / 2.88 |
| message_creator | 1.77 (2.33) | 1.71 (2.41) | 1.43 / 1.51 | 0.60 / 0.31 | 2.63 / 1.67 |
| message_content | 1.12 (1.66) | 1.08 (1.68) | 0.72 / 0.67 | 0.49 / 0.29 | 2.26 / 2.44 |

W/T/L on the point queries, pandas-idx: vs Neo4j SF0.1 3/0/0, SF1 1/2/0 (was 1/1/2 and 1/0/2); vs Kuzu 0/0/3 at both scales (0.62–0.88×, up from 0.40–0.63×); vs Memgraph 0/0/3 (0.16–0.43×). These historical comparisons measure the pandas changes.

Sentinel at 7eec9cc9e: gate ok against the master baseline; pandas points 8–28% faster (node-only-props 0.57 → 0.41 ms).

## Remaining performance work
The remaining 1–3 ms per call is orchestration: two chain entries per op list, the call executor, and Plottable rebinds. Memgraph answers these in 0.3–0.6 ms. Reaching that needs a dedicated point-lookup route (index gather → row table → projection in one pass) and, for seed_lookup, a one-call arm; both are follow-ups under the same objective. The polars point shapes decline the seeded lane and take the full chain; they need their own pass.

## Validation at e5f67a7ce
- DGX RAPIDS 26.02: 13,947 passed, 55 skipped, 67 expected failures, and eight non-strict unexpected passes. A real Polars GPU collect preflight passed.
- Local suites: 422 focused tests, 1,733 shared tests, 568 strictness/aggregate/row-pipeline tests, and 149 Polars admission/route tests passed. These suites overlap and are not summed.
- TCK: 4,143 tests passed. Repository lint and ten-source-file type checks passed.
- [Hosted CI](https://github.com/graphistry/pygraphistry/actions/runs/34322676723) succeeded; all 84 reported checks passed on the reviewed head.

The reviewed revision is pushed. Owner review and merge remain pending.

### DIFF
--- a/graphistry/compute/chain.py
+++ b/graphistry/compute/chain.py
@@ -1228,6 +1228,9 @@ def _chain_impl(
             if added_edge_index:
                 final_edges_df = g_out._edges.drop(columns=[g._edge])
                 g_out = self.nodes(g_out._nodes).edges(final_edges_df, edge=original_edge)
+            else:
+                from .gfql.exec_context import clear_row_exec_context
+                g_out = clear_row_exec_context(g_out)  # the row context must not escape on a result
             success = True
         else:
             # Phase 2: Backward pass to propagate downstream constraints.
--- a/graphistry/compute/chain.py
+++ b/graphistry/compute/chain.py
@@ -33,7 +33,7 @@ logger = setup_logger(__name__)
 
 
 def _filter_edges_by_endpoint(
-    edges_df: DataFrameT, nodes_df: Optional[DataFrameT], node_id: str, edge_col: str
+    edges_df: DataFrameT, nodes_df: Optional[DataFrameT], node_id: Optional[str], edge_col: Optional[str]
 ) -> DataFrameT:
     if nodes_df is None or not node_id or not edge_col or edge_col not in edges_df.columns:
         return edges_df
@@ -189,8 +189,7 @@ def combine_steps(
     Collect nodes and edges, taking care to deduplicate and tag any names
     """
 
-    id = getattr(g, '_node' if kind == 'nodes' else '_edge')
-    df_fld = '_nodes' if kind == 'nodes' else '_edges'
+    id = (g._node if kind == 'nodes' else g._edge)
     op_type = ASTNode if kind == 'nodes' else ASTEdge
 
     if id is None:
@@ -198,10 +197,10 @@ def combine_steps(
 
     logger.debug('combine_steps ops pre: %s', [op for (op, _) in steps])
     if kind == 'edges':
-        node_id = getattr(g, '_node')
-        src_col = getattr(g, '_source')
-        dst_col = getattr(g, '_destination')
-        full_nodes = getattr(g, '_nodes', None)
+        node_id = g._node
+        src_col = g._source
+        dst_col = g._destination
+        full_nodes = g._nodes
 
         has_multihop = any(
             isinstance(op, ASTEdge) and not op.is_simple_single_hop()
@@ -228,7 +227,7 @@ def combine_steps(
 
                 prev_nodes = label_steps[idx - 1][1]._nodes if label_steps and idx > 0 else g._nodes
                 next_nodes = label_steps[idx + 1][1]._nodes if label_steps and idx + 1 < len(label_steps) else None
-                direction = getattr(op, 'direction', 'forward') if isinstance(op, ASTEdge) else 'forward'
+                direction = op.direction if isinstance(op, ASTEdge) else 'forward'
 
                 if direction == 'undirected' and prev_nodes is not None and next_nodes is not None and node_id:
                     # isin() dedups internally -> the .unique() pass is redundant
@@ -253,8 +252,8 @@ def combine_steps(
     def apply_output_slice(op: ASTObject, op_label: ASTObject, df):
         if not isinstance(op_label, ASTEdge):
             return df
-        out_min = getattr(op, 'output_min_hops', None) or getattr(op_label, 'output_min_hops', None)
-        out_max = getattr(op, 'output_max_hops', None) or getattr(op_label, 'output_max_hops', None)
+        out_min = (op.output_min_hops if isinstance(op, ASTEdge) else None) or op_label.output_min_hops
+        out_max = (op.output_max_hops if isinstance(op, ASTEdge) else None) or op_label.output_max_hops
         if out_min is None and out_max is None:
             return df
         label_col = op_label.label_node_hops if kind == 'nodes' else op_label.label_edge_hops
@@ -273,19 +272,19 @@ def combine_steps(
 
     dfs_to_concat = []
     extra_step_dfs = []
-    base_cols = set(getattr(g, df_fld).columns)
+    base_cols = set((g._nodes if kind == "nodes" else g._edges).columns)
     for idx, (op, g_step) in enumerate(steps):
         op_label = label_steps[idx][0] if idx < len(label_steps) else op
-        step_df = apply_output_slice(op, op_label, getattr(g_step, df_fld))
+        step_df = apply_output_slice(op, op_label, (g_step._nodes if kind == "nodes" else g_step._edges))
         if id not in step_df.columns:
-            step_id = getattr(g_step, '_node' if kind == 'nodes' else '_edge')
+            step_id = (g_step._node if kind == "nodes" else g_step._edge)
             raise ValueError(f"Column '{id}' not found in {kind} step DataFrame. "
                            f"Step has id='{step_id}', available columns: {list(step_df.columns)}. "
                            f"Operation: {op}")
         dfs_to_concat.append(step_df[[id]])
 
     for _, (_, g_step) in enumerate(label_steps):
-        step_df = getattr(g_step, df_fld)
+        step_df = (g_step._nodes if kind == "nodes" else g_step._edges)
         if id not in step_df.columns:
             continue
         extra_cols = [c for c in step_df.columns if c != id and c not in base_cols and 'hop' in c]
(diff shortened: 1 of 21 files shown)


lmeyerov:
Real-GPU lane (dgx-spark, RAPIDS 26.02 image, safe_run) at b73cab42a: 13830 pass / 7 fail. The 7 failure ids are identical to master's known set in that image (#2064 strictness cases + the routing-contract decline pin); zero branch-specific failures. Sentinel and SNB arm receipts are in the body; broad CPU+cuDF suites, tck-gfql and CI follow.

lmeyerov:
## Receipts at 40edaf611
- Broad CPU+cuDF suites (chain, hop, gfql, gfql/, specializations, collision, internal columns, hypergraph): 13253 pass / 0 fail.
- tck-gfql: 4143 pass / 689 xfail (unchanged).
- Real-GPU lane (dgx-spark, RAPIDS 26.02): 13854 pass / 7 fail, failure ids identical to master's known set.
- CI: running; posted when it lands.

## Polars point latency on dgx is thread fan-out (sentinel, same head, same window, ms)
| point | default | POLARS_MAX_THREADS=4 | =1 | pandas |
|---|---|---|---|---|
| is1-official | 6.16 | 3.69 | 2.46 | 1.79 |
| is5-official | 4.98 | 2.07 | 1.21 | 0.89 |
| seeded-hop-props | 4.21 | 5.66 | 1.22 | 0.85 |
| seeded-hop-entity | 5.37 | 4.87 | 2.61 | 2.86 |
| node-only-entity | 2.48 | 3.73 | 2.07 | 2.35 |
| native-hop-alias-collision (full chain, bulk) | 27.35 | 41.19 | 61.81 | 1.84 |

On the 20-core ARM box the per-call cost of the polars point shapes is the thread pool, not the query: at one thread they land within 1.4× of pandas, and the bulk shape gets slower as expected. Polars has no per-collect thread control, so this PR does not change polars behavior; it records the finding. Consequence for the board: pandas is the point-query engine and polars the bulk engine, which is what the engines page already recommends.

lmeyerov:
CI on 40edaf611: 84/84 green. Ready for review: all receipts on this PR (broad CPU+cuDF 13253/0, tck 4143/689 xfail, real-GPU lane = master's known set, sentinel gate ok with pandas points 8–28% faster, SNB arm table, polars thread experiment).

graphistry/compute/gfql/row/frame_ops.py
```diff
@@ -46,14 +46,16 @@ def _empty_like(df: Any) -> Any:
...
-    return df.iloc[0:0].copy()
+    return df.iloc[0:0]  # a zero-row slice is already a new frame
 
 
 def _alias_true_mask(table_df: Any, source: str) -> Any:
     """Boolean row mask of an alias-marker column with NULL→False (pandas/cuDF; the
     polars equivalent expr is ``pl.col(source).fill_null(False).cast(pl.Boolean)``).
     Shared by ``rows``/``count_table`` so the null handling can't diverge."""
     mask = table_df[source]
+    if str(getattr(mask, "dtype", "")) == "bool":
```
lmeyerov: can we move away from getattr in all these? audit, fix..

graphistry/compute/chain_specializations/hotpaths.py
```diff
@@ -120,6 +124,36 @@ def _seeded_typed_hop_pandas_cudf(
...
 
 
+def _seeded_hop_tail_numeric(
+    cand: DataFrameT, edges: DataFrameT, n2f: Dict[str, object], src: str, dst: str, to_col: str, node: str,
+) -> Optional[Tuple[DataFrameT, DataFrameT]]:
+    """The endpoint validation and destination filter on the few gathered rows, as numpy array ops on
+    pandas frames with numeric id columns; None keeps the engine-generic Series path."""
+    import numpy as np
+    import pandas as pd
+    if not isinstance(cand, pd.DataFrame) or not isinstance(edges, pd.DataFrame):
```
lmeyerov: does cudf need this?

graphistry/compute/chain.py
```diff
@@ -1210,6 +1228,9 @@ def _chain_impl(
             if added_edge_index:
                 final_edges_df = g_out._edges.drop(columns=[g._edge])
                 g_out = self.nodes(g_out._nodes).edges(final_edges_df, edge=original_edge)
+            else:
+                from .gfql.exec_context import clear_row_exec_context
+                g_out = clear_row_exec_context(g_out)  # the row context must not escape on a result
```
lmeyerov: most of the comments in this pr seems unnecessary, better handled by better variable naming, adding tests for the intended behavior, or when straight up redundant, just delete

graphistry/compute/chain_fast_paths.py
```diff
@@ -73,6 +76,44 @@ def _tag_fast_path_aliases(
     return res.nodes(nodes).edges(edges)
 
 
+def _tag_fast_path_aliases_pandas(
```
lmeyerov: x-platform for cudf?

lmeyerov (commented):
feels brittle

* sufficient positive/negative tests on either side of boundaries?
* saw some dynamic typing
* looked like missing cross-platform support & testing, eg, cudf
* I have to wonder about test amplification in general, and true generlaization...

lmeyerov:
Addressed the inline comments and overall review in e5f67a7ce.

- Added 81 review-driven cases for null handling, alias combinations, input preservation, mixed ID widths, empty/repeated edges, dangling endpoints, and both traversal directions.
- Generalized alias insertion and numeric endpoint filtering to cuDF. Endpoint arrays remain on the GPU; unsupported nullable/mixed-ID cases retain the generic path.
- Fixed indexed all-null residual masks, nullable string comparisons in endpoint filtering, and nullable cuDF alias-mask normalization. Repeated requested aliases now retain generic handling.
- Replaced dynamic access to declared graph/AST attributes, corrected optional endpoint annotations, and removed redundant comments. Also included the missing-property strictness/count fixes needed for the real Polars GPU gate.

Validation at e5f67a7ce:
- DGX RAPIDS 26.02 (cuDF 26.02.01, Polars 1.35.2), with successful real Polars GPU collect preflight: 13,947 passed, 55 skipped, 67 expected failures, 8 non-strict unexpected passes; zero failures, process exit 0.
- Local final focused suite: 422 passed; shared suite: 1,733 passed; strictness/aggregate/row pipeline: 568 passed; Polars admission/routes: 149 passed. These suites overlap.
- TCK: 4,143 passed, 6 skipped, 689 expected failures. Full lint and changed-source type checks passed.

The new head is pushed; hosted CI is being checked separately. No merge performed.

PR #2087, opened 2026-09-15
perf(gfql): index low-cardinality columns so label and type predicates compare codes

lmeyerov:
## Problem

A label or edge-type predicate is a scalar equality on a column with a handful of distinct values. On the seeded bindings path each one still cost a frame filter: select the predicate column, gather the candidate rows, filter, read the survivors back. On LDBC SNB IC8 `recent-replies` that is seven such filters, about 1.7 ms of a 6.0 ms query.

Measured in isolation, the same predicate answered from a coded column costs 0.011 ms against 0.179 for a node label, and 0.004 against 0.145 for an edge type.

## Change

`gfql_index_all` now builds a `CategoryIndex` for each eligible column: a per-row integer code plus the value-to-code map. The array bindings path answers a predicate by comparing the candidates' codes, which is one array gather.

Three decisions carry the correctness:

- **Nulls get a reserved code**, never handed out for a queried value, so a null row matches no scalar predicate. That is exactly what the canonical filter's three-valued logic already decides, and it is what makes the index useful in practice: the SNB label columns are null-heavy after their concat, and a null-refusing version skipped every one of them.
- **A value the column never holds matches nothing**, answered here rather than deferred, because that is the same answer the filter gives.
- **Boolean and integer values never share a code.** `True == 1` in Python; the canonical filter does not conflate them, so the code lookup matches on type as well as value.

Everything else declines to the canonical filter, which is the same answer: other engines, float and other dtypes, and cardinality above 64.

Index build is a declared setup step like the adjacency build, not lazy per-query work. On SF0.1 it takes `gfql_index_all` from 101.5 ms to 144.7 ms.

## Measured

LDBC SNB SF0.1, H684 index lane recipe, local box, interleaved against the base branch over three rounds:

| query | base | this branch |
| --- | --- | --- |
| recent-replies | 6.087 / 5.930 / 6.033 | 4.595 / 4.658 / 4.997 |
| message-replies | 2.716 / 2.812 / 2.649 | 2.175 / 2.122 / 2.232 |

About -23% and -20%. Point queries are unchanged.

Three runs per arm against merged master 65c359b5d, canonical rows identical in every cell.
This measures the whole stack (#2084 + #2086 + this), since each builds on the last.

| scale | engine | query | master | stack | delta |
| SF0.1 | polars | recent-replies | 29.291 | 5.528 | -81.1% |
| SF0.1 | polars | message-replies | 8.102 | 1.465 | -81.9% |
| SF0.1 | polars | seed-lookup | 1.068 | 0.930 | -12.9% |
| SF0.1 | pandas | recent-replies | 44.743 | 27.376 | -38.8% |

Every other cell is flat or inside its own noise. SF1 pandas new-topics reads +10.3%, which is
the cell whose identical-code A/A control spans 472.8 to 567.5 ms — see #2084 for that
investigation; its GFQL surface reaches none of this code.

This branch carries three commits: category indexes, then an endpoint-row index that resolves each
edge endpoint to its node row at build time, then a temporal-text fact that turns the projection's
constructor-text guard into a build-time lookup.

Against the competitor arms, `message-replies` now beats Kuzu by about sixteen times.
`recent-replies` has gone from a 5.4x loss against Neo4j to **parity**: its three runs are
5.122, 5.607 and 5.528 ms against Neo4j's 5.425, which straddle it. That is parity, not a win,
and it should not be quoted as one. The two `seed-lookup` cells and `recent-replies` remain losses, budgeted in the
plan's remaining-gap document.

## Tests

`graphistry/tests/compute/gfql/index/test_category_index.py` covers code round-trips across Boolean, String and integer columns with and without nulls; declines for float, high cardinality, absent columns and non-polars engines; staleness against both a cloned frame and a reshaped one; and seven end-to-end predicate shapes checked against the canonical filter, including a value the column never holds and the boolean-versus-integer conflation guard.

Index suite 1127 passed. Broad suite over `tests/compute/gfql`, chain specializations, chain and hops: 16150 passed, 1 failed — the cuDF zero-hop test that fails identically on master with this box's cuDF version.

One existing test needed a fix rather than a waiver. `test_node_property_index_seeds_without_scanning` pins that a seeded query gathers its seed from the property index instead of scanning, by recording the widths passed to the frame filter. With the predicate answered positionally that probe recorded nothing, so it was blind rather than false. It now watches both filter seams; the assertion is unchanged.

Stacked on #2086, which is stacked on #2084. Review those first.


lmeyerov:
Your instinct on boundary testing found a **silent wrong answer** in this PR. Fixed in `153a7cf2d`, and the test that was written for exactly that case was passing vacuously.

## The defect

`_code_for` matched on `type(value) is type(expected)`, and `_positions_via_category_index` treated "no code found" as **matches nothing** (`positions[:0]`) rather than **cannot answer** (`None`). A polars filter *does* coerce across those types, so a cross-type scalar returned an empty result where the canonical route returns rows. Reproduced with engagement proven on the served leg:

| predicate | served (index) | canonical (oracle) | at parent commit |
| --- | --- | --- | --- |
| `{"flag": True}` on an Int64 column | `[]` | 2 rows | 2 rows, correct |
| `{"flag": 1.0}` on an Int64 column | `[]` | 2 rows | 2 rows, correct |
| `{"label__Message": 1}` on a Boolean column | `[]` | 3 rows | 3 rows, correct |

All three are correct at the parent, so this branch introduced them. The docstring asserting "the canonical filter does not conflate them" was simply wrong about its own oracle.

**Fix:** `_code_for` returns three states — the code; `None` for a value of the right type the column does not hold, which genuinely matches nothing; and `_DECLINE` when the scalar's type differs from the column's values, or the column is all-null and offers no type to compare against. The lookup defers to the canonical filter on `_DECLINE`. Same-type predicates still serve, so the fast path is not lost.

## Why the test could not see it

`test_boolean_predicate_never_matches_an_integer_column_code` existed for this exact case and passed. The whole differential corpus seeded at `n({"id": 1})` and walked `e_forward` — and node 1 has **no outgoing edge**, so every case compared `[] == []`. A fast path that drops every row is structurally invisible to an empty-vs-empty corpus.

Rewritten so it cannot recur: the corpus walks the direction that has edges; **every case asserts the oracle's row count**, so a case cannot silently go vacuous; `_both` turns `point-rows` off and asserts that both the specialization and the category lookup actually served; the canonical leg is a real oracle (`index_policy="off"` with the sibling routes off, not another fast path); and a new parametrized family covers the cross-type scalars plus a unit pin on the three-state verdict.

Verified by reinstating the bug: 5 of the new cases fail with it, 41 pass with the fix.

## Your question about more indexes — measured, and it points somewhere specific

I timed the primitives the traversal actually calls, at this head, LDBC SNB SF0.1 polars, as a share of each query's own median:

| query | median ms | searchsorted | unique | lexsort |
| recent-replies | 3.987 | 0.662 ms (16.6%) | 0.361 ms (9.0%) | 0.092 ms (2.3%) |
| message-replies | 1.855 | 0.030 ms (1.6%) | 0.052 ms (2.8%) | 0.006 ms (0.3%) |
| seed-lookup | 1.097 | 0.016 ms (1.5%) | 0.028 ms (2.6%) | 0 |
| new-topics | 68.701 | 0.049 ms (0.07%) | 0.090 ms (0.13%) | 0.008 ms (0.01%) |

Three conclusions, each killing or backing a candidate:

1. **A pre-sorted adjacency is not worth building.** The obvious next index — have the adjacency carry its rows already ordered by (key, tiebreak) so the expansion's `lexsort` disappears — is worth 2.3% of recent-replies and under 0.5% everywhere else. Rejected before building it.
2. **The lever that is indicated: key the CSR by dense node row position, not by sorted id.** `searchsorted` plus `unique` is 25.6% of recent-replies, and nearly all of it is mapping ids to positions and de-duplicating a frontier of ids. A CSR whose offsets are indexed by node row (0..N−1) makes every id→position lookup an O(1) array index and lets the frontier be a dense mark instead of a sort-and-unique. That is the same move `EndpointRowsFact` already made for edge endpoints, generalized to the traversal itself.
3. **No index helps seed-lookup or new-topics.** seed-lookup spends 0.043 ms of 1.097 in array work — it is dispatch-bound and needs a one-row materialization path, not an index. new-topics spends 0.15 ms of 68.7; it is polars compute and an index cannot reach it.

## Build cost and accounting — two gaps I have not closed yet

Worth stating rather than burying, and I would rather fix them before this merges if you agree:
- The three new structures are built **eagerly and unconditionally** by `gfql_index_all`, and `gfql_index_categories` defaults to every column of both frames, calling `unique()` on each. On a wide frame that is many full distinct passes that yield no index because the cardinality cap rejects them.
- `build_endpoint_rows_fact` has no engine gate, so a pandas graph builds a fact only the polars path can consume.
- `index_nbytes` and `show_indexes` do not count the new structures, so the memory signal under-reports. The endpoint-rows array is the largest per-edge sidecar added so far.

Local suite at this head: 13289 passed, 24 failed — the same polars-conformance set that fails identically on the untouched parent.

lmeyerov:
Two more commits since the comment above, head is now `581ac831f`, and the three build-hygiene items I flagged as open are now closed.

**`f9438d201` — build cost, engine gate, memory accounting.**
- `build_category_index` ran an exact distinct pass on every admitted-dtype column *before* the cardinality cap could reject it, so a wide frame paid a full pass per column that yields no index. It now rules out a clearly-over-cap column with `approx_n_unique()` first; the margin keeps the sketch's error away from the cap and the exact check still decides everything near it, so the cap itself cannot move. On a 200k-row, 30-high-cardinality-string-column frame: `gfql_index_categories` **230 → 56 ms**, whole `gfql_index_all` **0.490 → 0.376 s** against a 0.069 s base.
- `build_endpoint_rows_fact` is now polars-only, matching its only consumer. A pandas graph builds zero.
- `index_nbytes` and `show_indexes` now cover all three structures. On a 20k-node/80k-edge graph that is **1.38 MB of a 4.89 MB total** that was previously invisible. `test_every_kind_carries_per_row_reason` pinned an exact kind set and now uses `ALL_REPORTED_KINDS`, so a future fact that forgets to report fails rather than passing quietly.

Two docstrings asserted the opposite of the code and are corrected: both `CategoryIndex` and `build_category_index` claimed columns with nulls are not indexed. Nulls get a reserved code, which is exactly what makes the null-heavy `label__*` columns indexable.

What I did **not** optimize, deliberately: `gfql_index_endpoint_rows` is still 204 ms on 800k edges. That is real per-edge work bought to remove a per-hop search, a declared setup step like the adjacency build. It is now disclosed in `show_indexes` rather than optimized away on a hunch.

**`581ac831f` — Categorical columns.** `pl.Categorical` is admitted alongside `pl.String`, and the build joins on a mapping frame it constructs itself, which is the string-cache-sensitive case — but nothing tested it, and nothing tested that `pl.Enum` declines. Both verified and pinned, including an end-to-end family against the canonical filter with expected row counts.

lmeyerov:
## Provenance check: do the numbers in the body still hold at the current head?

The DGX A/B table in the description was measured at `6a0da0907`. The head is now `581ac831f`, which adds a correctness fix that **deliberately removes fast-path coverage** for cross-type scalars. That could have cost part of the win, so I checked rather than relabelled the numbers.

**Engagement is byte-identical.** Category lookups served vs declined, on the real LDBC SNB SF0.1 fixture:

| query | at `6a0da0907` | at `581ac831f` |
| --- | --- | --- |
| recent-replies | 6 served / 1 declined | 6 served / 1 declined |
| message-replies | 4 served / 1 declined | 4 served / 1 declined |

The one decline in each was already there before the fix — it is a column with no category index, not a cross-type scalar. No SNB predicate is cross-type, so the fix costs this benchmark nothing.

**Timings, interleaved A/B with an A/A control**, 5 rounds, arms alternating, local box, medians of per-run medians in ms. Arm A is `6a0da0907`, arm B is `581ac831f`, arm C is a second checkout of A — so C vs A is the measurement floor and only a B-vs-A delta beyond it would mean anything:

| query | A `6a0da0907` | C copy of A | B `581ac831f` | A/A floor | B vs A |
| message-creator | 0.603 | 0.602 | 0.593 | −0.1% | −1.6% |
| message-replies | 1.844 | 1.856 | 1.806 | +0.7% | −2.0% |
| recent-replies | 3.504 | 3.427 | 3.482 | −2.2% | −0.6% |
| seed-lookup | 0.966 | 1.060 | 1.049 | +9.7% | +8.6% |

Every B-vs-A delta is at or inside its own A/A floor; seed-lookup's floor is larger than its effect. Canonical rows matched expectations in all 15 runs.

So the published numbers stand at the current head. They are **not** being relabelled to it: they were measured at `6a0da0907`, and this is the evidence that the head since then did not move them.

lmeyerov:
## Retargeted to master, plus two fixes found in a final pass

`origin/master` is now `c5ce5e4f8` (#2086 landed). Base retargeted through the GitHub base field and master merged in — clean, one auto-merge in `bin/test-polars.sh`.

The coverage baseline was re-checked in both directions after the merge, since a wrong one turns master's lint red and a red lint lane skips every test lane repo-wide: `graphistry/compute/gfql/lazy/**/*.py` resolves to **24** files and the baseline holds **24** entries, with no target missing from the baseline and no baseline entry outside the target set. This PR adds no lazy-engine file, so it needs no new floor.

### 1. A docstring that contradicted its own design

`gfql_index_categories` described eligibility as *"no nulls, at most 64 distinct values"*. `build_category_index`, in this same PR, says *"Nulls ARE coded, under a reserved code"* — and the second one is right. It is also the design point the whole change rests on: the null-heavy `label__*` columns are only indexable because nulls get a reserved code, and that is where most of the measured win comes from.

Verified against the built index rather than the prose: a null-bearing column yields `value_codes {'BOUGHT': 0, 'IS_LOCATED_IN': 1}` with nulls carrying reserved code 2, and `None` is never a key — so `value_codes.get(None)` can never return the null code, and a null predicate declines structurally rather than by luck.

`test_category_index.py` already pins this (Boolean/String/Int64 columns containing None, asserting the reserved code is never handed out for a queried value), so the prose was cut to a pointer instead of being rewritten into a longer rule it had got wrong.

### 2. The one dynamic-typing probe this PR added

Scanning only this PR's own added product lines turned up exactly one: `getattr(values.dtype, "kind", "")` deciding control flow in the endpoint-row resolver. Twelve lines above, the same file already reads `str(keys.dtype.kind)` directly on the same kind of value from the same helper, with the note *"numpy/cupy arrays always carry dtype.kind"* — so the probe was guarding against a case the file itself states cannot happen. Replaced with the direct read. mypy clean across 349 files with it in place.

(The sibling probe at `build.py:434` is pre-existing on master and is deliberately left alone.)

### Verification on the merged tree

- ruff clean; lint exit 0 with **no growth** on either guard baseline; mypy **351 source files clean**; GFQL index suite **1094 passed**.
- **TCK at the pinned harness `84b098c`: 4264 passed / 182 skipped / 689 xfailed / 0 failed** — byte-identical to master `c5ce5e4f8` on the same harness.
- **Bench engagement receipts verified unchanged.** The SNB lane proves index engagement by patching `_seed_rows_via_property_index`, and this PR's category index also answers node-role lookups — so it could plausibly have bypassed that seam and silently degraded every lane receipt to "unverified". Ran the harness's own `verify_index_engagement` on both arms: identical in all six cells (`engaged` / `built_engaged` with matching probe and hit counts; `new-topics` reaches no seed seam on the **baseline** too).

PR #2088, opened 2026-09-16
perf(gfql): validate a query's ops once per call, not twice

lmeyerov:
## Problem

Every AST object in a query is validated **twice** inside a single execution. `gfql` builds a `Chain` from the caller's ops, which validates them; the execution path then constructs a throwaway `Chain` purely to validate them again, and deliberately unwraps an already-validated `Chain` to do it.

Measured redundancy factor 1.7–1.8x, by counting `ASTSerializable.validate` calls within one execution:

| query | validate() calls | distinct objects |
| --- | --- | --- |
| seed-lookup | 12 | 7 |
| recent-replies | 24 | 13 |

## This is not a benchmark artifact

Worth stating, because the obvious version of this change *would* be one. The SNB harness reuses AST objects across repetitions, so a memo keyed on object identity would show a large benchmark win and do nothing for a real caller. That was rejected.

What is removed here is redundant **within a single call**. A caller that passes a plain list to `gfql()` and reuses nothing pays both passes today.

## What is preserved

The re-validation exists to catch a caller that built a `Chain`, mutated its ops, and then executed it. That caller is unaffected and still raises — pinned for both entry points.

The signal is explicit and narrow: `Chain.gfql_validated()`, set only by `gfql` on a `Chain` it just built in the same call. It is never inferred from the `Chain` itself, and it cannot be forged onto one constructed with `validate=False`. Two sites needed it: the Polars branch, and the pandas path, which unwraps the `Chain` before validating and so loses any marker.

## Measured

In-process, arms alternated per round, 6 rounds × 31 reps, each against **its own A/A control**. LDBC SNB SF0.1, Polars.

| query | before | after | saved | A/A floor |
| seed-lookup | 0.995 ms | 0.897 ms | **−0.098 ms** (−9.8%) | 0.034 ms |
| message-replies | 1.714 ms | 1.640 ms | **−0.074 ms** (−4.3%) | 0.011 ms |
| recent-replies | 3.488 ms | 3.390 ms | **−0.098 ms** (−2.8%) | 0.022 ms |

Three to seven times its own floor on each. It is a fixed per-query cost, so it helps every query and every engine, not just these.

**No end-to-end claim is made here.** A cross-process A/B against master cannot resolve 0.1 ms: its own A/A floor ran +3.3% to +12.5%, larger than the effect. Anyone quoting a benchmark number from this needs to measure it in combination, on the bench.

## Tests

Eight new, covering both sides of the boundary:

- `gfql` validates once; `chain` called directly still re-validates.
- A `Chain` mutated after construction is still caught, through **both** entry points.
- The mark cannot be forged onto an unvalidated `Chain`.
- Malformed queries raise identical error codes through `gfql` (three shapes).

Validation suites: 746 pass. Broad compute suite: 14980 pass / 32 fail, an **identical failure set** to untouched master on this box (24 Polars-conformance, 8 umap, both pre-existing and verified by running the same selection on a clean checkout of `c9d7d5360`).

Based on current master `c9d7d5360`. Independent of #2084/#2086/#2087.


graphistry/compute/chain.py
```diff
@@ -984,8 +999,11 @@ def _chain_with_strictness(
         # POLARS_GPU = the same lazy engine with the GPU execution target.
         # (Dependency guards for polars / cudf_polars are above, pre-coercion.)
         if validate_schema:
-            # Construct a fresh validator: the constructor validates children once.
-            Chain(ops if not isinstance(ops, Chain) else ops.chain)
+            # Construct a fresh validator: the constructor validates children once. Skipped
+            # only when `gfql` built this Chain in this same call, where re-validating cannot
+            # observe a change -- every other caller may have mutated its ops since.
+            if not getattr(ops, "_gfql_validated_in_call", False):
```
lmeyerov: aviod dynamic typing

PR #2090, opened 2026-09-16
perf(gfql): narrow a seeded point query's edges on codes before building a frame

lmeyerov:
## Problem

A seeded point query gathers the seed's incident edges into a frame and then filters that frame. Attributing `seed-lookup`'s polars calls by call site shows where the time actually goes:

| polars call | ms/query | call site |
| --- | --- | --- |
| `DataFrame.filter` | 0.233 | the edge predicate |
| `LazyFrame.collect` | 0.132 | underneath that filter |
| `DataFrame.__getitem__` | 0.175 | gathering the candidates |

That is **46% of a 1.17 ms query spent materializing rows the predicate then discards.**

## Change

The candidates are row **positions** before they are a frame. When the edge predicate is one the category index from #2087 can answer, the filter runs on uint8 codes and only the surviving rows are gathered.

This is the same structure #2087 built for the array traversal, reused by the point path — the concrete answer to *"maybe its indexing work hints at more indexes we can use for the others"*. The index already existed; the point path simply was not asking it.

Anything the index cannot answer falls back to gather-then-filter unchanged, so the canonical filter stays the authority on three-valued and dtype behaviour. `_index_edge_rows` becomes a thin wrapper over a new `_index_edge_positions`, so the lookup and the gather are separable without changing what the gather does.

## Measured

Interleaved, arms alternated per round, 5 rounds × 41 reps, each against **its own A/A control**. LDBC SNB SF0.1, polars, local.

| query | before | after | delta | A/A floor |
| **seed-lookup** | 1.019 | **0.476** | **−53.3%** | −5.0% |
| message-creator | 0.572 | 0.540 | −5.5% | +4.1% |
| message-replies | 1.774 | 1.710 | −3.6% | −0.6% |
| recent-replies | 3.449 | 3.522 | +2.1% | +1.4% |
| message-content | 0.395 | 0.395 | 0.0% | +1.4% |

Only seed-lookup moves beyond its own floor, and it moves by **ten times** that floor with non-overlapping ranges ([0.987–1.065] against [0.472–0.489]). Rows correct in every arm and repetition.

**No bench claim is made here.** Local and bench are not related by a ratio — measured on the same commit, the local/bench ratio across queries spans 0.84x to 2.54x — so the competitor comparison needs its own bench run. That run is queued.

## Tests

Both sides of every boundary, from the first commit rather than retrofitted:

- **Serves:** three indexed predicates (one survivor, several survivors, a value the column never holds), each against gather-then-filter as oracle, with expected row counts asserted so a case cannot go vacuous.
- **Declines:** six predicates the index cannot answer — absent, empty, an unindexed column, a cross-type scalar, a missing column, and a mixed indexed/unindexed pair — which must fall back and still agree, including agreeing on the exception type.
- **Engagement pin:** fails if the narrowing never runs. Verified by disabling the narrowing and watching it fail.
- **No-index case:** same answer with no index built at all.

Registered in the polars lane, marked `route_engaged`, and verified across all ten routes-off modes. Broad compute suite: 15176 passed, failure set identical to untouched master.

Stacked on the experiment tree (`f332bbf9b` = #2087's head with #2088 applied), so it carries #2084, #2086, #2087 and #2088 beneath it.


lmeyerov:
## Parked pending bench numbers — and the local evidence does not currently justify merging

### What the narrowing is actually worth, isolated

Measured with both arms on the **same tree** — arm A forces `_gathered_edges_matching` back to gather-then-filter, so the only difference is this PR's narrowing — plus an A/A control arm. 10 warm / 25 reps / 3 rounds, interleaved, LDBC SNB SF0.1, polars.

| query | unnarrowed | narrowed | delta | absolute | A/A |
| **seed-lookup** | 4.309 | **4.013** | **-6.9%** | **-0.296 ms** | -0.1% |
| message-creator | 1.010 | 0.981 | -2.8% | -0.029 ms | +0.3% |
| message-replies | 2.089 | 2.063 | -1.2% | -0.026 ms | -0.3% |
| message-content | 0.801 | 0.800 | -0.1% | -0.001 ms | +0.8% |
| recent-replies | 3.536 | 3.570 | +1.0% | +0.035 ms | +0.2% |
| new-topics | 53.424 | 54.096 | +1.3% | +0.672 ms | +0.4% |

Rows identical and equal to expected in every arm and every query — no correctness cost. But it **improves one query of six, by about 0.3 ms**, and nothing else moves outside the control band.

Two corrections that follow from this:

1. **The published seed-lookup win belongs to the five-PR stack, not to this PR.** SF1 seed-lookup 1.061 -> 0.279 was measured as *all five PRs vs master*. This PR's own marginal contribution is the 0.296 ms above.
2. An earlier note credited "-53.3%" to this PR. That came from a lane whose entire envelope is 0.476 ms, where a 0.3 ms saving is indeed about half. Both readings are real; neither licenses the other, and they are not convertible by a ratio. The honest common ground is the absolute: **~0.3 ms on one query shape**.

### Where that leaves it

The standing bar for this release says a number that only helps the query it was built for is not a result. On this evidence, that is what this PR is. It is parked rather than closed because seed-lookup is the cell that was historically a **loss**, and 0.3 ms measured against a competitor is a different claim from 0.3 ms measured against our own 4 ms envelope — so the dgx bench lane, which has to run for the release anyway, decides merge-or-close.

### Kept regardless of that outcome

The tests were rewritten onto the caller-visible boundary (`9e34a8dd6`): 15 cases, 8 where the category index can answer the edge predicate and 7 where it cannot, each stating its expected rows outright so no case can pass by both routes being empty, plus a null predicate, a cross-type scalar that must raise identically on both routes, parallel edges, and a graph with no index at all. The old engagement pin was also **measuring the wrong thing** — it counted every `_positions_via_category_index` call, but that helper serves node seed lookups too, so it could report engagement for a predicate that was never narrowed; it is now scoped to the edge frame and demoted to one labelled pair.

Mutation-checked: dropping the survivor mask fails 4 cases, reversing source order fails 3. (Answering a null predicate from the code map instead of declining is an equivalent mutant — the value map never holds `None`, so both paths return empty.)

Local numbers throughout: they attribute work inside one box, sit beside no competitor arm, and are not convertible to bench numbers.

lmeyerov:
## Correction: the bench reverses my previous comment. This should merge.

My earlier comment argued this PR helps only the query it was built for and was not worth merging. **That was wrong, and the bench shows it.**

Now that #2084/#2086/#2087/#2088 are on master, I re-measured the release scoreboard on the tree that would ship **without** this PR. SF0.1 polars, 5 runs per arm, rows verified equal across arms and stable within each for all 7 cells, zero bad probes.

| arm | seed-lookup runs | median |
| --- | --- | --- |
| five PRs, **including** this one | [0.218, 0.223, 0.216, 0.223, 0.215] | **0.218** |
| four PRs, **this one removed** | [0.655, 1.013, 0.793, 0.902, 1.239] | **0.902** |

**4.1x worse without it.** Against Memgraph 0.460 that is the difference between a 2.1x win and a clear loss on the release's headline cell.

The two arms are comparable — checked, not assumed. Every other cell agrees closely between them: message-content 0.161/0.163, message-creator 0.225/0.239, message-replies 1.766/1.385, new-topics 42.026/42.266, recent-replies 5.364/5.326, tag-cooccurrence 32.849/32.176. Only seed-lookup moved.

It also **removes the variance**: ±2% with this PR (0.215-0.223), ±89% and bimodal without (0.655-1.239). That fits the mechanism precisely — the materialization this PR avoids is what was variable.

### Why my local isolation misled me

Local isolation measured -6.9% / 0.296 ms on one query of six. Both readings are real. The local SNB-mirroring lane's seed-lookup envelope is **~4.3 ms**, dominated by costs this PR never touches; the bench envelope for the same query is **0.2-0.9 ms**. The same absolute saving is 7% of one and most of the other.

I never scaled the local number into a bench claim, which is why this ended as a wrong recommendation rather than a wrong published figure. It is a good argument for the rule.

### Before merging

1. An interleaved lane measuring A = master vs B = master + this PR directly, rather than resting on a cross-day comparison — running next.
2. **pyg-bench pins on this cell are mandatory.** It is demonstrably fragile, and unpinned it would regress silently.

Tests, comments and typing are already addressed: the suite was rewritten onto the caller-visible boundary (15 cases, 8 answerable / 7 not, expected rows stated outright) and mutation-checked, the docstrings were trimmed to what the tests do not already carry, and the PR adds no dynamic typing.

lmeyerov:
## Lane 4 settles it — merging

Interleaved A/B on one master, 5 runs per arm alternating under the idle gate and the host perf lock. A = master `24c0b1eba` (the tree without this PR), B = the same master with this branch at `e9c4494f3`. Canonical rows verified equal across arms and stable within each arm for all 7 cells; zero probes reported a non-ok status.

| query | A (without) | B (with) | delta |
| --- | --- | --- | --- |
| **seed-lookup** | **0.842** | **0.222** | **-73.6%** |
| message-replies | 1.438 | 1.355 | -5.8% |
| message-creator | 0.237 | 0.224 | -5.5% |
| new-topics | 42.815 | 42.463 | -0.8% |
| message-content | 0.164 | 0.165 | +0.6% |
| recent-replies | 5.226 | 5.466 | +4.6% |

**The seed-lookup ranges are disjoint** — A [0.630, 0.787, 0.842, 0.966, 1.021], B [0.219, 0.219, 0.222, 0.224, 0.227]. Every B run beats every A run, and B's spread is 3.6% against A's 62%. Nothing else moves outside its noise, so this PR owns exactly one cell.

Against Memgraph 0.460 on the same board, that cell is a 2.07x win at 0.222 and a loss at 0.842.

## Pins

graphistry/pyg-bench#270 adds an SNB cell-floor checker — SNB had none, which is why the regression above could only be found by re-measuring an arm by hand — and commits floors for this cell at both scales. The floors guard the median AND the worst run, because without this change the cell is bimodal rather than uniformly slow, and a median-only floor would pass a cell that is intermittently 4x slow. Verified against both arms of the lane it describes: exit 0 with the change, exit 1 without, tripping on both criteria.

## Bar

CI 89/89 at `e9c4494f3`; perf measured interleaved with an unambiguous result; no dynamic typing added; tests rewritten onto the caller-visible boundary (15 cases, 8 answerable / 7 not, expected rows stated outright) and mutation-checked; comments trimmed to what the tests do not already carry; pins committed.

